PR #137, opened 2014-01-12
Add midi controls for layers (Gain & Pitch) and for instruments (CutOff)

jdekozak:
Hi

I have a Korg-ES1 where these changes are useful to control Hydrogen.

Regards,
Julien


jdekozak:
Hi,

It is now pending for almost 2 years, what should I do to get it merged into main tree ?

Regards,

trebmuh:
@jdekozak : testing maybe? If you give me a use-case step-by-step scenario, I can help here with testing and reporting/commenting.

thijz:
i have to agree with @mauser that we need to clean up the way midi actions are currently orginised
that said, i'm a big fan of the patch @jdekozak created ...

revamping the whole midi thing thing is not going to be a 0.9.7 thing so IMHO we might as well merge this patch into master and enjoy the extra functionality :-)

jdekozak:
Hi,

Thanks for your interest.

I will try to provide a step by step scenario, it should not be too hard, I have plugged 3 buttons in the end. Just give me more time to write it.

About reworking the midi actions, I am not sure to understand how to reorganize them in a tree. What I can do for now, is to provide with a unit test for the dispatching mechanism I have put in place in this patch.
=> Checking that correct function is called according to the input string, and parameters of the event are well forwarded. (It might require extracting an abstract class)

Regards,

trebmuh:
@jdekozak : no worries. Ping me whenever your step-by-step scenario is ready. I'm eager to try it!

mauser:
@jdekozak :

Hi! 
I've started to work on the foundations of Hydrogen 0.9.8 in a separate branch and began to review/integrate this pull request. From my side everything looks fine, and the integrations of midi actions as function has been well done. I will be testing this in the next days..

mauser:
@thijz : Without a new way to represent organize the midi actions, the new gain/pitch actions make the midi action dialog very hard to use for people which do not use those actions. It takes me some seconds (~8s) to scroll to the end of the list, so it is not an option for me to enable the new actions before we have a new way to display them

jdekozak:
Hi all,

@trebmuh Here is a short procedure to test the feature on Linux

Run Hydrogen
Open Preferences
Select Midi tab
Configure channel 10
Plug CC 93 with GAIN_C0_L0_LEVEL_ABSOLUTE 0
Click OK
Load a default drumkit

Run Virtual MIDI Piano Keyboard
Select 93 for the control

Connect VMPK MIDI out into Hydrogen MIDI In

Move the VMPK rotary button
See the Gain of the Kick moving in the Layer panel

Rerun the procedure for PITCH_C0_L0_LEVEL_ABSOLUTE. And of course the same procedure can also be applied for the FILTER_CUTOFF_LEVEL_ABSOLUTE

@mauser I can only agree with your remark, 8s is really too long to reach the desired action. As a workaround I can suggest to add the "targeted_element" as 2 new columns in the MidiTable widget. Hence the actions would become GAIN_LEVEL_ABSOLUTE and PITCH_LEVEL_ABSOLUTE. Code would also be clearer, the unused parameter in some actions would simply disappear. Would it be acceptable ?

mauser:
Hi @jdekozak , 

i've merged this pull request with the master branch, so you're code is now part of hydrogen's main branch. Maybe i will comment out the GAIN_LEVEL_ABSOLUTE  and PITCH_LEVEL_ABSOLUTE actions until we have a better way for managing them. 

My idea is that every action gets its own description (a translatable string) and a display category (also a translatable string). After those changes, you could browse those actions via the category. They could also have a weight attribute which defines how the actions are sorted in such a view. Does this sound ok for you?

jdekozak:
Wow ! Merged ! Thanks !

A category to sort them, a weight to prioritize them, it sounds good. If I have time I will give it a try and raise some pull requests.

Regards.

PR #143, opened 2014-02-16
Fix lash path in include directives

jdekozak:
Hi

Compilation was failing with lash support on Debian Sid, include directives have been changed accordingly.

Regards,
Julien

### DIFF
--- a/src/core/include/hydrogen/LashClient.h
+++ b/src/core/include/hydrogen/LashClient.h
@@ -27,7 +27,7 @@
 #ifndef LASH_CLIENT
 #define LASH_CLIENT
 
-#include <lash-1.0/lash/lash.h>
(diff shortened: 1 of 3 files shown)


jdekozak:
Hi,

It is now pending for almost 2 years, what should i do to get this patch merged into main tree ?

Regards,

jdekozak:
Hi,

Muse sequencer include directive for LASH can be found here
https://github.com/muse-sequencer/muse/blob/706f84d71dae4a66f0023ecba3f4cfb240c6b8ca/muse3/muse/main.cpp#L65

```
#ifdef HAVE_LASH
#include <lash/lash.h>
#endif
```

Regards,

mauser:
Thanks, merged finally!

PR #191, opened 2014-08-05
hydrogen.cpp: Added fallback in case desired MIDI driver cannot be found...

milasudril:
I got a problem when there is no support for the default MIDI driver upon first startup. This change looks through all available MIDI drivers, given a currently fixed priority list.

### DIFF
--- a/src/core/src/hydrogen.cpp
+++ b/src/core/src/hydrogen.cpp
@@ -205,6 +205,18 @@ void		audioEngine_restartAudioDrivers();
 void		audioEngine_startAudioDrivers();
 void		audioEngine_stopAudioDrivers();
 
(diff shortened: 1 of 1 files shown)


trebmuh:
@milasudril and @mauser : any news on this one? Does that need testing? If so, what could be a use-case I could test?

milasudril:
Disable the default MIDI driver when configuring hydrogen. (Was it ALSA, I do no remember now). It should choose another instead. Before this change,Hydrogen starts without MIDI support in that case.

trebmuh:
OK, so it's about a configure option before building/compiling Hydrogen if I got it right. Thanks.

theGreatWhiteShark:
Superseded by #1397

theGreatWhiteShark:
Thanks for the input and sorry it took so long to incorporate it

PR #217, opened 2015-01-06
Fix h2player to use applicationDirPath for data

dpurgin:
Hi, 

I've ran into this when using h2player without installation. It couldn't read any data relative to its executable location because of late QApplication instantiation.

Cheers
Dmitriy

h2player: QApplication instantiation moved before H2Core::Filesystem::bootstrap to satisfy applicationDirPath() call


mauser:
Thanks for the fix, merged!

PR #218, opened 2015-01-06
Fix audioEngine_startAudioDrivers() to deal with PulseAudio

dpurgin:
Hi,

I've built hydrogen with PulseAudio driver only, tried to run it with h2player but it turned out that PA is not probed if Auto driver is set in preferences.

Cheers
Dmitriy


mauser:
Thanks a lot!

PR #219, opened 2015-01-07
h2player: instantiate MidiMap before Preferences

dpurgin:
Hi,

Uninitilized MidiMap instance is attempted to be used in Preferences ctor when running h2player on a system without hydrogen configs (i.e. a system that didn't see hydrogen before or hydrogen GUI wasn't ever launched). Hydrogen tries to recreate config, assertion fails at MidiMap::get_instance().

Cheers


mauser:
Thanks a lot!

PR #228, opened 2015-02-15
Set configuration with cmake

jdekozak:
Hi,

This pull request will let the user choose the configuration with cmake without changing a header file.

Regards,
Julien


mauser:
Thanks!!

Issue #246, opened 2015-04-06
HiHat instruments


blablack:
Instead of having a boolean for hihat instrument (for V-Drum, to associate HiHat instruments with how open the hihat pedal is) it would be better to have a hihat group with an integer (similar to mute group) as many V-Drum would support top/edge hihat)

blablack:
After discussion with @mauser it was decided that https://github.com/hydrogen-music/hydrogen/issues/253 was not correct as it contained changes for another features.

The patch attached fixes this issue

```
From 696e7565ce830b9638da75f56ebcc6f9d6b75376 Mon Sep 17 00:00:00 2001
From: Aurelien Leblond <blablack@gmail.com>
Date: Sat, 8 Aug 2015 15:39:56 +0100
Subject: [PATCH] HiHat instrument group

---
 data/xsd/drumkit.xsd                               |   2 +-
 src/core/include/hydrogen/basics/instrument.h      |  16 +++---
 src/core/src/IO/midi_input.cpp                     |   4 +-
 src/core/src/basics/instrument.cpp                 |  10 ++--
 src/core/src/basics/song.cpp                       |   6 +++
 src/core/src/helpers/legacy.cpp                    |   2 +-
 src/core/src/local_file_mgr.cpp                    |   3 ++
 src/gui/src/InstrumentEditor/InstrumentEditor.cpp  |  58 +++++++++++++++------
 src/gui/src/InstrumentEditor/InstrumentEditor.h    |   6 ++-
 9 files changed, 72 insertions(+), 35 deletions(-)

diff --git a/data/xsd/drumkit.xsd b/data/xsd/drumkit.xsd
index 007a63a..855eac3 100644
--- a/data/xsd/drumkit.xsd
+++ b/data/xsd/drumkit.xsd
@@ -85,7 +85,7 @@
             <xsd:element name="midiOutChannel"   type="xsd:integer"     default="-1" minOccurs="0"/>
             <xsd:element name="midiOutNote"      type="xsd:integer"     minOccurs="0"/>
             <xsd:element name="isStopNote"       type="h2:bool"         default="false" minOccurs="0"/>
-            <xsd:element name="isHihat"          type="h2:bool"         default="false"/>
+            <xsd:element name="isHihat"          type="xsd:integer"     default="-1"/>
             <xsd:element name="lower_cc"         type="xsd:integer"     default="0"/>
             <xsd:element name="higher_cc"        type="xsd:integer"     default="0"/>
[...]

PR #253, opened 2015-04-11
Hihat pressure group

blablack:
Implements this DRQ:
https://github.com/hydrogen-music/hydrogen/issues/246


mauser:
Hi Aurelien! 

As already discussed via mail, i will close this pull request now, because it has also a lot of commits in it which do not belong to this pull request. Please create a new branch on your repository (based on hydrogen's main branch and use git cherry pick to select the corresponding high hat pressure group commits. You can also send me an patch via mail..

PR #260, opened 2015-06-04
Hydrogen Windows Cross Compile Scripts.

mikotoiii:
These are the changes that I've done for the Cross Compiling of Hydrogen for Windows. This has so far only been tested on my system, so some additional testing may be required. The cross_compile script will download and install the necessary environment, and I tried to make it as friendly as I could.

I have also put up a packaged version of the installer on my dropbox and I can share that link with the devel list if you want.

Also for some reason my github gui was logging my changes as unknown, even though I had to log in as myself, so all those unknown entries are me (still fairly new to github).


mauser:
Hi! 

If i remember correctly, you stated on our mailing list that this pull request is outdated. Is this still the case? If it is, then i would like to close it.. 

Is there sth. i can do to help with testing your cross compilation scripts?

mikotoiii:
Hey Mauser, 

I merged the changes from upstream into my repository, so everything should be up to date now.

I also quickly reviewed the README.me file in the windows directory which has the cross compile by hand instructions, and those look like they should be fine though I will perform a manual install with them as soon as I have time to verify. It's a little different than what the script does, but should still work.

I have also updated the mxe install script via instructions from the mxe site, and it should install mxe for just about every platform they support (most *nix, and apple).

If you would like to test the cross-compile that would be great.

mauser:
Hey,

i've reviewed the pull request yesterday/today and after some changes i was able to cross-compile hydrogen successfully! Thanks for your work! I will push my changes tomorrow and write a mail to the mailing list so that we can discuss the changes..

PR #263, opened 2015-06-27
Enable C++11 support in CMakeLists

elpescado:
I added `-std=c++11` switch to CMakeLists to enable support for C++11. This also adds `-Wno-deprecated-register` to silent warnings emitted by Qt headers.

This patch has been tested on Debian unstable (gcc 4.9.2 and clang 3.4.2) and on Mac OS X 10.10 Yosemite (clang 3.6.0svn). I have no access to Windows machine right now.

I have created this patch to check if anyone experiences any problems with compilers that do not support C++11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,10 +97,9 @@ ELSE()
 ENDIF()
 
 #SET(CMAKE_CXX_FLAGS "-fno-implement-inlines -finline-small-functions -findirect-inlining -fpartial-inlining")
-IF(APPLE)
-	SET(CMAKE_CXX_FLAGS "-O2")
-ELSE()
-	SET(CMAKE_CXX_FLAGS "-O2 -fno-implement-inlines")
+SET(CMAKE_CXX_FLAGS "-O2 -std=c++11 -Wno-deprecated-register")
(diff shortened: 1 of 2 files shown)


mauser:
Hi! 

Just checked this yesterday with the new windows cross compile script, and there is currently some problem when building hydrogen with mingw/g++ for windows. If one enables c++11 support, the WIN32 macro seems to be undefined. Quite suprising, but i had no time to look further into the issue yet...

Shaac:
As of CMake 3.1, there is a `CXX_STANDARD` property to be set, that enable c++11 (or 14), without playing with the `CXX_FLAG` directly. https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html

CMake philosophy is letting it add the flags, as there are compiler-depend.

That being said, it raises the dependency to CMake 3.1, and for instance it is not in the official repositories of the latest Ubuntu LTS, as it was released only a year ago. So you may not want to use this feature. I just wanted to point out that it exists.

mauser:
Hi! 
I've been looking into this one today (after looking at #356) and got the mingw support working (hopefully). I will merge and close this pull request now and add my mingw addition later.

PR #288, opened 2015-09-04
Win 64 Cross Compile

mikotoiii:
Here are the changes for 64 bit cross compilation. This will generate either a 32-bit installer, or a 64-bit installer.

This will also add the hydrogen.exe.manifest into the exe for UAC permissions to elevate Hydrogen to allow installation of drumkits, patterns, etc. since they are installed in the root Program Files/Hydrogen directory.


mauser:
Thanks again! I've tested it on my systems and it seems to work well so far.. Soundlibraries can be imported without problems.

Issue #267, opened 2015-07-27
Instrument Component / Banks are not undo-able


mauser:
If you drag an instrument with component b to a drumset with only component a, you can end up with a drumset with components a and b. If you're now using the undo functionality, only the instrument will be removed, but the component b stays in the drumkit.

blablack:
Here is a patch that fixes this requirement.

```
From af5d52936a8e67a7b955b8e3c9ade086e699c739 Mon Sep 17 00:00:00 2001
From: Aurelien Leblond <blablack@gmail.com>
Date: Tue, 1 Sep 2015 21:38:06 +0100
Subject: [PATCH] Remove additional bank when undoing drag/drop instrument

---
 src/gui/src/PatternEditor/DrumPatternEditor.cpp | 20 ++++++++++++++++++--
 src/gui/src/PatternEditor/DrumPatternEditor.h   |  4 ++--
 src/gui/src/UndoActions.h                       |  8 +++++---
 3 files changed, 25 insertions(+), 7 deletions(-)

diff --git a/src/gui/src/PatternEditor/DrumPatternEditor.cpp b/src/gui/src/PatternEditor/DrumPatternEditor.cpp
index 2939f68..10dc803 100644
--- a/src/gui/src/PatternEditor/DrumPatternEditor.cpp
+++ b/src/gui/src/PatternEditor/DrumPatternEditor.cpp
@@ -1241,11 +1241,25 @@ void DrumPatternEditor::functionMoveInstrumentAction( int nSourceInstrument,  in
 }
 
 
-void  DrumPatternEditor::functionDropInstrumentUndoAction( int nTargetInstrument )
+void  DrumPatternEditor::functionDropInstrumentUndoAction( int nTargetInstrument, std::vector<int>* AddedComponents )
 {
 	Hydrogen *pEngine = Hydrogen::get_instance();
 	pEngine->removeInstrument( nTargetInstrument, false );
 
+	std::vector<DrumkitComponent*>* pDrumkitComponents = pEngine->getSong()->get_components();
+
+	for (std::vector<int>::iterator it = AddedComponents->begin() ; it != AddedComponents->end(); ++it) {
+		int p_compoID = *it;
+
+		for ( int n = 0 ; n < pDrumkitComponents->size() ; n++ ) {
+			DrumkitComponent* pTmpDrumkitComponent = pDrumkitComponents->at( n );
+			if( pTmpDrumkitComponent->get_id() == p_compoID ) {
[...]

PR #293, opened 2015-09-13
Undoable banks

blablack:
This pull request fixes: https://github.com/hydrogen-music/hydrogen/issues/267

After an instrument is dragged/dropped on the drumkit and a bank is added, when the user is trying to undo it, the bank is removed.


mauser:
Hi Aurelien,

just tested the pull request and i'm running into an assertion (InstrumentEditor line 567). I'm just adding an instrument with component "B" to a drumkit which has instruments of component "A" with the "Merge" functionality and using undo (ctrl+z).

blablack:
Hi @mauser 

I can't reproduce it:
- I tried with instrument with component "A" & "B" into drumkit with only "A"
- I tried with instrument with component "B" into drumkit with only "A"

In both cases undoing went fine.

Can you re-test again? maybe something i'm missing from your scenario?

mauser:
Ah, retested and found the difference: You have to drag the instrument to the last position. But this error seems to occur also without your last changes if you drag an instrument to the end of the instrument list (this is important) and then do the undo. 

But back to this issue: If you're doing an undo while the layer tab is activated and show the new component (the one that will be deleted during undo), hydrogen will crash

blablack:
@mauser Issue occurs as well with moving an instrument inside a drumkit.
Move it much fiurther than the last position, try to undo and Hydrogen will crash.

I found out what the problem is, I just need to find a solution:
- means this pull request is valid, the bug is unrelated
- I'll create a bug report for it
- I'll fix it in another branch

mauser:
@blablack : Wow, that was fast. Thanks! I will be testing the new pull request later.. Btw, have you been able to reproduce the second problem i mentioned above? Or will this also be fixed by the other pull request? 

"But back to this issue: If you're doing an undo while the layer tab is activated and show the new component (the one that will be deleted during undo), hydrogen will crash"

blablack:
@mauser : I kind of been able to reproduce it (in my case, Hydrogen doesn't
crash but the component is not removed). It wont be fix by the other pull
request (unrelated problem) - I'll need to fix it in this pull.

blablack:
@mauser: fixed :)

mauser:
Thanks again, everything merged now! I've just added a call to "	EventQueue::get_instance()->push_event( EVENT_SELECTED_INSTRUMENT_CHANGED, -1 );" to refresh the instrument editor after a "drop instrument" redo. Btw, please check your indentation settings :)

PR #302, opened 2015-09-20
Renamed bank to component

blablack:
The last traces of "bank" in the GUI/code should now be gone.
It's "component" all thw way!


mauser:
Thanks a lot!

Issue #305, opened 2015-09-27
Undo move/drag instrument to last position crashes Hydrogen


blablack:
There are 2 scenarios to reproduce the bug:
1. move (with the mouse) an instrument to way lower than the last position (keep going even if nothing happens)
2. drag/drop an instrument to way lower than the last instrument

Try to undo, Hydrogen will crash:
```
void H2Core::InstrumentList::move(int, int): Assertion `idx_a >= 0 && idx_a < __instruments.size()' failed
```

PR #306, opened 2015-09-27
Bug fix for https://github.com/hydrogen-music/hydrogen/issues/305

blablack:
In a nutshell, when moving or drag/dropping, if going to far away Hydrogen
will put the instrument back to last position


mauser:
Thanks Aurélien!

Issue #307, opened 2015-09-28
Exported MIDI files contain only Hi Bongo note on/note off events.


dawookieuk:
Current Master branch of 0.9.7 generates MIDI files which have all note events as instrument number 60 which corresponds to the GM Midi Kit Hi Bongo instrument.

All the below was performed on Debian Linux 7.9 'Wheezy' and replicated in OS X 10.9 'Mavericks'

For testing, I cloned the 0.9.6 and 0.9.7 branches into separate folders and built them with ."/build.sh m"

I opened the executable for 0.9.6 and generated and saved the test file "Test Scale.h2song".

Still within the 0.9.6 application I exported the song to a MIDI file named "0.9.6-Master.mid"

I closed the 0.9.6 application.

I opened the 0.9.7 application.

I within the application I loaded the file "Test Scale.h2song" and then exported the song to a MIDI file named 0.9.7-Master.mid.

I closed the 0.9.7 application.

I installed the python-midi package from https://github.com/vishnubob/python-midi

I dumped the contents of the MIDI file 0.9.6-Master.mi

d to a text file called 0.9.6-Master.mid.txt with the line: mididump.py 0.9.6-Master.mid > 0.9.6-Master.mid.txt

I dumped the contents of the MIDI file 0.9.7-Master.mid to a text file called 0.9.7-Master.mid.txt with the line: mididump.py 0.9.7-Master.mid > 0.9.7-Master.mid.txt

Viewing the two files it is apparent that 0.9.6 is generating the note events correctly and that 0.9.7 does note, all not events have a value of 60 for the instrument number.

I apologise for the inline texts but the site would allowed me to upload.

0.9.6-Master.mid.txt Contents:

midi.Pattern(format=1, resolution=192, tracks=\
[midi.Track(\
  [midi.TrackNameEvent(tick=0, text='Hydrogen song!!', data=[72, 121, 100, 114, 111, 103, 101, 110, 32, 115, 111, 110, 103, 33, 33]),
   midi.NoteOnEvent(tick=0, channel=9, data=[36, 101]),
   midi.NoteOffEvent(tick=48, channel=9, data=[36, 101]),
   midi.NoteOnEvent(tick=48, channel=9, data=[37, 101]),
   midi.NoteOffEvent(tick=48, channel=9, data=[37, 101]),
   midi.NoteOnEvent(tick=48, channel=9, data=[38, 101]),
[...]

Issue #124, opened 2013-10-09
Hydrogen creates midifile which has an unchangeble tempo in Roland's drumtutor program


ThomasDrummer:
Hi,
I am a user of the Roland HD-1 drumtutor program. When I created a drumtrack on Hydrogen I saved it as a midifle and then imported in the HD-1 Drumtutor program. The tempo 120 BPM couldn't be changed then.  I solved it more or less by first importing the midifile in the program:  Midieditor and then without any change saving it with that program. After that I loaded the midifile in the Drum tutor program and the tempo could be changed. Also Audacity did the same trick with importing the midile and exporting it again. How is that possible?

thijz:
ThomasDrummer, xjjx, 
any update on this ?

grtz
Thijs

xjjx:
I still can't find this :( Cannot confirm or deny this issue. At least not yet.

dawookieuk:
I can confirm this behaviour. My setup is OS X 10.9 running hydrogen 0.9.6 and Roland Drum Tutor v 1.0.1.

I have forked a repo and modified the MIDI file writing in the 0.9.6 branch and the MIDI files now import into Roland Drum Tutor with the song name being extracted from the Song Properties dialog as displayed by the Project->Show Info menu item. The rest of the MIDI file is now much more compliant with the SMF standard but that may not bother you too much. ;-)

If some nice people could pull from my repo and test (currently the 0.9.6 branch, not master) and confirm a fix I'll generate a pull request.

mauser:
Hi! 

I can't test it (i don't own the software) but i had a quick look at your branch. If i understand your changes correctly, the problem was that in our files, there was no tempo meta track?

dawookieuk:
Hi. 

That's correct. The only meta event being written to the file was the Track Name event. 

From the perspective of SMF compliance there was another issue, which is that there was only a single track. From the reading I've done I've the last week or so, it appears that a format 1 SMF file should have the first track chunk in the file followed immediately by a copyright notice meta event then (in no particular order) Set Tempo, Time Signature and Track Name meta events.  There should then be a track end chunk, a new track chuck and then any note on/note off events which arevthen terminated with a track end chunk.

The master branch code was producing a single track chunk with Track Name as the only meta event, followed by note on/note off events and a track end chunk. This structure doesn't appear to be compliant with the Standard Midi Format documentation.

[...]

PR #308, opened 2015-09-28
Produce Standard MIDI Format (SMF) compliant MIDI files

dawookieuk:
Resolves these issues:

https://github.com/hydrogen-music/hydrogen/issues/307
https://github.com/hydrogen-music/hydrogen/issues/124

Correct a bug introduced in 0.9.7 which caused all
note on/note off events to be written with a note number of
60 irrespective of the note in the song.

Modify SMF.h, SMFEvent.h, smf_event.cpp, & smf.cpp to generate the MIDI
Tempo track as the first track and populate it with the Copyright Notice
,Time Signature & Set Tempo meta events. The Track Name meta event is
moved to the tempo map.

The name in the Copyright Notice meta event will be extracted from the
Song properites as filled in at the Project->Show Info dialog, which
will be preceded with the copyright (C) symbol and followed with the
current year as specified in the SMF standard. This will be the first
event in the file, also as specified in the SMF standard.

The Track Name meta event will now be extracted from the Song properites
as filled in at the Project->Show Info dialog and will be located in the
tempo map as per the SMF standard.

The Time Signature meta event will be created assuming 4/4 time.

The note on & note off events will start in the second track in the
file.

### DIFF
--- a/src/core/include/hydrogen/smf/SMF.h
+++ b/src/core/include/hydrogen/smf/SMF.h
@@ -55,7 +55,8 @@ class SMFTrack : public SMFBase, public H2Core::Object
 {
 	H2_OBJECT
 public:
-	SMFTrack( const QString& sTrackName );
+
+	SMFTrack();
 	~SMFTrack();
 
 	void addEvent( SMFEvent *pEvent );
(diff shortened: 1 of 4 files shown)


dawookieuk:
Issue #307 a false positive with an ongoing discussion, the observed behaviour is actually by design. I have restored the default behaviour in the latest commit.

Issue #124 is still resolved by this fork.

mauser:
Thanks! I've merged your commits now after testing them.

PR #313, opened 2015-10-22
Export to older version of Hydrogen

blablack:
This functionality enables a user to export to older version of Hydrogen.
For a version under 0.9.6, user will have to chose the component they which to export.


src/gui/src/SoundLibrary/SoundLibraryExportDialog.cpp
```diff
@@ -175,6 +232,32 @@ void SoundLibraryExportDialog::on_browseBtn_clicked()
...
+	for (QStringList::iterator it = p_compoList.begin() ; it != p_compoList.end(); ++it) {
+		QString p_compoName = *it;
+
+		componentList->addItem( p_compoName );
+	}
+}
+
+void SoundLibraryExportDialog::on_versionList_currentIndexChanged( QString str )
+{
+	if( str.toStdString() == "0.9.7 and higher" )
```
mauser: Those strings should be made translatable (in the gui..). So it would be better to use trUtf8() or to use the combo box indices for comparisons.

mauser:
Hi! 

A first test fails at my system, i always get the 0.9.7 xml file in the drumkit package, no matter what the version combo box says. I can debug it further in the next days..  But the gui is fine so far!

src/gui/src/SoundLibrary/SoundLibraryExportDialog.cpp
```diff
@@ -81,6 +85,28 @@ void SoundLibraryExportDialog::on_exportBtn_clicked()
 	QString drumkitDir = Filesystem::drumkit_dir_search( drumkitName );
 	QString saveDir = drumkitPathTxt->text();
 
+	int componentID = -1;
+	Drumkit* info;
+	if( versionList->currentText().toStdString() == "0.9.6 and lower" ) {
+		for (uint i = 0; i < drumkitInfoList.size(); i++ ) {
+			info = drumkitInfoList[i];
+			if( info->get_name().compare( drumkitName ) == 0 ) {
+				QString temporaryDrumkitXML = QString("/tmp/hydrogen/drumkit.xml");
```
mauser: This won't work for systems which don't have a unix file system layout. There is a QT helper function which gives you an appropriate temporary path, we are using that to implement Preferences::get_instance()->getTmpDirectory()

blablack:
@mauser corrected what you reported.

mauser:
Hi! 
i'm still experiencing a strange problem: On my ubuntu computer, the 0.9.7 version gets always exported. On my laptop running OS X, everything works fine and i can export 0.9.6 files. I will see if this might be a permission issue or sth. similar and report back.

blablack:
One possible explanation is that Hydrogen is not compiled with libarchive and the alternative method is used.
In the alternative method, a simple shell method is used to tar the content of the drumkit folder.

I tried to modify this part of the code as well but never managed to compile Hydrogen without libarchive. To me it looked like it was mandatory and the code that manage if libarchive is not installed was just legacy.

mauser:
Good hint, that's exactly the problem :) I'm creating a fix for that problem..

Issue #285, opened 2015-09-03
DrumkitComponents: SampleEditor's "Play" function previews the complete instrument, not the current sample


mauser:
If you have an instrument with multiple components and you try to edit the sample of one component, the "Play" functionality of the SampleEditor previews the complete instrument with all its components. In  my opinion, the SampleEditor should operate only on the Sample of the current component.

PR #320, opened 2015-10-31
Sample Editor and Layer click play all component instead of one selected

blablack:
This fixes this bug https://github.com/hydrogen-music/hydrogen/issues/285 in addition to clicking on a layer itself (the fix was the same for both bugs).


mauser:
Thanks @blablack , works as expected :+1:

PR #322, opened 2015-10-31
Add an appdata to Hydrogen

blablack:
Add an appdata to Hydrogen as requested in https://github.com/hydrogen-music/hydrogen/issues/178

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ moc_*.cc
 /hydrogen
 *.cbp
 *.layout
-hydrogen.*
+./hydrogen.*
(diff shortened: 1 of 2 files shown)


mauser:
Merged! But i will exchange the mail address for the hydrogen-devel mailing list address and maybe create new screenshot directory on the server. In addition, i suppose we have to install the file with cmake on linux systems. I will check that when i'm back at a linux computer..

blablack:
@mauser adding this to the CMakeList.txt should do it

```
INSTALL(FILES ${CMAKE_SOURCE_DIR}/linux/hydrogen.appdata.xml DESTINATION "${CMAKE_INSTALL_PREFIX}/share/appdata")
```

It should go in that block
```
IF(NOT MINGW AND NOT APPLE)
    INSTALL(FILES ${CMAKE_SOURCE_DIR}/linux/hydrogen.desktop DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
ENDIF()
```

Sorry for not creating a branch, thought that was a bit overkill :)

mauser:
Thanks, done!

Issue #324, opened 2015-11-02
Size and Resolution dropboxes continuously reopen


pvint:
Since revision 36a2268 the Size and Resolution dropboxes reappear after making a selection.

It seems to be related to the timing code in that commit - commenting the "m_timer->start" out in the following method in src/gui/src/widgets/Button.cpp prevents the problem (but also negates the new timer feature):

void Button::buttonPressed_timer_timeout()
{
        emit clicked(this);

        if(m_timerTimeout > 50)
                m_timerTimeout = m_timerTimeout / 2;
        m_timer->start(m_timerTimeout);
}

blablack:
I will try to have a look at this bug this evening.

Aurélien

pvint:
Looked at the commit - looks good, and I thought it would do it, but the problem persists (for me, anyway. I'm testing running KDE Plasma 5.4.1, and when the dropdown appears, it works and I can select a different resolution, but it pops right back up and I actually cannot even use any other windows as it keeps the focus. I have to CTRL-ALT-F2 to a tty console and kill hydrogen)

blablack:
I retested it and I can't reproduce it anymore - as far as I can tell the logic (signal emissions) for the non "press&hold" buttons should be back to what it used to be.

Can you clean and recompile maybe? Could someone else test it?

pvint:
Tested this morning on a fresh git clone (even tried from your fork to be sure) and have the same thing.

To see better what is happening I added a bit of logging:
void Button::mouseReleaseEvent(QMouseEvent* ev)
{
        setPressed( false );
WARNINGLOG("void Button::mouseReleaseEvent(QMouseEvent* ev");
        if (ev->button() == Qt::LeftButton) {
                m_timer->stop();
        }
        else if (ev->button() == Qt::RightButton) {
                emit rightClicked(this);
        }

}


void Button::buttonPressed_timer_timeout()
{
        emit clicked(this);
WARNINGLOG( QString::number( m_timerTimeout) +  " Button::buttonPressed_timer_timeout()");
        if(m_timerTimeout > 50)
                m_timerTimeout = m_timerTimeout / 2;
        m_timer->start(m_timerTimeout);
}



With it logging like this, I ran hydrogen, clicked and held the Stop button for a second or so, then clicked and released the Resolution button, here's the relevant part of the output:

(W) Button::buttonPressed_timer_timeout 31 Button::buttonPressed_timer_timeout()
(I) AlsaAudioDriver::stop stop
[...]

PR #327, opened 2015-11-02
Button press hold bug fix

blablack:
This pull request fixes bug https://github.com/hydrogen-music/hydrogen/issues/324

Ok so with the way I was emitting the clicked signal, when a QMenu displays its menu with the popout method mouse release wasn't triggered.

I thought about trying to solve the issue with the QMenu themselves but it would have forced each developer to apply the same workaroud. Instead I now pass as argument of the button if it will use the press&hold option amd decide to use a QTimer at all.

With the same pull request, I slowed down the rate and use the same methods for the Tempo buttons (they were using the same method but it was coded in the class itself and didn't implement the acceleration).


blablack:
TBH, i'm not sure.
Did you check this out?
http://stackoverflow.com/questions/6743514/how-can-i-fetch-an-unmerged-pull-request-for-a-branch-i-dont-own

Issue #212, opened 2014-12-08
add image to drumkit properties


thijz:
adding a picture to the drumkit properties makes it much more appealing for the user to use
something like this would be nice : 

![mockup_image_in_drumkit_library](https://cloud.githubusercontent.com/assets/1384210/5347703/77ec8efa-7f2a-11e4-8537-517c4b7e2764.jpg)
![mockup_image_in_drumkit_properties](https://cloud.githubusercontent.com/assets/1384210/5347704/77ecbe3e-7f2a-11e4-84a5-89f754f3e762.jpg)

pvint:
Is this being worked on by anyone?

I've been wanting something do to contribute, and this sounds like a reasonably easy way to get my feet wet!

thijz:
hi pvint
there is nobody working on this one, so go ahead and get your feet wet :-)

before you get started it's probably best to check out how drumkuts are packaged 
basically these are tgz files that contain all samples + an XML that describes the drumkit.  so i guess the picture would also have to go in that tgz

some time ago i created a py script that creates drumkits, so this might give you some insight on the way drumkits are structured (see http://audio-and-linux.blogspot.com/2012/10/hydrogen-drumkit-creator-script-v5-is.html)

important note : it's best to clone the hydrogen git repo and create small patches for what you are working on.  if anything would break it is easier to track down where things went wrong (maybe of this particular feature it wont be necessary to split up, but for other, bigger, features it it the way to go)

all info on getting the code and building it can be found on the wiki : 
https://github.com/hydrogen-music/hydrogen/wiki

good luck and dont hesitate to ask questions (best via the mailing list)

grtz
Thijs

pvint:
I have this partially done, I added an <image> tag to the xml with the filename of the image, and it displays in the properties dialog much like your image above, and is copied to a new drumkit correctly when the kit is saved under a new name.

It is *not* in the Instruments->Save dialog yet, nor in the library browser, but I thought I'd let you have a look before I go too much further. 

You can have a look at the diffs at https://github.com/hydrogen-music/hydrogen/compare/master...pvint:master or I can do a pull request if you'd like.

Also note: The TR-808 drumkit is updated and it get's rid of the XML errors.

Question: Is the source for feeds/drumkit_list.php available somewhere?

thijz:
Hi Paul
Yesterday i built hydrogen from your fork but i didnt have time to play
around with it yet
I'll have a go during my lunch break

Grtz
Thijs
[...]

Issue #333, opened 2015-11-20
Drumkit->info bad html conversion


pvint:
When saving a drumkit from H2 the "Info" text gets saved as HTML, but there's a bad HTML conversion. Rather than explain, here's an example, using Millo-Drums_v.1:

1. Import the kit via Instruments->Import Library
2. Once the kit is installed, view the ~/.hydrogen/data/drumkits/Millo-Drums_v.1/drumkit.xml file and you should see the following for the info element:
```html
<info>This drumkit was created recording my ludwig set, trying to obtain a&#x0A;&quot;sixties&quot; sound. Recorded with ecasound and edited with rezound</info>
```
3. In Sound Library, right-click on that kit and select Properties.
4. Click "save"
5. View the file ~/.hydrogen/data/drumkits/Millo-Drums_v.1/drumkit.xml again, and in my case I see:
```html
<info>&lt;!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">
&lt;html>&lt;head>&lt;meta name="qrichtext" content="1" />&lt;style type="text/css">
p, li { white-space: pre-wrap; }
&lt;/style>&lt;/head>&lt;body style=" font-family:'Lucida Grande'; font-size:10pt; font-weight:400; font-style:normal;">
&lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">This drumkit was created recording my ludwig set, trying to obtain a&lt;/p>
&lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">&amp;quot;sixties&amp;quot; sound. Recorded with ecasound and edited with rezound&lt;/p>&lt;/body>&lt;/html></info>
```

In particular, note at the end the `&lt;/html>` - note that the `<` gets converted but the `>` does not.

I've tested back to 9.5.1 and the problem is the same. I've also put tons of logging in to try to find where it happens, but was unable to quickly find the source. I can continue looking, but I thought I was getting too distracted and should be looking at finishing that other feature first. ;)

pvint:
[...]

PR #334, opened 2015-11-20
Add Images to Drumkit

pvint:
Ref: issue #212 

See issue for full details - use http://dotslash.ca/h2/list.xml for testing (the xml is a bit rough - no info on kits, a couple images missing, so ignore any image errors).

There may be changes we want to make to the way kits are named/stored, but I think that should be handled separate from this (this "simple addition" got bigger than I thought!)

Thanks to @thijz for all the help and testing on this!

Cheers
Paul


data/drumkits/TR808EmulationKit/drumkit.xml
```diff
@@ -8,6 +8,14 @@ p, li { white-space: pre-wrap; }
 &lt;/style>&lt;/head>&lt;body style=" font-family:'Lucida Grande'; font-size:10pt; font-weight:400; font-style:normal;">
 &lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">This drum kit contains 16 original sounds that were synthesized from scratch (using only basic synth waves, like sine, noise etc.). It simulates the sounds of the legendary Roland TR-808 Rhythm Composer. More kits from ArtemioLabs: http://artemiolabs.com/downloads/audio/hydrogen/&lt;/p>&lt;/body>&lt;/html></info>
```
pvint: Note: This line has the problem with HTML noted in issue #333 
I just noticed that after the pull request.

PR #339, opened 2015-12-01
Add drumkit images

pvint:
(Re-submitting this request - the old request was cancelled when I deleted my "master" branch when tidying up my branches)

Ref: issue #212

See issue for full details - use http://dotslash.ca/h2/list.xml for testing (the xml is a bit rough - no info on kits, a couple images missing, so ignore any image errors).

There may be changes we want to make to the way kits are named/stored, but I think that should be handled separate from this (this "simple addition" got bigger than I thought!)

Thanks to @thijz for all the help and testing on this!

Cheers
Paul


data/drumkits/TR808EmulationKit/drumkit.xml
```diff
@@ -2,12 +2,16 @@
...
- <info>&lt;!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">
-&lt;html>&lt;head>&lt;meta name="qrichtext" content="1" />&lt;style type="text/css">
-p, li { white-space: pre-wrap; }
-&lt;/style>&lt;/head>&lt;body style=" font-family:'Lucida Grande'; font-size:10pt; font-weight:400; font-style:normal;">
-&lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">This drum kit contains 16 original sounds that were synthesized from scratch (using only basic synth waves, like sine, noise etc.). It simulates the sounds of the legendary Roland TR-808 Rhythm Composer. More kits from ArtemioLabs: http://artemiolabs.com/downloads/audio/hydrogen/&lt;/p>&lt;/body>&lt;/html></info>
+ <info>This drum kit contains 16 original sounds that were synthesized from scratch (using only basic synth waves, like sine, noise etc.). It simulates the sounds of the legendary Roland TR-808 Rhythm Composer. More kits from ArtemioLabs: http://artemiolabs.com/downloads/audio/hydrogen/</info>
  <license>GPL</license>
+ <image>Roland_TR-808_drum_machine.jpg</image>
```
mauser: It would be nice to have the imageLicense set as well. Btw, it might be problematic for us if the image is not licensed under a GPL (or GPL compatible) license. Is this the CC BY-SA 3.0 licensed image from wikipedia??

src/gui/src/SoundLibrary/SoundLibraryImportDialog.cpp
```diff
@@ -360,7 +433,40 @@ bool SoundLibraryImportDialog::isSoundLibraryItemAlreadyInstalled( SoundLibraryI
 	return false;
 }
 
+void SoundLibraryImportDialog::loadImage(QString img )
+{
+	QPixmap pixmap;
+	// TODO: Need to make sure that this isn't getting called after user clicks a different item!
```
mauser: Is this still a "valid" TODO? Is there sth. to be done here?

src/gui/src/SoundLibrary/SoundLibraryPropertiesDialog.cpp
```diff
@@ -56,20 +58,104 @@ SoundLibraryPropertiesDialog::SoundLibraryPropertiesDialog( QWidget* pParent, Dr
...
+	if ( fileName == NULL )
+	{
+		return;
+	}
 
+	// If this file is in different directory copy it here
+	
+	QFile file( fileName );
+	QFileInfo fileInfo(file.fileName());
+	//ERRORLOG(fileInfo.dir().path().toLocal8Bit() + drumkitDir);
```
mauser: This doesn't look intentional :)

data/drumkits/TR808EmulationKit/drumkit.xml
pvint: I'll verify to be sure - I forgot about this one, as I had updated it before adding the imagelicense tag.
Confirmed: it is at https://commons.wikimedia.org/wiki/File:Roland_TR-808_drum_machine.jpg and the license/usage info is:
`Eriq at Dutch Wikipedia [GFDL (http://www.gnu.org/copyleft/fdl.html) or CC-BY-SA-3.0 (http://creativecommons.org/licenses/by-sa/3.0/)], via Wikimedia Commons`
 I'll admit to not being 100% clear on how you can or cannot use GFDL or CC licensed stuff in GPL projects

src/gui/src/SoundLibrary/SoundLibraryPropertiesDialog.cpp
pvint: At a glance, it looks intentional to me, but it doesn't look like it will work! I'll have a better look :)

src/gui/src/SoundLibrary/SoundLibraryImportDialog.cpp
pvint: This was a "Done" TODO - removed the comments and committed.
BTW: This is something to watch for in my code... I have a bit of a bad habit of fixing things that I left a TODO comment on and forgetting to remove the TODO

src/gui/src/SoundLibrary/SoundLibraryPropertiesDialog.cpp
pvint: Took a look, and maybe I'm missing what you thought may not have been intentional... or maybe not....

If it's the `// If this file is in different directory copy it here` bit - the idea is that if a user loads an image from ~/Downloads/ or something, it will copy it into the drumkit's directory.

I just had a quick look, and I'm having that "I'm missing something obvious" feeling.... let me know. ;)

mauser: The "obvious thing" was that i was talking about the comment (ERRORLOG) :smile:   Sorry for the confusion, it was a totally minor thing.. 

What i meant was: Let's remove this ERRORLOG completeley if it was meant just for debugging or turn it into an INFOLOG if the information might be interesting for the user.

data/drumkits/TR808EmulationKit/drumkit.xml
mauser: For the CC license: 
  **attribution** – You must attribute the work in the manner specified by the author or licensor (but not in any way that suggests that they endorse you or your use of the work).
  **share alike** – If you alter, transform, or build upon this work, you may distribute the resulting work only under the same or similar license to this one.

So if i understand those lines correctly, then including the image without any attribution and mentioning of the license would be a clear violation (if one would choose the CC license). In my opinion we should definitely not put any images with such restrictions in our default kits (if my interpretation of the license is correct).

I'm not sure sure about the GFDL, have to investigate further.

src/gui/src/SoundLibrary/SoundLibraryPropertiesDialog.cpp
pvint: Aha! I WAS missing something obvious! I'll remove the ERRORLOG bit - it was just there for debugging.

data/drumkits/TR808EmulationKit/drumkit.xml
pvint: My understanding is the same, and I *believe* that so long as the license/attrubution is in the XML (which is displayed along with the image) that we're OK, but IANAL.

It sure would be nice to be able to use the wikipedia images, as it is a great resource for images, but until that's sorted out I think I'll just remove it.

pvint: #       deleted:    data/drumkits/TR808EmulationKit/Roland_TR-808_drum_machine.jpg
#       deleted:    data/drumkits/TR808EmulationKit/drumkit.xml

mauser:
After reading a little bit more, i think we have to somehow add a possibility for the "attribution". Maybe this could be done adding a imageAuthors tag.

 Another method would be to include the author in the imageLicense tag: "<imageLicense>GFDL Author: xyz</imageLicense>", but this is quite a dirty hack :) Especially because "Author" wouldn't be translatable by our translation tools.

mikotoiii:
I did a little reading on CC images and GPL software. Most of the things I
read said that even with the attribution set you can't use CC images in GPL
software due to the redistribution of said image which violates the CC
license. This was in particular reference to people wanting to use CC
images in GPL Wordpress themes. One thing it did mention is that CC0 does
allow redistribution so if we could find images with that, we might be good.

Not sure how accurate that is as I haven't sat down and read the whole CC
license. Just figured i'd throw it out there for thought.

mauser:
@mikotoiii : Do you know if this also applies to GFDL licensed stuff? 

@þvint: Could you write up some lines about setting up the server side of the this feature? I think will merge this soon, even though there are no images currently available.

mauser:
@pvint : Could you try to create a new "[squashed](https://help.github.com/articles/about-pull-request-merge-squashing/)" pull request for this issue?

mikotoiii:
@mauser : Although the GPL and GFDL are sister licenses, they might not necessarily be compatible with each other, CC is a different beast, though we may still be able to use it.

GPL: We're probably all familiar with the GNU Public License, but in effect it's what we use to distribute our software. We use the older 2.0 license if I recall. It can be found here: http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
GFDL: I was not overly familiar with this license, though in 1.2+ it seems to be more compatible with CC.it can be found here: http://www.gnu.org/licenses/old-licenses/fdl-1.2.html 
CC: Another license type we encounter very frequently. There are several different ones, and each have their own distinctions. https://creativecommons.org/licenses/

Wikipedia has a fairly decent explanation on its license, which can be found here: https://en.wikipedia.org/wiki/Wikipedia:Reusing_Wikipedia_content#Images_and_other_media. To sum it up, each image usually falls into the CC-BY-SA license, though images CAN be subject to other licenses. An example relevant to us is here with the Roland TR-808 image: https://en.wikipedia.org/wiki/File:Roland_TR-808_drum_machine.jpg
This basically says that if we attribute the image to Eriq at Wikipedia Dutch, we can use it.

There is a clause in the GFDL license that says anything uploaded to Wikipedia BEFORE 2008 can be used with CC-BY-SA 3.0.

From what it looks like to me, we could potentially adopt the CC license for our repo, especially since there is the option in the program to add other repos, which means that most of the drumkits are not included in the main application, which might not violate the license (I'm not a lawyer though). I should also point out that Debian had some issues with the GFDL license which makes it incompatible with its software: https://wiki.debian.org/DFSGLicenses.

So there are a couple of different ways that we can approach the situation as I see it. Our main program can remain under GPL 2.0, but we can change the license type on the repo for drumkits to something more compatible, since the repo isn't necessarily in the program, and the CC license may cover it if the attribution is put into the drumkit xml.
We can change the license on the application, which I'm pretty sure we don't want to do.
Another option is we can seek out the original author of the images, and ask permission to use the pictures in for our repo which will satisfy the GPL license. We could also approach people on our forums, other drum machine forums, or even sites like ebay, to seek permission to use the images that they post.

pvint:
@mauser  I took a quick look, and I can't see how to "squash" the pull request, but it appears as though you can accept it as squashed, I think? Again, I only took a quick look. I'll give it another look later tonight when I have more time.

@mikotoiii  Good comments, and I think the idea to simply having CC (other licensed) stuff only in the drumkit repo makes a lot of sense.

mikotoiii:
I'm pretty sure that all we'd have to do is on the import instrument page
have a note that the drumkits are released under cc-by-sa 3.0 and link to
the license. Here's a link to the article describing multi licensing
https://en.m.wikipedia.org/wiki/Multi-licensing

@mauser <https://github.com/mauser> I took a quick look, and I can't see
how to "squash" the pull request, but it appears as though you can accept
it as squashed, I think? Again, I only took a quick look. I'll give it
another look later tonight when I have more time.

@mikotoiii <https://github.com/mikotoiii> Good comments, and I think the
idea to simply having CC (other licensed) stuff only in the drumkit repo
makes a lot of sense.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/hydrogen-music/hydrogen/pull/339#issuecomment-217291203>

mauser:
@pvint : Sorry, my fault! The button appears only *after* the merge button has been clicked, so you can't see it in the pull request.

data/drumkits/TR808EmulationKit/drumkit.xml
```diff
@@ -1,525 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
mauser: Your pull request removes the drumkit.xml file. This seems to be erratic...

mauser:
@mikotoiii : I also think that we're ok if we keep the cc-by-sa images out of hydrogen's core but provide functionality to specify the license and the author in the import dialog. The license information can already be specified as a property of the drumkit, and the author could be named directly in the drumkit info property.

mauser:
Hi, 

the pull request has now been merged, thanks a lot! I'm going to write a small introduction for this feature which documents how the server-side has to look like...

pvint:
Awesome!  If any bugs or changes come up that needs to be addressed related to this, I'll be happy to take care of it, so if there are bug reports that I don't notice don't hesitate to drop me a line.

trebmuh:
Great!

@mauser and @pvint : I'm more than happy to help testing here (if it needs still some testing) if you can provide a step-by-step way for that.

Anyway: congratulations to you for this new great design improvement!

PR #354, opened 2016-02-13
update translation strings + update FR translation

trebmuh:
I
1) ran data/i18n/updateTranslations.sh,
2) made the FR translation in the .fr.ts file
3) ran again data/i18n/updateTranslations.sh

Hope that this is the correct way of doing so.


mauser:
Merged, thanks! It should not be necessary to run the Update script twice.

trebmuh:
Sweat. Thanks for the advice.

PR #359, opened 2016-02-28
Apply Velocity

blablack:
# Apply Velocity

"Apply velocity" is a new functionality for Hydrogen that let the user decide if Hydrogen will apply the note velocity to the sample being played.

screenshot

When activated, Hydrogen will apply the note velocity to the sample gain in addition to the layer gain, the instrument gain, the component gain or the instrument volume.
By default this option will be selected as this is the way older version of Hydrogen used to work.

When not activated, the note velocity will only be used to select the sample to be played, but the sample gain will not be changed. 
This is useful for set of samples that already have their gain "hard-coded".


blablack:
Would you guys have some feedback here?

mauser:
Merged, thanks a lot Aurelien!

trebmuh:
@blablack : I've built Hydrogen from GIT as of today (ie: https://github.com/hydrogen-music/hydrogen/commit/d7dacfedf73ede1af24ddedf260cfce5ca77df18) which includes this new feature. That said, I'm not 100% sure to understand what's the goal of this feature (even reading 10 times at the information you provided above, which is great, thanks! ). Would you mind trying to re-explain me that (maybe French could be easier in this case) ?
Thanks.

blablack:
@trebmuh 
So you have to imagine 2 scenarios:

1. You have a set of samples for a snare that are all normalized. If you tick "apply velocity", when you have a beat in Hydrogen at full velocity, it will be played at full gain, and when you have a beat at low velocity, Hydrogen will apply a lower gain to the sample. This is how Hydrogen used to work always, and is by default selected in Hydrogen

En francais :) imaginons un set de samples de caisse claire tous normalisés. Si on select "apply velocity", Hydrogen va changer le gain des samples par rapport a la force du coup. Les coups moins fort jouerons diminuront le volume des samples, et vice-versa.

2. Now you have a second set of samples where the velocity is "hardcoded", quieter samples have actually a lower gain, when louder samples are....well....louder :) If you untick "apply velocity", Hydrogen won't change the gain of the sample based on the velocity of the beat. It will only select the right layer in the layer list.
In the oldver version of Hydrogen where "apply velocity" was the default, I would have had to normalise my samples first, otherwise there was a double gain "applied": the one "hardcoded" in the sample and the one applied by Hydrogen.

Cette fois ci, si on a un set de samples de caisse claire ou les coups moins forts ont un volume bas et les coups plus fort on un volume haut. Dans ce cas, deselectionner "apply velocity" n'applique aucun volume additionel, sinon les samples qui ont deja un volume bas Hydrogen va diminuer le gain pour les coups les plus bas.


I hope it's clearer!

Let me know otherwise :)

PR #371, opened 2016-04-20
Fixed midi mapping

kurddt:
Following my message to the developer mailing list,

I have created a branch which use the midi output note instrument's member to find a instrument to play when an midi input note event occurs.

As Sebastian suggested me, it is an optional mode selectable using the preference dialog

Guillaume

### DIFF
--- a/src/core/include/hydrogen/Preferences.h
+++ b/src/core/include/hydrogen/Preferences.h
@@ -207,6 +207,7 @@ public:
 	QString				m_sMidiPortName;
 	int					m_nMidiChannelFilter;
 	bool				m_bMidiNoteOffIgnore;
+	bool				m_bMidiFixedMapping;
 	bool				m_bMidiDiscardNoteAfterAction;
 
 	//	alsa audio driver properties ___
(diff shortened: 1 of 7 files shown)


mauser:
Thanks for the pull request! It has been merged now..

trebmuh:
Hi @kurddt , I'm reading at this bug and, as far as I understand it, it's a new feature for Hydrogen. If I'm not wrong, it does look to be checkable in the _Hydrogen menu -> Tools -> Preferences -> Midi System -> Use output note as input note_. Is that right?
Further, I'm trying to understand a use-case for this feature:
1) for myself :)
2) in order to put something in the 0.9.7 changelog that you can find here: https://github.com/hydrogen-music/hydrogen/wiki/New-features-in-Hydrogen-0.9.7
Would you mind please to explain/show me a use-case for this feature? Thanks.

kurddt:
Hi Olivier,

You are right the feature can be disabled (and it is by default). I will explain you my personal use case.
I had several already made midi songs, with a drum-kit track. My goal was to be able to play the drum-kit part throw hydrogen, so used hydrogen as an synthesizer.
But I came across the following problem, as is the played instrument does not correspond of what I was expecting (crash plays instead of snare of instance).

As it was no way to choose which input midi note triggered which instrument, I did one. I used the already existing output note to do select the desired input note.

So my work flow is the following.

Rosegarden (imported midi ) -> Hydrogen (midi input) - Hydrogen (audio output) > sound card / DAW

I hope it is clear, if not we can switch in French.

Issue #74, opened 2012-11-29
manual not available in 0.9.5 and 0.9.6


thijz:
```
  on 2011-10-17 16:00 * By AutoStatic
```

This is Ubuntu and maybe Debian specific and can be traced down to the Rules file provided by the Debian Multimedia maintainers git repository of Hydrogen. It's about this section:
 # Put aside upstream-shipped autogenerated files during build
 upstreamtmpfiles = $(wildcard data/doc/_.html data/doc/_.docbook)
 upstreamtmpfiles_ = $(patsubst %.upstream,%,$(wildcard data/doc/_.html.upstream data/doc/_.docbook.upstream))
 pre-build:: debian/stamp-upstreamtmpstuff
 debian/stamp-upstreamtmpstuff: debian/stamp-copyright-check
    for file in $(upstreamtmpfiles); do \
        [ ! -e $$file ] || [ -e $$file.upstream ] || mv $$file $$file.upstream; \
    done
    touch $@
 clean::
    for file in $(upstreamtmpfiles_); do \
        [ ! -e $$file.upstream ] || mv -f $$file.upstream $$file; \
    done
    rm -f debian/stamp-upstreamtmpstuff

 common-build-indep:: debian/docs.stamp
 debian/docs.stamp:
    $(MAKE) -C data/doc
    touch $@
 clean::
    rm -f debian/docs.stamp

Something goes wrong there, apparently the included .docbook and .html files get renamed properly (as they should according to the readme that comes with the documentation) but don't get regenerated.

Best,

Jeremy

---

```
  on 2011-10-18 12:50 * By AutoStatic
```

Found the reason why $(MAKE) -C data/doc doesn't seem to work. The manual.docbook file doesn't match with the different po files:

make -C data/doc
make: Entering directory `/home/jjongep1/PPA/Hydrogen/hydrogen.test/data/doc'
cp -f manual.docbook manual_en.docbook
xmllint --noout --valid manual_en.docbook
touch manual_en.docbook_validated
LL=$(echo -n manual_en.docbook | sed 's/.*_\(..\)\.docbook/\1/') ; \
    xmlto html-nochunks --stringparam section.autolabel=1 --stringparam toc.max.depth=2 --stringparam xref.with.number.and.title=0 \
        --stringparam l10n.gentext.language=$LL \
        manual_en.docbook
po2xml manual.docbook manual_es.po > manual_es.docbook
line 143: unexpected token: #. Tag: title
[...]

PR #394, opened 2016-08-11
Update manual_fr.po

trebmuh:
A few fixes spotted when I was fiddling around https://github.com/hydrogen-music/hydrogen/issues/74 .


trebmuh:
Sweet, thanks for your quickness.

Issue #342, opened 2015-12-20
compile problems on Fedora 23


jin-eld:
Hi,

I just pulled master head (2db8f6b6ea41a0ad4bd4cc6dcc2bdae404571c78)and tried to compile, unfortunately it fails:

In file included from /hydrogen/src/core/src/fx/effects.cpp:36:0:
/usr/include/lrdf.h:8:20: fatal error: raptor.h: No such file or directory
compilation terminated.
src/core/CMakeFiles/hydrogen-core-0.9.7.dir/build.make:638: recipe for target 'src/core/CMakeFiles/hydrogen-core-0.9.7.dir/src/fx/effects.cpp.o' failed
make[2]: *** [src/core/CMakeFiles/hydrogen-core-0.9.7.dir/src/fx/effects.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:181: recipe for target 'src/core/CMakeFiles/hydrogen-core-0.9.7.dir/all' failed
make[1]: *** [src/core/CMakeFiles/hydrogen-core-0.9.7.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2

I found a closed issue #194 which more or less describes the same problem, so theoretically it should be fixed, but somehow it still fails here.

I do have raptor2 installed (raptor2-devel-2.0.15-4.fc23.x86_64):

$ pkg-config raptor2 --cflags
-I/usr/include/raptor2 

If it was autotools I'd probably come up with a pull request, but I'm not friends with cmake, so I hope you can help me :)

Kind regards,
Jin

pvint:
I had this happen on my Gentoo system, I worked around by creating a symlink /usr/include/raptor.h -> /usr/include/raptor2/raptor.h

I've been meaning to look into a proper fix (so this report is a good reminder!), but that symlink should let you compile fine.

Paul

jin-eld:
Mhmm.. usually I'd pass some CFLAGS/CXXFLAGS from command line (which I tried and failed, but not sure if cmake looks at it at all), linking around on the rootfs is quite ugly, but thanks for the hint.

pvint:
Ha - yeah, I agree that making the symlink(s) is pretty ugly.... that actually makes me really glad that you raised this issue, since I had completely forgotten that I did that!

[...]

Issue #194, opened 2014-08-11
add raptor2 compatibility for lrdf


schivmeister:
Hello fellow drummers

Hydrogen will fail to build with lrdf if raptor2 is installed:

``` .sh
/usr/include/lrdf.h:8:20: fatal error: raptor.h: No such file or directory
```

This is obviously because the includes are in a different path for raptor2, and although cmake (by way of pkg-config, whether or not LRDF is ON) populates the relevant env vars, they are not used:

``` .sh
[hydrogen-0.9.6]$ grep -Ri 'raptor2'
CMakeCache.txt:LRDF_CFLAGS:INTERNAL=-I/usr/include/raptor2
CMakeCache.txt:LRDF_INCLUDE_DIRS:INTERNAL=/usr/include/raptor2
CMakeCache.txt:LRDF_STATIC_CFLAGS:INTERNAL=-I/usr/include/raptor2
CMakeCache.txt:LRDF_STATIC_INCLUDE_DIRS:INTERNAL=/usr/include/raptor2
CMakeCache.txt:LRDF_STATIC_LDFLAGS:INTERNAL=-llrdf;-lraptor2;-lcrypto
CMakeCache.txt:LRDF_STATIC_LIBRARIES:INTERNAL=lrdf;raptor2;crypto
```

This is actually an old issue and I have propagated fixes to some other upstream projects since 2011, but I never got around to reporting to hydrogen.

Now that the buildsystem changed to cmake, my old patch no longer works. The fix is buildsystem-dependent, so here I propose a simple cmake-friendly modification:

``` .diff
--- hydrogen-0.9.6.orig/CMakeLists.txt
+++ hydrogen-0.9.6/CMakeLists.txt
@@ -193,6 +193,9 @@
 FIND_HELPER(PULSEAUDIO pulseaudio pulse/pulseaudio.h pulse)
 FIND_HELPER(LASH lash-1.0 lash/lash.h lash)
 FIND_HELPER(LRDF lrdf lrdf.h lrdf)
+IF(LRDF_FOUND)
+    include_directories(${LRDF_INCLUDE_DIRS})
+ENDIF(LRDF_FOUND)
 FIND_HELPER(RUBBERBAND rubberband RubberBandStretcher.h rubberband)
 FIND_HELPER(CPPUNIT cppunit cppunit/TestCase.h cppunit)
 # CHECK FOR Qt4
```

That should be it -- keep up the good work, guys!

PR #395, opened 2016-08-12
Fixes Raptor2 compile issue

pvint:
See issue #342 

I have tested this on several Gentoo installations and on Debian Jessie - all work fine (the gentoo installs failed to compile without this change, Jessie was OK either way).


mauser:
Hi! 

I'm unsure about this fix..  i'm afraid that this change breaks build which are not using pkgconfig but having the raptor issue described in #194 ? It would also be interesting to see why you have to use the PC_ prefix. Maybe you could have a look at the behaviour of FindHelper.cmake on your system? There are some lines which can be commented out to give more details about the populated variables..  

In addition, i do not really understand why we need that include in the main CMakeLists.txt in the first place , since there is an similar statement in src/core/CMakeLists.txt. I believe that the #194 would be better fixed by including ${LRDF_INCLUDE_DIRS} in CMakeLists.txt.  Could you investigate further in this direction? At the moment i don't have access to a gentoo system or such, so it is hard for me to look deeper into the problem.

pvint:
I will look into it more... to be honest I don't remember anymore how I came up with the PC_ prefix some months ago.... all I know is I need it to make it compile. I now have two systems (one Gentoo and one Debian Jessie) and one needs this fix and one doesn't, so I can test fairly well.

I will test out src/core/CMakeLists.txt - I think you're right in that it's a more appropriate place.

I will also look into FindHelper.cmake.

Thanks @mauser 

Paul

pvint:
I'm glad you questioned this, @mauser as it was good to have a deeper look (and I learned a bit about cmake along the way!).

I have removed the whole if(LRDF_FOUND) bit in CMakelists.txt, and in src/core/CMakelists.txt I simply added "${RAPTOR_INCLUDE_DIRS}" to the includes.

This made more sense, as pkg-config returns the correct include dir on the machine with raptor2 in /usr/nclude/raptor2/raptor.h and on the machines with raptor.h in the default includes dir pkg_config returns nothing, which also works.

On my Gentoo system:
pvint@nautilus /usr/local/src/hydrogen $ pkg-config lrdf --cflags
-I/usr/include/raptor2

Let me know what you think

Paul

mauser:
Great! That looks good. 

@schivmeister : Could you please check if hydrogen still compiles for you? This pull request modified a part of code from your isse #194.

trebmuh:
added to https://github.com/hydrogen-music/hydrogen/wiki/New-features-in-Hydrogen-0.9.7#bugfixes

schivmeister:
Hey guys, I don't think I will be able to get to it in a timely manner as I have been away from my Linux systems. However, the proposed changes look like the better solution as mine was just a hack (and does not work for everyone as we can see from the newer bug reports).

Issue #372, opened 2016-04-21
9.7 Beta 2 for Win opens an unnecessary CMD window


Peregrino69:
When I start the current beta in Win 7, it opens a CMD window with following:

```
InitTime : multimedia timer resolution set to 1 milliseconds
InitTime : multimedia timer resolution set to 1 milliseconds
(E) JackOutput::init Unknown status with JACK server.←[0m
(E) JackOutput::init Unknown status with JACK server.←[0m
(E) ::H2Core::AudioOutput* H2Core::createDriver(const QString&) Error starting a
udio driver [audioDriver::init()]←[0m
(E) XMLDoc::read XML document C:/Users/lainear/.hydrogen/data/drumkits/AVL-Black
Pearl-4A-1.0/drumkit.xml is not valid (C:/Program Files (x86)/Hydrogen/data/xsd/
drumkit.xsd), loading may fail←[0m
(E) Legacy::load_drumkit loading drumkit with legacy code←[0m
(E) XMLDoc::read XML document C:/Users/lainear/.hydrogen/data/drumkits/AVL-Black
Pearl-4B-1.0/drumkit.xml is not valid (C:/Program Files (x86)/Hydrogen/data/xsd/
drumkit.xsd), loading may fail←[0m
(E) Legacy::load_drumkit loading drumkit with legacy code←[0m
(E) XMLDoc::read XML document C:/Users/lainear/.hydrogen/data/drumkits/AVL-Black
Pearl-5-1.0/drumkit.xml is not valid (C:/Program Files (x86)/Hydrogen/data/xsd/d
rumkit.xsd), loading may fail←[0m
(E) Legacy::load_drumkit loading drumkit with legacy code←[0m
(E) XMLDoc::read XML document C:/Users/lainear/.hydrogen/data/drumkits/AVL-RedZe
p-4-1.0/drumkit.xml is not valid (C:/Program Files (x86)/Hydrogen/data/xsd/drumk
it.xsd), loading may fail←[0m
(E) Legacy::load_drumkit loading drumkit with legacy code←[0m
(E) XMLDoc::read XML document C:/Users/lainear/.hydrogen/data/drumkits/AVL-RedZe
p-5-1.0/drumkit.xml is not valid (C:/Program Files (x86)/Hydrogen/data/xsd/drumk
it.xsd), loading may fail←[0m
(E) Legacy::load_drumkit loading drumkit with legacy code←[0m
[...]

PR #398, opened 2016-08-16
Windows: create console window only in debug builds

elpescado:
This may fix issue #372. This patch creates console window only in debug builds.


mauser:
Thanks a lot, that did the trick!

PR #404, opened 2016-08-29
Updated Windows cross compile scripts and 64 bit fixes

mikotoiii:
Hey there,
I updated the cross compile scripts today a fair bit, here are some of the changes: 
Fix: Cleaned up the path errors in the script which was preventing it from cleaning up after itself.
Added: ability to let people make cross_compile builds without the need to hardcode file names in the CMakesLists file
Fix: the 64 bit packages should now compile again @mauser, my test ones went fine.
Updated: Readme.md to reflect the changes made to the way you build Windows packages now.


mauser:
Thanks a lot! I've uploaded the new 64bit package already to Sourceforge.

PR #405, opened 2016-08-30
ML links update + typo

trebmuh:
Sourceforge mailing lists links update + one typo.

### DIFF
--- a/src/gui/src/AboutDialog.cpp
+++ b/src/gui/src/AboutDialog.cpp
@@ -50,7 +50,7 @@ AboutDialog::AboutDialog(QWidget* parent)
 	about += "http://sourceforge.net/projects/hydrogen<br><br>";
 	about += tr( "<b>Mailing lists</b><br>");
 	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-users<br>";
-	about += "https://lists.sourceforge.net/lists/listinfo/hydrogen-announce"<br>";
+	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-announce<br>";
 	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-devel<br>";
 	aboutTxt->setText( about );
 
--- a/src/gui/src/AboutDialog.cpp
+++ b/src/gui/src/AboutDialog.cpp
@@ -49,6 +49,8 @@ AboutDialog::AboutDialog(QWidget* parent)
 	about += tr( "<b>Project page</b><br>");
 	about += "http://sourceforge.net/projects/hydrogen<br><br>";
 	about += tr( "<b>Mailing lists</b><br>");
+	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-users<br>";
+	about += "https://lists.sourceforge.net/lists/listinfo/hydrogen-announce"<br>";
 	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-devel<br>";
 	aboutTxt->setText( about );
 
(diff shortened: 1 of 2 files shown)


src/gui/src/AboutDialog.cpp
```diff
@@ -49,6 +49,8 @@ AboutDialog::AboutDialog(QWidget* parent)
 	about += tr( "<b>Project page</b><br>");
 	about += "http://sourceforge.net/projects/hydrogen<br><br>";
 	about += tr( "<b>Mailing lists</b><br>");
+	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-users<br>";
+	about += "http://lists.sourceforge.net/lists/listinfo/hydrogen-announce<br>";
```
elpescado: Those mailing lists are essentially dead.

trebmuh: Maybe a part of the "dead" situation is coming from the fact they are not advertised ? I'm thinking especially of the use ML here.

mauser: The announce mailing list was listed there until last year, i have removed it. Those mailing lists are really dead. Btw, both are still listed on the homepage, so i can't see why that they are not advertised.

trebmuh: Still listed on the homepage **but** in the "developer zone" which is not the place the users are going to in the first place I guess.

That said, my first intention was to correct the fact that the code here above states "Mailing lists" (note the plural form). So I checked the ML available and I found the 2 others. Then I put these in the code in this commit.

It does look that keeping those ML is not the way you guys want to take, and I've got no opinion about that. I would only suggest here that it should be consistent across the software, the documentation and the website.

So, it does look to me that we have 2 possibilities:

1. these 2 ML are to be considered as "dead", then we should:
- remove these from the "developer zone -> ML"
- remove the plural form in this commit

2. these 2 ML are not that dead, then we should:
- move these from the "developer zone -> ML" to somewhere else where the common users will feel that this information is for them

If we agree on 1), I'm then removing these from the src/gui/src/AboutDialog.cpp and will enjoy to remove the plural form in the same time, making it clearer.

Agree ?

mauser:
Hi,

i would vote for 1) ...

trebmuh:
OK, I updated in regards to.

What is the way to follow to ask for the 2 dead lists to be removed from http://hydrogen.popez.org/hcms/node/27 ?

Also, while we are here talking about the AboutDialog.cpp file, would you think that putting a link straight to the user forum would make sense? It would make the user to know directly where the best place to go is for a user. What do you think?

trebmuh:
@thijz , @wolkstein :  what do you think about this? Do you think we can move forward?

thijz:
+1 for option 1)

trebmuh:
Sweet. Thanks thijs.

@mauser : what is the correct way for 'remove these from the "developer zone -> ML"' ? Is @wolkstein still in charge with this? Maybe should I send a message to the -devel list. Should I?

mauser:
Hi,

i've removed the links from the website.

trebmuh:
Sweet. Thanks @mauser . Would you mind merging this PR then?

mauser:
Done! 

About the forum link: We're currently working on a new homepage, so the link to the forum will most certainly change in the next time. I would like to keep it off the current version, but this is of course a good idea for the future. We could also create a "visit forum" or "hydrogen community" entry in the "info" menu.

trebmuh:
thanks for the merge.
+1 for keeping this idea while waiting for the new website to pop up

PR #408, opened 2016-09-03
DMG install script

elpescado:
OK, that was easier than anticipated. Basically, `macdeployqt` and `hdiutil` do the job:

zrzut ekranu 2016-09-03 o 21 44 19

This is a basic script that creates dmg disk image. There's no fancy background like most installers have. I'd like to eventually replace it with CPack, but it's more complicated and poorly documented:/

Last but not least, here's 0.9.7-beta3 for Mac:
https://dl.dropboxusercontent.com/u/1474460/Hydrogen/hydrogen-0.9.7-beta3.dmg


mauser:
Hi! 

Great! That was really fast. I've tested the dmg on my system and everything went well. I (or to be more precious, a friend of mine..) will test it on another system (which has no QT installed) and report back..

mauser:
Hi! 

I've merged the pull request, but couldn't upload the dmg yet to SF due to login problems on the SF server. Hope that those problems will be solved soon..

PR #428, opened 2016-11-15
Initial website redesign

mikotoiii:
Hey @mauser,
Here's the pull request for the initial version of the website. We can start from here, and make any changes that we need. We'll want to flesh out the pages, and content, and I'd like to see the downloads page redesigned.


mauser:
Hi Steve, 

could create a pull request for our organisation's repository instead of this one? I gave this another thought and i think it is better to put all of the homepage content to a different repository to keep the size of the source-code repository small. I guess that it is not unlikely that the main repository will contain high-res images and such stuff at some point.. 

I've created the organization repository here: https://github.com/hydrogen-music/hydrogen-music

mikotoiii:
I've been trying to get a pull request set up, but my git complains about
there being no HEAD in the repository, since it's empty, and I can't push
directly into the organizations repo. If you want to copy the gh-pages
branch files into the organization repo, I can then get the pull request
going.

Thanks,

Steve

mauser:
Hi! 

The repository should now be ready and i've also sent you an invitiation which grants you write access to the website repo!

mikotoiii:
Thanks!
The initial website has been pushed up. We'll need to set up the GHPages in
the settings of the repo to get it to display, and since the path has
potentially changed from hydrogen to hydrogen-music, we might have to
update the website files to reflect that (not sure how organizations work
on GH vs personal). Otherwise it should be good to go.

Steve

mikotoiii:
Uploaded to the new repo for the website. Closing this pull request.

mauser:
Great! Github is now configured to use the master branch for the page and it seems to be already working: https://hydrogen-music.github.io/hydrogen-music/

mikotoiii:
Excellent. Now all we have to do is flesh out the content we want on the
site, and we'll be golden. Any new md files put in the root should
automatically add themselves to the menu, and forums should always be the
last option.

trebmuh:
Excellent!
I suggest to reorganise the header in a more user point of view orientation :
Instead of "Developer Zone | Documentation | Downloads | Features | Screenshots | Forums" -> "Features | Screenshots | Downloads | Documentation | Forums | Developer Zone"
What do you guys think?

mikotoiii:
I like the idea of that, the trouble becomes that we need to sacrifice our
menu being dynamic at that point, since it loops through the pages
alphabetically (with the exception of forums which is hosted externally to
github, thus needs to be hardcoded).

PR #430, opened 2016-11-17
Updated Windows cross compile scripts for QT5 and other goodies

mikotoiii:
Added qt5 to mxe.sh
Added compilation for Windows using QT5
Added a new option (-o) called old build to compile for Windows using QT4 for backwards compatibility.
Added a new option (-r) to build releases for Windows. This will build both the 32bit and 64bit packages
Added a new option (-c) to clean the CMake files out of the windows directory, making it a fresh compile next time you run it.
Fixed up some path issues so now you can run the script from either the windows directory, or from the parent hydrogen directory. This could be used to chain load the cross compile from the main build script.

Note: I ran into some issues when making qt5 on my existing mxe installation. It hung up on the qt5system package (IIRC). I had to manually copy the new qt5system.mk file from the mxe github to the src dir in /opt/mxe and then I was able to proceed with my install (which took a little over an hour to complete). This issue should not be present on a fresh install of mxe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ SET(LIBSNDFILE_VERSION_PREV "1.0.17")
 #
 SET(WANT_LIBTAR TRUE)
 OPTION(WANT_DEBUG           "Build with debug information" ON)
-IF(WANT_OLD_BUILD FALSE)
+IF(NOT WANT_OLD_BUILD)
     OPTION(WANT_QT5             "Build with Qt5" ON)
 ENDIF()
 IF(APPLE)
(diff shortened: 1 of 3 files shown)


mauser:
Merged! Thanks for the update! 

About the Qt4/Qt5 option: I suppose that we should remove Qt4 support soon, maybe after the release of a first alpha version.. So i hope that the Qt4/Qt5 duality is not present for a long time..

mikotoiii:
It will be easy enough to rip it out though like anything we'll need to spend a little time cleaning up afterwards since it checks a fair bit if it's building on qt4 or qt5.

trebmuh:
added here: https://github.com/hydrogen-music/hydrogen/wiki/New-features-in-Hydrogen-1.0

PR #433, opened 2016-11-19
Fix for issue 416 - Wrong BPM when positioned before first marker

pvint:
Changes:
- by default, there will always be a BPM marker on the first bar
- BPM marker on bar 1 and song BPM are linked
- in the Add BPM Marker dialog, BPM value is now selected by default (this was just something that bugged me during testing :)


theGreatWhiteShark:
Superseded by #1418. Thanks for the input!

PR #403, opened 2016-08-26
Triplet position fix

pvint:
Fix for issue 366 where notes get placed in wrong position when at highest zoom level and placing triplets.

Issue #366, opened 2016-03-28
When zoomed out completely, adding a note using triplet resolution doesn't place the note where you click


JohnCPun:
Windows 7 x64
Hydogen 0.9.7-windowsPreview [Aug 31 2015]

In pattern editor zoom out completely (using the - button).
Now change the size of the pattern to the max (32) and the resolution to 16T or 32T. 
Add a note on count 15.

Expected:
It is added on count 15

Observation:
It is added on count 16

The distance between where you click and where the note appears actually is relative to where you click. For example: if you add a note on count 8, it will show up on count 8.5.
It is also related to the resolution you select. At 32T it will behave the same way as at 16T but at 8T it really only starts showing around count 13.
When you zoom in just 1 level it will add the note where you click.

pvint:
I can confirm this on Linux with latest git (Aug 25, 2016)

pvint:
A fix to test out on pull request #403

mauser:
Hi,
this is fixed now in the master branch..

trebmuh:
@pvint and @mauser : sorry guys but fresh rebuild from today's git (https://github.com/hydrogen-music/hydrogen/commit/8869f5af04db6d8ad63aa1d45e45fe573cc27646) and this bug is still here. Let me know if I can be of any help with that.

pvint:
Strange, I find the same thing on my desktop. 

I should be able to have a look on Friday.

Thanks for the report @trebmuh

trebmuh:
No worries. Keep me in the loop.

pvint:
Finally looking back on this again, and I must admit that I'm still a bit puzzled as I was perfectly confident it was tested well! It's still broken, and I will figure it out. 
Just too a quick refresher look at it tonight, and I think I will have to start right from the beginning where the grid is laid out and see how the grid width etc is all calculated.

More to come.

pvint:
Note to self: I originally "fixed" this on my laptop - possibly something different there... QT maybe??

Note to anyone who has access: Can we re-open this issue since it is clearly not solved? Thanks!

mauser:
Done!

trebmuh:
Feel free to ping me if you need someone to test stuff @pvint

pvint:
@trebmuh  After my previous failure, maybe it's best that someone else tests this this time! ;)
If you have a minute, test my pull #434 

Thanks!

pvint:
[...]

PR #434, opened 2016-11-19
Triplet position fix - Issue 366

pvint:
There was in int used which needed to be a float when calculating the position.

See issue #366  (and for reference, the incorrect fix at #403 )

Paul


trebmuh:
Well done :)

PR #444, opened 2016-11-30
Fix usleep argument

0-wiz-0:
According to POSIX (http://pubs.opengroup.org/onlinepubs/009695399/functions/usleep.html):
````The useconds argument shall be less than one million.````
Make it so.

### DIFF
--- a/src/core/src/logger.cpp
+++ b/src/core/src/logger.cpp
@@ -31,7 +31,7 @@
 #define LOGGER_SLEEP Sleep( 100 )
 #else
 #include <unistd.h>
-#define LOGGER_SLEEP usleep( 1000000 )
+#define LOGGER_SLEEP do { usleep( 500000 ); usleep( 500000 ); } while (0)
 #endif
 
 namespace H2Core {

### CODE REVIEW

**Comment by mauser** [comment]
Hi! 
Thanks for the pull request! Just out of interest: How have you found this problem? Does it case an error on certain system?

**Comment by elpescado** [comment]
I am ambivalent about this PR.

On one hand, it's clearly stated in POSIX docs that usleep argument should be less than 1 000 000.

On the other hand, I have never seen system that actually enforced that. Mac OS man page does not mention that limitation, and Linux one says:

> ERRORS
>        (...)
>        EINVAL usec is not smaller than 1000000.  (On systems where that is considered an error.)

On the other other hand, if we are so picky about standards, according to [POSIX 1.2008](http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap01.html) `usleep` function is removed. That is also mentioned in Linux man page:

> CONFORMING TO
>        4.3BSD, POSIX.1-2001.  POSIX.1-2001 declares this function obsolete; use nanosleep(2) instead.  POSIX.1-2008 removes the specification of usleep().
> 
>        On the original BSD implementation, and in glibc before version 2.2.2, the return type of this function is void.  The POSIX version returns int, and this is also the prototype used since glibc 2.2.2.
> 
>        Only the EINVAL error return is documented by SUSv2 and POSIX.1-2001.

Windows does not support it at all.

So, finally, I propose, instead of calling `usleep` twice, use standard C++11 feature:

> std::this_thread::sleep_for(std::chrono::seconds(1));

Or C++14:
```

using namespace std::literals;
std::this_thread::sleep_for(1s);
```

And get rid of that whole `#ifdef`:

```
#ifdef WIN32
#include <windows.h>
#define LOGGER_SLEEP Sleep( 100 )
#else
#include <unistd.h>
#define LOGGER_SLEEP do { usleep( 500000 ); usleep( 500000 ); } while (0)
#endif
```

0-wiz-0:
@mauser: It caused 100% CPU usage for a user, because the sleep was immediately terminated, so hydrogen was busylooping.
That's on NetBSD, where the one million limit is also documented in the man page:
http://netbsd.gw.com/cgi-bin/man-cgi?usleep++NetBSD-current

trebmuh:
Written here: https://github.com/hydrogen-music/hydrogen/wiki/New-features-in-Hydrogen-1.0

Issue #454, opened 2016-12-12
Autosave - optionality and visibility of Autosave files


jduemes:
Hi people,

I'd like to suggest two things:

1. That the decision regarding use or not of autosave files should be on the user side.
2. If autosave are in use, it would be great to be able NOT to see them. It is confusing to me seeing all my files in double. The view on "Sound library" it is even more confusing since one can only see the filename hovering mouse over the song name.

In case some one will consider this ideas, my suggestion would be to have on Preferences/General a check box with the name "Generate Autosave files" and on Preferences/Appearance a check box with the name "Show Autosave files".
OS: Arch Linux
Hydrogen 1.0.0-'1297e09'

I personally don't really see a need of having the Autosave file after the project is saved.

jduemes:
I just removed all Autosave files from songs folder and now I don't see new Autosave files anymore. Am I missing something ?
If new versions of H2 don't bother anymore do generate Autosave files, this issue has no reason to exist and can be closed.

pvint:
The way it works currently, I believe, is that it stores the autosave in the directory that your song is saved in, but if it is a new song and has not been saved yet, it stores the autosave in the current working directory.

I'm all for making it optional and/or configurable for how often it autosaves. Might be worth looking at it to be sure that it removes autosave files when the song is saved by the user to help keep things cleaner. 

As for not seeing them - I hear what you are saying and I agree. Might be worth looking into autosaving the files somewhere hidden but have H2 check for autosaves when opening a song and prompt "There's an autosave dated X, and this file is dated Y. Do you want to recover the auto-saved file?" or something similar. I would want to think about that though, because these things can be irritating to users as well.

Short answer: Making autosave an option is a good idea, IMO.

trebmuh:
+1

pvint:
Partly because this has bugged me a bit too - I got looking at this tonight.

Not merged into master yet, but can be tested at https://github.com/hydrogen-music/hydrogen/tree/autosave_454

Added:
- Option in preferences to set the interval for the autosave (or disable it)
- No longer autosave when there have been now changes

TODO:
- Be nice to make the new autosave value take effect without restart
- Look further into where it saves the autosave and how it is presented to users to load or recover

PR #458, opened 2016-12-14
Autosave 454

pvint:
Updates for autosave functionality - see issue #454

### DIFF
--- a/src/core/include/hydrogen/Preferences.h
+++ b/src/core/include/hydrogen/Preferences.h
@@ -187,6 +187,9 @@ public:
 	int					m_countOffset;
 	int					m_startOffset;
 	//~ beatcounter
+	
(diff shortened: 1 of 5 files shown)


mauser:
Hi Paul! 

Thanks for the pull request! I've just tested it and saw that i breaks the layout of the preferences dialog for me (see screenshot below, the text of the first four options is truncated). In addition, there is a TODO comment left PreferencesDialog.cpp  :-) 

pref

jduemes:
I compiled this PR and I have the same problem you described above @mauser. And it breaks the layout of the preferences dialog in english too.

Running system:
4.11.9-1-ARCH

trebmuh:
@pvint : could you please update this PR against master, then I can test it?

PR #460, opened 2016-12-18
Introduce an event to refresh instrument panel with changes

jdekozak:
Hi

Here is a new event to refresh the instrument editor panel instead of the previously buggy timer. Let me know your remarks, I will try to change code accordingly.

Regards,


mauser:
Thanks a lot Julien, i've merged it now!

PR #477, opened 2017-02-16
pattern state indicator improvement for live performance

thijz:
bigger, brighter icons that can indicate more pattern states : 

- 'green arrow' = pattern is playing
- 'no arrow' = pattern is not playing
- 'yellow arrow' = pattern queued = pattern will be active when the pattern loops
- 'red arrow' = pattern is currently playing, but will be de-activated when the pattern loops

icons will only be shown in pattern mode


thijz:
test pull request. cleaning up

PR #482, opened 2017-02-21
Pattern indicators

thijz:
(created new branch for this feature)

1) bigger, brighter icons that can indicate more pattern states :
- 'green arrow' = pattern is playing
- 'no arrow' = pattern is not playing
- 'yellow arrow' = pattern queued = pattern will be active when the pattern loops
- 'red arrow' = pattern is currently playing, but will be de-activated when the pattern loops
(icons will only be shown in pattern mode)

2) added event in PlayerControl so the indicators get updated when switching between song and pattern mode


theGreatWhiteShark:
Superseded by #1544. Thanks for the idea!

PR #487, opened 2017-03-09
Feature/hidpi

elpescado:
Hi,

I have been playing with high dpi support in Qt5. This branch should fix pixelated text on pattern list and song editor ruler. It also makes Hydrogen use higher resolution icons.

Before:
zrzut ekranu 2017-03-09 o 18 01 00

After:
zrzut ekranu 2017-03-09 o 17 05 20

(to see the difference, open two images in new tabs, and switch between them back and forth)

**I need your help!** I don't have high DPI screen, so I was only able to test this issue by forcing Qt5 UI scaling by setting environment variable `QT_SCALE_FACTOR=2`. I'd be grateful for testing this PR on real Retina-like screens, preferably on different OSes.

### DIFF
--- a/src/gui/src/AboutDialog.cpp
+++ b/src/gui/src/AboutDialog.cpp
@@ -36,7 +36,6 @@ AboutDialog::AboutDialog(QWidget* parent)
 	setupUi( this );
 
 	setWindowTitle( tr( "About" ) );
-	setWindowIcon( QPixmap( Skin::getImagePath() + "/icon16.png" ) );
 
 	setMinimumSize( width(), height() );
 	setMaximumSize( width(), height() );
--- a/src/gui/src/DonationDialog.cpp
+++ b/src/gui/src/DonationDialog.cpp
@@ -33,7 +33,6 @@ DonationDialog::DonationDialog(QWidget* parent)
 	setupUi( this );
 
 	setWindowTitle( tr( "Donations" ) );
-	setWindowIcon( QPixmap( Skin::getImagePath() + "/icon16.png" ) );
 }
 
 DonationDialog::~DonationDialog()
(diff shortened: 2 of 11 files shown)


mauser:
Hi! 

Great! I also can't test it, but i've build windows installers for this branch: [32bit](http://stuff.smoors.de/Hydrogen-1.0.0-alpha-win32.exe) / [64bit](http://stuff.smoors.de/Hydrogen-1.0.0-alpha-win64.exe)

pvint:
Sounds good - I will try to test soon.

elpescado:
Both Windows installers give error about missing libraries:

liblzo-error

By the way, high resolution window icons (which means bigger than 16x16 px) should be available regardless of screen pixel density, so even users without Retina-like screen should benefit from it.

mauser:
@elpescado : Sorry, my faullt.. i've fixed that problem some weeks ago in the master branch, but haven't checked if the fix is also present in your branch. I'm going to rebuild at today or tomorrow..

mauser:
The installers should work now, both installed fine on a Windows 7 machine (after rebuilding..).

elpescado:
Updated builds work fine.

Unfortunately, it seems that window icons got lost, at least on Windows 7:

Before:
icon-before

After:
icon-after

I'll try to look into this issue.

elpescado:
OK, I think I have fixed missing window icons.

elpescado:
Hi @mauser ,

Could you provide updated Windows packages?

mauser:
@elpescado : Done! The installers are now up-to-date.

elpescado:
Window icons on Linux:

alt-tab

:)

mauser:
Hi! 

I'm merging this now, i can't see any disadvantages when using it on an FHD screen. At the same time, i still can't test it on a high resolution screen.

alqoholic:
okay so i read all of this.. no idea how to fix the dpi problem here. new to all of this and have no experience in programming. step by step how do i fix my issue??? all the icons are so tiny that i cannot see or read anything...

elpescado:
Hi, you have to wait for 1.0.0 release.

Issue #242, opened 2015-04-01
MIDI actions : event parameter not saved for PROGRAM_CHANGE event


thijz:
message from Alexandre DENIS (contact@alexandredenis.net) that was sent to us via the dev ml : 

Hi,

I encountered a bug in Hydrogen MIDI settings.

Steps to reproduce:
-- open Preferences, select "MIDI System" tab
-- create the following mapping :
  PROGRAM_CHANGE 1 SELECT_NEXT_PATTERN 1
  PROGRAM_CHANGE 2 SELECT_NEXT_PATTERN 2
-- close the dialog
-- open it again, the settings now read:
  PROGRAM_CHANGE 0 SELECT_NEXT_PATTERN 2

Looks like columns Event, Action, and Action Param are saved, but column
Event Param is lost. A quick look in hydrogen.conf confirms that it's
actually the case: column Event param is not saved.

I am using Hydrogen 0.9.6.1 from kxstudio package.

Thank you,
-Alexandre.

thijz:
confirmed : also present in latest 0.9.7

aldenis:
I did some more tests. Looks like only PROGRAM_CHANGE is affected.
I tried to manually add a <eventParameter> in hydrogen.conf, but it didn't work.

jdekozak:
Hi Alexandre,

Would you please describe the symptom of the bug ? Not saving a value is not bug, it might already be a cause.

I took a look at code and MIDI specs, here is my understanding :
* PROGRAM_CHANGE is unique and the event parameter is useless to carry,
* if you want to compare with CONTROL_CHANGE (aka CC), it is not unique hence the event parameter must be used to identify the source of the change.
* but both events can carry a value !

My guess, to solve the issue that is not described in the first e-mail, you have nothing to do, it works !
;)

Regards,

aldenis:
Hi,
It looks like there is an ambiguity on how the MIDI actions are supposed to work and how event filtering is performed.
To me, it looked like it was possible to catch a programm change with a specific number by giving the prog. number as "param", since the param box is not greyed out when PROGRAM_CHANGE is selected as event type, and MIDI learn feature actually fills this column with the program number.
However, there is actually no "catch all" value for param so as to distinguish between catch all events v.s. catch events with a specific number.

Thus we have actually two bugs:
[...]

PR #490, opened 2017-03-26
Fix for #242 program changes handling, program number as parameter

chabanenk0:
Hello! Could you please review my PR. It fixed the bug https://github.com/hydrogen-music/hydrogen/issues/242 for me, because early fix didn't work for me.

 I need to change my loop when I change program number (sound) on my midi keyboard. I need to take the program number into consideration, for example for sound 1 (grand piano) I want to swith to loop number 3 etc.


jdekozak:
Hi,

Quote from #242 : 

But it will only solve half of your problem, you won't be able to toggle the start/stop with PROGRAM_CHANGE 8, Hydrogen has no built-in "mapping by value" feature. Here I won't say this is a feature for Hydrogen, you can use qmidiroute as you did, or mididings as I do with my Korg ES-1.

Have you tried a midi router as described above ?

Regards,

chabanenk0:
Hello, @jdekozak !
I've tried to check if I can start-stop Hydrogen with program change and it worked ok. Please see the screenshot:

hydrogen_config

Hope I understand your question correctly. I was changing the program on my midi keyboard and Hydrogen was successfully changed patterns, and stopped, started playing.

I didn't checked the mms_play or mms_stop events, I don't use them and I don't have it on my keyboard, it's easier to start playing by pressing space key on Hydrogen. And I think my patch should not affect it. If you consider it can cause any issues, please tell the details.
Could you please explain what do you mean by telling "Hydrogen has no built-in "mapping by value" feature." ? 

P.S. yes, I'm using qmidiroute for other purposes. I need a useful event type to switch my hydrogen loops. Program changes are the most useful for me. Of course I could create one more additional output channel and transform my program changes to some note events and connect it to Hydrogen for loops changing, but it is more useful to send the same midi events to the sampler and to the Hydrogen.

jdekozak:
Hi,

First of all, a short disclaimer, I am not a maintainer of Hydrogen, I am giving my opinion on this topic because I have changed a little bit Hydrogen midi code in the past. And I was involved in #242 Also, I am not saying you are wrong doing this change, as always, if there is a need, it reveals a missing piece of software somewhere...

That being said, let me explain deeper "mapping by value". I will compare the CONTROL_CHANGE (aka CC) and the PROGRAM_CHANGE (aka PC). According to MIDI specs, CC can be many while PC is unique, and both events can carry a value, eg. : 
CC X VALUE Y
PC _ VALUE Y

You have coded a "mapping by value" for PC event in Hydrogen. It is visible in midi_map.h file, you have put side by side
 		MidiAction* __cc_array[ 128 ];
		MidiAction* __pc_array[ 128 ];
But there is a difference between those arrays, CC is indexed by X while PC is indexed by Y.
This is first thing that makes me think Hydrogen is not ready to hold a mapping by value feature, and your need must be outsourced.

Next in your pull request, in midi_input.cpp
	MidiAction *pAction = mM->getPCAction( programNumber );
	pAction->setParameter2( QString::number( programNumber ) );
The program_number is used in 2 different lines of code but with a different semantic, first is an index, second is a value, you have coded a "mapping by value" but still carry the value. Repeating things in code often reveals that something is not clean.
This is second thing that makes me think Hydrogen is not ready to hold a mapping by value feature, and your need must be outsourced.

Hope things are clearer now.

Regards

mauser:
Hey everyone! 

I suppose the most confusion comes from the fact that you're using the parameter of the program change message as the second parameter to the midi action and at the same time in other places as the first parameter (via lastMidiEventParameter). I haven't tried the code yet nor looked into this deeper, but this design looks a litttle bit weird to me..

Issue #455, opened 2016-12-13
Double-click on the sample waveform to open the sample editor


trebmuh:
I'd suggest that a double-click on the sample waveform (see picture below) to open the "edit layer" action. Even if it's not here (yet), I'm 90% of the time doing it (certainly a Pavlov thing) so I'm guessing it makes sense from a user-friendly point of view.

![sampleeditor-h2](https://cloud.githubusercontent.com/assets/8705846/21137174/f84e1cde-c128-11e6-84c2-14c25e9a3399.png)

What do you guys think about this?

alworx:
I second that :)
Maybe you and/or mauser want to have a look and test my first contribution to this wonderful project: https://github.com/alworx/hydrogen/commit/e898cbc81f2486c8add5ed0cc9620a8c4331678b

(Sorry for the commit-mess, had some troubles with github/user management)

xjjx:
+1 from me for the idea

Try that .. IMHO much more readable ;-)
BTW. are you sure there shouldn't be:
if( ! pCompo ) return;

?

```C
void InstrumentEditor::waveDisplayDoubleClicked( QWidget* pRef )
{		
	if ( ! m_pInstrument ) return;

	InstrumentComponent* pCompo = m_pInstrument->get_component(m_nSelectedComponent);
	if( pCompo ) return;
	
	H2Core::InstrumentLayer *pLayer = pCompo->get_layer( m_nSelectedLayer );
	if ( pLayer ) {
		Sample* pSample = pLayer->get_sample();
		if( ! pSample ) return;

		QString name = pSample->get_filepath();
		HydrogenApp::get_instance()->showSampleEditor( name, m_nSelectedComponent, m_nSelectedLayer );
	} else {
		loadLayer();
	}	
}
```

pvint:
I don't have time to test, but I had a quick look and I have one comment, @alworx :

In:
`connect( m_pWaveDisplay, SIGNAL( doubleClicked(QWidget*) ), this, SLOT( waveDisplayDoubleClicked(QWidget*) ) );`

I seem to recall there being an issue with using the doubleClicked signal (not sure though!). Maybe it's not supported in different versions of QT or something?  I think I tried this before but it did not work for me, but maybe I just did it wrong. ;)

Overall, looks about right, but I just wanted to raise that question (and perhaps someone else has some input). I will test when I can.

Thanks for the work!
Paul

alworx:
[...]

PR #491, opened 2017-04-06
Enhancement #455: Open sample editor on double-click on waveform in instrument editor

alworx:
Open the sample editor when a sample waveform is double-clicked in instrument editor.

When the selected layer (and thus the waveform) is empty, double-clicking opens the "Load Layer"-dialog.
No action is triggered when the selected layer exists, but the sample cannot be found - to prevent accidental overwriting of sample editor settings.

Addressing (enhancement) issue hydrogen-music#455

Contributor @xjjx made a code-style proposal in the issue, as I'd like to contribute more in the future I would be happy to hear the maintainer's opinion :)

### DIFF
--- a/src/gui/src/InstrumentEditor/InstrumentEditor.cpp
+++ b/src/gui/src/InstrumentEditor/InstrumentEditor.cpp
@@ -388,6 +388,7 @@ InstrumentEditor::InstrumentEditor( QWidget* pParent )
 	m_pWaveDisplay = new WaveDisplay( m_pLayerProp );
 	m_pWaveDisplay->updateDisplay( NULL );
 	m_pWaveDisplay->move( 5, 241 );
+	connect( m_pWaveDisplay, SIGNAL( doubleClicked(QWidget*) ), this, SLOT( waveDisplayDoubleClicked(QWidget*) ) );
 
 	m_pLoadLayerBtn = new Button(
 						  m_pLayerProp,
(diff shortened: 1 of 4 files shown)


trebmuh:
Hi,
just tried it and it does work.
I'd like to suggest one small enhancement and show something which looks like a bug.

A double-clic open the sampler window which is great. That said, an enhancement could be for this sampler windows to take automatically the focus which isn't the case (please see the screencast below).

Something which looks to be a bug is that if I'm double-clicking once, make some changes in the sampler window, then double-click a second time on it, then I'm asked between 2 choices:
1) discard the changes made
2) cancel the action
If I cancel the action, still the window is reloaded and the changes are lost which isn't what one end user would expect to happen.

Thanks for the work already done on it!

alworx:
Hi trebmuh,
thanks for the input! 
The behaviour you've discovered when double-clicking is analog to using the "Edit layer"-button, at least I've not introduced a new bug :)

Nevertheless, you're right that the behaviour is not what the user expects. Even worse, if you open a sample editor and delete the layer in the main form, hydrogen segfaults when clicking "Apply changes".
You could also load a new drumset with the sample editor in background, also leading to a crash of course.

As fixing this bug could change overall behaviour of the sample editor (eg. a modal dialog?) I'll start a separated issue for that, ok?

trebmuh:
Thanks for the explanation. I'm happy to have rose up some legitimate concerns.

mauser:
Hi  @alworx,

first of all, thanks for your contribution! That's really appreciated! 

About the coding style: I would always prefer 

if(a){ 
 b();
}

to a single line if or the version with a newline (but without braces). http://stackoverflow.com/questions/97506/formatting-of-if-statements holds some of the "standard" problems which can come up if you just want to save up a bit of space :) 

We had a bug in once in Hydrogen which resulted from such a two-line-if (without braces), and after that, i always encourage people to use braces..

alworx:
Updated - I'll keep an eye on that in the future :)

mauser:
Perfect, thanks a lot!!

Issue #492, opened 2017-04-07
Sampleeditor crashes when edited instrument layer is changed or deleted


alworx:
As the instrument panel's sample editor is openend as a non-modal dialog, misbehaviour can occur when the user

- changes the sample and opens the sample editor again before applying changes (results in a "save changes" dialog box)
- deletes the layer that contains the currently edited sample and clicks "Apply changes" (crash / segmentation fault)
- loads a new layer in place of the currently edited one (discards changes)

User @trebmuh reported this behaviour and made screencast that can be found here: https://github.com/hydrogen-music/hydrogen/pull/491#issuecomment-292242192

Several ways to fix this:

1. Open the sample editor as modal dialog to suppress configuration changes in the background
2. Check if the sample is still valid when clicking "Apply changes" or closing the sample editor (and show error message if it's not); further warn the user when he tries to re-open the sample editor containing unsaved changes
3. Monitor the editor sample and warn the user when he's about to change or delete the layer
4. Monitor the editor sample and close the sample editor when the user is about to change or delete the layer

I'd prefer the first or second approach, maybe someone has some other thoughts about this?

mauser:
Hi! 

at the moment i would prefer the first approach (making the dialog modal). It is quite simple to implement and robust. I'm not a fan of modal windows at all, but editing the sample is quite a critical thing... adding checks to other parts of the code gets easily complicated and it might be hard to find all critical places..

mauser:
The pull request has been merged, so this bug should be fixed now...

PR #494, opened 2017-04-13
Sample editor dialog is now modal

alworx:
Adressing issue #492: Sampleeditor crashes when edited instrument layer is changed or deleted


mauser:
Thanks a lot!

PR #495, opened 2017-04-20
Setting default values for controls with ctrl+click

alworx:
Coming from Cubase, I've missed the possibility of resetting faders, rotary controls etc. to a reasonable default value by holding the control key and left-clicking the control.

For now, I've used following values for the control's default values:
- 1.0 for no change on gain and volume
- 0.0 for deactivating sends
- 0.0 for withdrawing optional "add"-style settings like attack, delay, random pitch (instrument editor) or humanizing parameters (mixer/master channel)
- 1.0 for withdrawing optional "reduce"-style settings like sustain, cutoff (instrument editor) or return (mixer/plugin section)
- 0.5 for centering balance and pan (though i'd prefer a range of -1 to 1 here, maybe look at it separately)
- 0 for withdrawing pitch adjustments
- whatever value ladspa plugins propose as default value for their individual settings

Implementing this, I've found the fine pitch rotary in instrument editor / layers defined as float though it can only be set in integer steps. I've changed this rotary type to integer, though it could be set to a non-integer value through a drumkit.xml config file - I'd suggest rounding the value to integer when loading?

Looking forward to comments and reviews :)

### DIFF
--- a/src/core/include/hydrogen/fx/LadspaFX.h
+++ b/src/core/include/hydrogen/fx/LadspaFX.h
@@ -102,6 +102,7 @@ public:
 	QString sName;
 	bool isToggle;
 	bool m_bIsInteger;
+	LADSPA_Data fDefaultValue;
 	LADSPA_Data fControlValue;
 	LADSPA_Data fLowerBound;
 	LADSPA_Data fUpperBound;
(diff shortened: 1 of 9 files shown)


mauser:
Hello @alworx,

thanks for the pull request! I've tested it and encountered no errors, so i've merged it directly. It would be great if you could mention it in our manual (https://github.com/hydrogen-music/documentation) , if you have time to do so... 

About the float-loading: Rounding to integer sounds fine to me..

PR #499, opened 2017-05-13
Setup travis

jdekozak:
Hi

I have tried minimal settings for Hydrogen to build on Travis CI. I did it through a lot of trials and errors, and let them in the history for others to understand how to achieve it.

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: cpp
+script: mkdir build && cd build && cmake .. && make
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ addons:
     packages:
       - cmake
       - cmake-data
+      - qt5-default
 
 language:
   - cpp
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
+dist:
+  - trusty
+sudo:
+  - false
 addons:
   apt:
-    sources:
-      - kalakris-cmake
-      - debian-sid
     packages:
-      - cmake
-      - cmake-data
-      - debian-archive-keyring
       - qt5-default
 
 language:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ addons:
     packages:
       - cmake
       - cmake-data
+      - debian-archive-keyring
       - qt5-default
 
 language:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ addons:
   apt:
     sources:
       - kalakris-cmake
+      - debian-sid
     packages:
       - cmake
       - cmake-data
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist:
   - trusty
 sudo:
-  - false
+  - required
 addons:
   apt:
     packages:
@@ -13,7 +13,9 @@ addons:
       - libjack-jackd2-dev
       - liblo-dev
       - libpulse-dev
-
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install ladspa-sdk -y
 language:
   - cpp
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ addons:
       - libarchive-dev
       - libsndfile1-dev
       - libasound2-dev
+      - libjack-jackd2-dev
 
 language:
   - cpp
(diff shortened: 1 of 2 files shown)


elpescado:
That's super nice! Unfortunately, Hydrogen has very few tests adn, to make matters worse, not all pass. Maybe we should comment out that test, and run test suite after `make`?

jdekozak:
Hi

I have noticed also some errors in unit tests, is it the problem you are talking about ?

........................QEventLoop: Cannot be used without QApplication
QDBusConnection: system D-Bus connection created before QCoreApplication. Application may misbehave.
Error XSDError in file:src/tests/data/drumkit/drumkit.xml, at line 14, column 17: Element {http://www.hydrogen-music.org/drumkit}instrumentList is not defined in this scope.
F.Error XSDError in file:src/tests/data/drumkit/drumkit.xml, at line 14, column 17: Element {http://www.hydrogen-music.org/drumkit}instrumentList is not defined in this scope.
Error XSDError in file:data/xsd/drumkit_pattern.xsd, at line 31, column 75: Type psfloat of element element cannot be resolved.
.tests: xxx/hydrogen/src/core/include/hydrogen/Preferences.h:239: static H2Core::Preferences* H2Core::Preferences::get_instance(): Assertion `__instance' failed.

I would rather decouple this pull request from the fix of the unit tests, and focus on the build part only here. Next developer that would fix the unit test would be able to prove it thanks to the Travis build by adding the run of the test suite.
What do you think ?

Regards,

mauser:
Hi!

Thanks for the pull request! I have no experience with Travis, but i like the idea a lot. Fixing the unit tests in a different pull request is fine for me. I would also like to couple the Travis build with an Coverity scan. 

I suppose that i have to configure our repository for the actual travis build, but this has to wait a few days until i have more time to look into this :-/

jdekozak:
Hi,

Thanks for the merge.

Next round will be a OS X build, and I have a few amount time, so let me check what is possible to do with Coverity Scan.

Regards,

elpescado:
`XmlTest::testPattern` fails when run outside root Hydrogen directory:
```
 $ src/tests/tests 
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
..............F


!!!FAILURES!!!
Test Results:
Run:  14   Failures: 1   Errors: 0


1) test: XmlTest::testPattern (F) line: 108 ../src/tests/xml_test.cpp
assertion failed
- Expression: dk0!=0
```

Warning messages can be ignored, but it would be nice to clean them up.

BTW. I noticed that this scripts sets up some version of Ubuntu. Is it possible to specify more than one distribution? AFAIR, there has recently been bug that happened only on certain distributions.

jdekozak:
Hi,

I have a similar output for tests, but not exactly the same, I will try to look at it later : 

QCoreApplication::applicationDirPath: Please instantiate the QApplication object first
........................F.F.tests: xxx/hydrogen/src/core/include/hydrogen/Preferences.h:239: static H2Core::Preferences* H2Core::Preferences::get_instance(): Assertion `__instance' failed.


if I understood correctly, Travis CI provides 2 Linux flavors 'ubuntu precise' and 'ubuntu trusty'. On top of them, a lot of packages can be installed, I saw many gcc and clang flavors.

It is also always possible to tweak the .travis.yml per branch, let's imagine a bug occurs with the latest jack master branch, nothing prevents to git clone jack and compile it within Travis for reproduction purpose. Of course this option will slow down compile time...

Regards,

elpescado:
Here is example issue on Fedora 23: #342

mauser:
>Next round will be a OS X build, and I have a few amount time, so let me check what is possible to do with Coverity Scan.

@jdekozak : I have created a branch named "coverity-scan" which is used as an input to coverity. The first build happened yesterday, i just have to see have i can make the results public.

jdekozak:
Hi

@elpescado I am not sure to understand the problem, and why it is linked to the Fedora/Gentoo distros, but I don't think Travis CI will be able to cover the case. I see it as a tool to prove that "it works for me", of course once the unit tests will be properly setup, and a better way than a README to explain how to build it for distros packagers. It can help also merging faster pull request, build logs are available online.
Also, I don't think building for every kind of setup is possible for Open Source projects that cannot pay for the service and the associated resources.

@mauser This is also what I have noticed from browsing quickly the site, reports are private. I think you will have to customise the Travis build, I cannot do it myself. Doc is available here https://scan.coverity.com/travis_ci

Regards,

PR #502, opened 2017-05-15
Setup travis osx

jdekozak:
Hi,

Added the OS X build to Travis CI matrix, again a lot of commits to setup the build, I did it through trials and errors, I let them for later insight on how to achieve it.

Regards,


elpescado:
That's super cool. Couple of notes:
- Homebrew installs each package into its own directory and creates symlinks in `/usr/local`, except for some packages, and libarchive is one of them. You can run `brew link libarchive` to create those symlinks anyway. After doing that, `cmake` started to detect `libarchive`
- I don't think PulseAudio is needed on Mac

BTW. Could Travis be used to create dmg packages? That would be über cool. I could provide help if needed.

jdekozak:
Hi,

Thanks for your feedback.

As you have noticed, I had issue with libarchive, I decided to use libtar instead. I will try to follow your advice to set it properly. But it raises another question, why does Hydrogen support libtar if libarchive is always preferred over it ?

About pulseaudio, the CMakeLists.txt sets it to ON whatever the platform, I found the package with brew, I decided to keep it for the platform. IMO, it doesn't hurt. On top of that compared to ALSA, pulseaudio is not Linux specific.

Last but not least, about packaging a .dmg, I have not reached that point yet. I can take a look, and you will be able to provide help as soon as this pull request will be merged in master branch. Travis CI is very easy to activate for our forks.

Regards,

elpescado:
Having looked at source code, it seems that libtar support was added first, then libarchive was added (I assume, for Windows compatibility, as libtar doesn't seem to support Win32), and libtar support was simply kept for whatever reason. I'd vote for removal of libtar support.

mauser:
libtar and zlib are kept around for compatibility reasons. They are present on nearly every platform you can think of.. There were issue in the past with libarchive, and people were glad to have a fallback with the classic libtar/zlib combination. 

As elpescado already found out by looking at the sources, libarchive was mainly added to make the cross-platform builds easier. I would recommend to keep libtar and zlib, as they do not create any problem (not that i know of)...  Of course there is some code that has to be maintained for both libraries.

elpescado:
> libtar and zlib are kept around for compatibility reasons. They are present on nearly every platform you can think of..

Except Windows... and Debian once considered removing it: https://lists.debian.org/debian-devel/2011/03/msg01019.html

libtar is also a lit awkward, as it doesn't support compression, so we have first use zlib to gunzip `*tar.gz` to a temporary file, and then untar it with libtar.

On the other hand, maintenance cost of libtar support doesn't seem to be high, so if there are users preferring libtar, it could be kept.

mauser:
Hi Julien,

thanks for the update! The pull request is merged now..

Issue #496, opened 2017-04-26
Revisiting raptor2: New configuration misses include


schivmeister:
I refer to an older issue https://github.com/hydrogen-music/hydrogen/issues/194

And:

- https://github.com/hydrogen-music/hydrogen/issues/342
- https://github.com/hydrogen-music/hydrogen/pull/395

I apologize for not being able to test that last change to the CMake config, but now that I did (I left hydrogen out-of-date on Arch Linux for a long time), it appears @mauser's suspicions came true..

This is what make tries to do:

```
[  0%] Building CXX object src/core/CMakeFiles/hydrogen-core-0.9.7.dir/src/fx/effects.cpp.o
cd /home/schiv/arch/hydrogen/trunk/src/hydrogen-0.9.7/src/core && /usr/bin/c++   -Dhydrogen_core_0_9_7_EXPORTS -I/home/schiv/arch/hydrogen/trunk/src/hydrogen-0.9.7/src/core/include -I/home/schiv/arch/hydrogen/trunk/src/hydrogen-0.9.7/include -I/usr/include/qt4/QtDesigner -I/usr/include/qt4/QtDeclarative -I/usr/include/qt4/QtScriptTools -I/usr/include/qt4/QtDBus -I/usr/include/qt4/QtXml -I/usr/include/qt4/QtSql -I/usr/include/qt4/QtOpenGL -I/usr/include/qt4/QtMultimedia -I/usr/include/qt4/QtNetwork -I/usr/include/qt4/phonon -I/usr/include/qt4/QtXmlPatterns -I/usr/include/qt4/QtWebKit -I/usr/include/qt4/QtHelp -I/usr/include/qt4/QtUiTools -I/usr/include/qt4/QtTest -I/usr/include/qt4/QtScript -I/usr/include/qt4/QtSvg -I/usr/include/qt4/Qt3Support -I/usr/include/qt4/QtGui -I/usr/include/qt4/QtCore -I/usr/share/qt4/mkspecs/default -I/usr/include/qt4  -O2 -std=gnu++11 -Wno-deprecated-register -fno-implement-inlines -g  -fPIC   -o CMakeFiles/hydrogen-core-0.9.7.dir/src/fx/effects.cpp.o -c /home/schiv/arch/hydrogen/trunk/src/hydrogen-0.9.7/src/core/src/fx/effects.cpp
[...]

PR #508, opened 2017-06-11
#496 raptor.h issue. Possible fix.

vladimir-kirillovskiy:
I've removed changes that was made previously for issues #194, #342 and #395.

It seems that the problem was not with the raptor includes, but with LRDF. So I removed lines in CMakeFiles for raptor and corrected a line for LRDF. 

It works for me on Arch Linux.


pvint:
Compiles fine in both Gentoo and Ubuntu! :)

vladimir-kirillovskiy:
I've also noticed spelling error on variable name. I added it on this pull request  as it is in the same file.

@pvint could you re-test it?) Thank you!

pvint:
Still compiles fine with Gentoo - I cannot test at the moment with Ubuntu.

Side note: Good catch on the typo there! Generally it's better to have unrelated changes in a separate PR, but if not one else minds, I'd say to leave this one here.

Since it might affect compiling on a Mac, it would be good if someone with a Mac could verify this works OK.

Thanks!
Paul

mauser:
@pvint : Hydrogen now uses Travis CI, so every new commit to our repo triggers a build on Ubuntu and Mac OS X. The last build on OS X has failed, but it looked more like a timeout problem. I've restarted the build..

pvint:
@mauser  Awesome - thanks for the info. I saw that but didn't have time to look into it. :)

mauser:
@schivmeister : Ray, could you re-test this issue on your system?

mauser:
Hi! 

I've merged this pull request now, it compiled on all system which i've tested so far..

schivmeister:
@mauser Really sorry I failed to respond and act again in time, and thanks for testing it yourselves. I will try to attend to this soon on Arch Linux.

mauser:
@schivmeister No problem, just mention it if there is an issue!

simonvanderveldt:
> Compiles fine in both Gentoo and Ubuntu! :)

@pvint Bit of a random question, but which ebuild are you using for this? Or are you compiling from source?
Asking because the current ebuild in the gentoo tree still has this comment https://github.com/gentoo/gentoo/blob/master/media-sound/hydrogen/hydrogen-0.9.5-r1.ebuild#L39 about issues with finding the raptor2 headers.
Would this PR fix those issues?

Issue #470, opened 2017-01-17
master MUTE does not mute FX returns


thijz:
- select an FX for FX slot 1 (eg a delay)
- create a simple pattern with a snare 
- turn up de FX1 send of the snare and make sure the FX1 return is turned up
- now you should hear the original sound + the delay FX on the snare

- on the mixer hit the master mute button

expected : no sound at all
instead the original snare sound is muted, but the FX sound is still there

tested on latest git code

trebmuh:
confirmed with git https://github.com/hydrogen-music/hydrogen/commit/3002477bfc6cb450b988f49d0a8a643a84e9530d on a Debian Jessie stable up-to-date

trebmuh:
_(Hydrogen built from git master https://github.com/hydrogen-music/hydrogen/commit/bfc355ee5c3fa1a62f12d056427bc1a827764a61)_

Hi @thijz and @jeremyz : reopening this one since looking and retesting it, pressing MUTE on the mixer master track removes the delay effect but I can still hear the original snare sound.

trebmuh:
Freshly rebuild from https://github.com/hydrogen-music/hydrogen/commit/5556c8dbc2a2c58f1441281739eab444273861d0 and it does work as expected
Merci @jeremyz !

jeremyz:
tout le plaisir est pour moi @trebmuh !!

PR #524, opened 2017-08-30
closes issue #470

pmusgrave:
sampler.cpp now checks if instrument is muted before reading effects
level


jeremyz:
thanks a lot for your help,
it is much appreciated.
I've closed this issue with a slightly different solution 18001c4

PR #528, opened 2017-09-16
fix empty detected include dirs

ycollet:
with pkg-config if a include directory is a system include directory, then pkg-config returns nothing.
This make FindPackageHandleStandardArgs fails.
So, I propose to detect empty include dirs and put "/usr/include" is the cmake variable.


jeremyz:
hello,
I would like to know more about your setup,
all my libs stand in /usr/lib/ and /usr/include
and I see no issues ?!?

I've added the above just after  include(FindPackageHandleStandardArgs)
MESSAGE(STATUS " hell ${${prefix}_INCLUDE_DIRS}")

./build.sh  r mm | grep hell
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include/lash-1.0
--  hell /usr/include
--  hell /usr/include
--  hell /usr/include

ycollet:
I use Fedora 25.
cmake version is 3.9.0.
pkg-config version is 0.29.

When I start cmake on hydrogen project, I've got this error message:
```
-- Checking for module 'libarchive'
--   Found libarchive, version 3.2.2
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find LIBARCHIVE (missing: LIBARCHIVE_INCLUDE_DIRS)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindHelper.cmake:43 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:195 (FIND_HELPER)


-- Configuring incomplete, errors occurred!
See also "/home/collette/repositories/hydrogen/build/CMakeFiles/CMakeOutput.log".
See also "/home/collette/repositories/hydrogen/build/CMakeFiles/CMakeError.log".
Makefile:318 : la recette pour la cible « cmake_check_build_system » a échouée
make: *** [cmake_check_build_system] Erreur 1                                                                                                                           
```

jeremyz:
Ok, thanks, on my archlinux :
  $pkg-config libarchive --libs -> '-larchive'
  $pkg-config libarchive --cflags -> ''
so it's definitly the find_path() macro that does the job.
I have cmake 3.9.2, maybe that's where this issue comes from.
what if you add "/usr/include" to the HINTS param of find_path() macro ?
I would prefer to do it that way.
but if it doesn't fix anything, I'll merge your patch and hope we are done with does silly cmake issues ;))

ycollet:
I added /usr/include in the HINT option of find_path:
```
        # find_path
        find_path(${prefix}_INCLUDE_DIRS
            NAMES ${header}
            HINTS ${PC_${prefix}_INCLUDEDIR} ${PC_${prefix}_INCLUDE_DIRS} /usr/include
            ENV ${prefix}_INCLUDE
            )
```

The error message is still the same. I will install cmake-3.9.2 to check if this version fixes the problem.

ycollet:
Same problem with cmake-3.9.2 ...

jeremyz:
ouch !
I merge your patch, I'm sick of these cmake things ;)
thanks a lot for your time.

PR #540, opened 2017-11-21
set package _FOUND flag when using cached variables

m-grabner:
cmake stores the paths of found libraries and header files in its cache. However, when running cmake a second time without clearing the cache before (a common scenario when working with Qt Creator), libraries which were found in the first pass are now reported as "not found". Here is a proposed fix.


jeremyz:
maybe we're now done with cmake issues ... time to move to meson ;)

Issue #547, opened 2018-02-08
Very limited number of layers


dreng:
Hi,
a couple of weeks ago I decided to go with Hydrogen because all of the big players lack Linux (namely LV2) support.

I was very exited when I read about Hydrogen's new round robin feature and I got even more exited when I found a huge drum kit with 127 velocity layers and up to 4 round robin layers for each Instrument here: https://smmdrums.wordpress.com. Unfortunately my intention to create a new very cool drum set for Hydrogen has been stopped when I noticed that only 16 layers can be used with Hydrogen. After compiling a version with a bigger MAX_LAYERS value I've been a little bit bugged about the wasted time and the need of installing a lot of development stuff I'd never need again. At least we can't expect anyone without any programmers experience to compile a custom version just for the sake of using a decent drum kit, can we?

Long story short: I decided to make some code changes which make it possible to set the max number of layers in the preferences dialog. It's not perfect so I'm not sure if it's worth a pull request. At lease it's a better solution than we have in the current release IMO.
1. One has to restart Hydrogen after changing the value in the preferences.
2. I left MAX_LAYERS in core/include/hydrogen/basics/instrument_component.h. Otherwise I had to define a vector and wasn't sure if dynamic arrays are a good idea in this case.
3. If the number of layers isn't a fixed value we need to take care while loading drum kits into Hydrogen. Otherwise it's possible that an instrument with e.g. 60 layers is being cut off at layer no. 16.

I'd be glad if someone could have a look into my fork and tell if it's worth to make a pull request: https://github.com/dreng/hydrogen/tree/max_layers_pref
Suggenstions are also welcome.

If the code will be commited I'll add a german translation.

mauser:
Hi dreng! 

Thanks for your work! This topic has been brought up before, but up to now nobody really looked into the exact consequence of making MAX_LAYERS dynamic. 

About your three concerns: 

The necessity to restart Hydrogen is not a problem, since other settings are also requiring a restart. 

Using a dynamic array for the instrument components could be harmless if all additions / resize operations are not happening in the real-time code (driver callbacks). This has to be checked, but i would suppose that this can be done. 

About the third point: Maybe we should add a messagebox when such a kit is loaded, so that the user is warned that he might loose some layers if he saves that kit again.

mauser:
And of course: A pull request is welcome :)

dreng:
Hi Sebastian,
[...]

PR #550, opened 2018-02-17
Make number of layers configurable #547

dreng:
As stated here https://github.com/hydrogen-music/hydrogen/issues/547 I added the ability to set the max number of layers in the preferences dialog. Because the LayerPreview window can be a bit confusing which becomes worse with a lot of layers I added some text (enumeration and filename) for each layer.

hydrogenprefs
hydrogenlayerpreview


mauser:
@dreng : Thanks a lot! But the travis build of your pull request seems to fail currently due to a syntax error

dreng:
That's strange. It compiled fine on my machine even after running "build.sh c && build.sh r" (otherwise I wouldn't have been able to provide screenshots). Unfortunately I don't know nothing about Travis CI, but I'll get into it.

I see what the problem with the code might be but I should be able to reproduce the error at home though. Can you point me to the right direction please? Are you aware of any differences between Travis and local build options? Compiler flags? Something else?

mauser:
Hi!
you can see the Travis Build Status if you click on "Dout etails" in the "All checks have failed"-Box first and then on one of the linux or os x icons at the lower half of the travis status page. Direct link to the result: https://travis-ci.org/hydrogen-music/hydrogen/jobs/342763319

About the problem itself: Your local compiler seems to be not as strict as the compiler which is used for the travis build. For example: Your using the class "Preferences" from the namespace "H2Core" without adressing it correctly with "H2Core::Preferences".

dreng:
I finally made it to a commit that compiled fine. What really puzzels me is that I was able to build every single commit on my System but Travis wasn't. The only difference I found was the g++ version (7.3.0 in my case and 4.8.4 at Travis CI). Since I couldn't find out how to reproduce the build errors I had to let Travis build the changes and git push and fix bugs a lot of times. This is a really annoying workflow. As it seems that none of you has got such kind of problems, any advice would be appreciated.

elpescado:
Maybe we could enable more build configurations on Travis CI

https://docs.travis-ci.com/user/languages/c/#Choosing-compilers-to-test-against

dreng:
@elpescado There seems to be nothing wrong with the Travis config. It found some really silly mistakes I made. Question is why the local build process didn't complain about them. Maybe somebody can checkout a commit that failed with Travis, build it on his local machine and post the results. Just to make sure I'm not the only one.

trebmuh:
Hi @dreng and all.
I'm reading at this thread for the first time (and the related https://github.com/hydrogen-music/hydrogen/issues/547 as well) and am wondering if you would benefit from any help with testing this enhancement or if it is all good and ready to be merged into master soon.

Please tell me what I can do.

dreng:
Hi trebmuh,
This should be ready to merge. As it has been tested only by myself, it won't hurt to test the patch by somebody else, of course.

jeremyz:
Hi @dreng, @trebmuh,

while the need for a parameter to set the max# of layers is fully understood (and I would like to use your huge drumkit when realy ;) there are 2 things that I would like to see made different.

the first one is that if you change the value in preference dialog and don't restart H2, there is very high chances to segfault. Someday somewhere a call to Preferences::get_instance().getMaxLayers() will lead to something bad (out of bounds access to an array of layers).

the second one is that I strongly feel that these calls to Preferences *pref = Preferences::get_instance(); have no place within basics/*.cpp

one solution to fix all that :
as you set maxLayer as a static var in InstrumentComponent (perfect),
set it at H2 startup, (you can then change it in Preferences when running, it won't affect this one, safe).
then within instrument, drumkit, song, etc. use a static method InstumentComponent::getMaxLayers().

I'm available to help on this if needed

best regards

dreng:
Hi @jeremyz ,

I agree that a global var (respectively a static method) would be a better solution. To be honest I found it quite difficult to implement it this way, not least because of a not very well structured code base.

I'm very busy these days and can't work on it. You seem to be able to do something off the cuff. Feel free to fork my repository and improve whatever you like.

jeremyz:
@dreng ok, PR #586 is up
thanks a lot for your job !!

PR #556, opened 2018-03-08
Update Demo Songs

mikotoiii:
Updated the demo songs to reflect the change to GMRockKit from GMKit, and to fix the default midi setup advice message coming up everytime a demo song was loaded.


mauser:
Hi! 

Thanks a lot, merged!

PR #557, opened 2018-03-09
Update macOS dmg disk image script

elpescado:
Hi,

I have finally found some time to work on Hydrogen. I wanted to update script used to build .dmg disk image for macOS. The original one was more proof-of-concept hack than robust solution so I'm happy to having finally fix it.

Highlights
----------

* Error handling - it should gracefully handle errors encountered during execution
* Flexibility - it's possible to provide path to source binary as well as output .dmg image
* Help (`-h`) ;)

Travis-CI
------

The best part is that I have integrated this script with Travis-CI, so that binary releases for macOS can be built by Travis. I have updated Travis configuration to upload build to a GitHub release. You can see example builds in my fork:
https://github.com/elpescado/hydrogen/releases/tag/untagged-52840cdaf62886f1b8f4

Integrating this with main repository will probably need fine-tuning `deploy:` section in `.travis.yml` file, most important ajdusting `api-key:` and possibly restricting releases to tagged commits (by setting `on.tags` to `true`) and only on `master` branch (by removing `on.all_branches`). Deployment documentation is fairly comprehensive:
https://docs.travis-ci.com/user/deployment/releases/
https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A

Travis provides many other deployment options, for example, it can deploy to GitHub Pages.

Issues
-----

Currently, built binary runs only on macOS 10.12 and higher. To control minimum macOS version `MACOSX_DEPLOYMENT_TARGET` can be set, but setting this variable breaks Travis builds (example: https://travis-ci.org/elpescado/hydrogen/jobs/351026402), despite the fact it builds fine on my machine.

Minimum macOS version required by binary to run can be [examined](https://stackoverflow.com/questions/17143373/determine-minimum-osx-version-a-binary-was-compiled-for) by running `otool -l /path/to/binary` and examining `LC_VERSION_MIN_MACOSX` field. If anyone has exprrtise on macOS and Travis, I welcome every help.


mauser:
Hey ho! 

Thanks for the update! The deployment possibility sounds pretty interesting, this is sth. i haven't done before with Travis. Creating DMG images by Travis for tagged releases would be awesome! 

Is anyone here who could test the new build scripts (not the travis part)? I do not own a mac with a recent OS version any more..

elpescado:
What macOS version do you have?

Unfortunately, I no longer have any of my virtual machines with older macOS releases, so I'm unable to test Hydrogen builds on anything but macOS Sierra.

elpescado:
By the way, I'd be grateful if anyone could review my shell script. I'm definitely not an expert when it comes to shell scripting.

elpescado:
I have cleaned this branch a bit. I hope this branch gets merged before beta2 in order to Mac packages.

I have tried to fix compatibility issue, but with no success. I'll try to address it later. As of now, built binaries are compatible with macOS 10.12 +.

mauser:
@elpescado : I'm stuck with 10.11, as this is the latest version which is compatible with my Macbook. But i wrote a mail to our developer mailing list to ask for help with testing. But i would also apply this pull request if nobody answers in the next time.. in my opinion a "not perfect" Mac OS build is better then no Mac OS build at all :) At least for the moment..

elpescado:
Success! 

I managed to build Hydrogen with compatibility set to macOS 10.10 or greater.

See: https://github.com/elpescado/hydrogen/releases/tag/macos-test-build-1

elpescado:
The solution is simple, but it was far from obvious.

Mac binaries can have minimum OS version required to run. CMake can be instructed to set that value by setting `MACOSX_DEPLOYMENT_TARGET` variable.

Our travis is configured to use Homebrew to install dependencies. Homebrew installs packages in `/usr/local`. Our CMake config is set to expect dependencies in `/opt/local`, I think because MacPorts installs packages there.

Strangely, when building Hydrogen on Travis without setting `MACOSX_DEPLOYMENT_TARGET`, it finds libraries installed in `/usr/local`, but when `MACOSX_DEPLOYMENT_TARGET` is set, header files cannot be found. So I symlinked `/usr/local` to `/opt/local` and it works!

Problems
----

I have noticed that CMake by default configures build in Debug configuration, that is with debug information and without optimizations. I think it would be beneficial to make packages from Release configuration.

Currently we have two jobs per build (Linux and macOS). I'm thinking whether we should change build type (on both OSes or just on Mac?), or have each OS build in both configurations (that would make 4 jobs). Or, select build configuration based on tags, so commits tagged as "release" (eg. 1.0.0 would be built in Release mode whereas rest of commits would be built in Debug mode). I think that would add unneeded complexity.

I think that changing build type to Release makes sense on CI server, because we don't really need debugging symbols there, because even if there are problems with build, debugging is not possible on Travis.

What do you think?

mauser:
@elpescado : Just a quick feedback: It works like a charm with Mac OS 10.11 !! Really awesome! Will care about the travis integration in the next days..

mauser:
The commit is now merged, along with the modifications which have been proposed. Let's see how it works :) Thanks a lot!

Issue #567, opened 2018-03-16
Installation requirements are out of date


theGreatWhiteShark:
The requirements part of the INSTALL.txt file is deprecated. Especially the listing of Qt4 gave me a headache at first.

mauser:
Hi! 

Thanks for the report and the pull request (those are the best issues :smile: )

PR #568, opened 2018-03-16
updating installation requirements

theGreatWhiteShark:
The requirements for compiling Hydrogen were out dated. Especially the listing of previous Qt4 gave me a hard time in the beginning.

In addition I added a compilation of all required packages for Debian-based systems and OS X, which I took from the .travis.yml file. This way, the user can simply copy and paste the dependencies to install them all at once.

I actually have no clue about Mac OS. I just used the commands from the Travis config. Hope they are sufficient.

Linked to issue #567

### DIFF
--- a/linux/debian/control
+++ b/linux/debian/control
@@ -2,7 +2,7 @@ Source: hydrogen
 Section: Multimedia
 Priority: optional
 Maintainer: Sebastian Moors <smoors@users.sf.net>
-Build-Depends: debhelper (>= 4.0.0), libsndfile1-dev, libjack-dev, libasound-dev, libtar-dev, libqt4-dev, liblrdf-dev, cmake, librubberband-dev
+Build-Depends: debhelper (>= 4.0.0), libsndfile1-dev, libtar-dev, liblrdf-dev, cmake, librubberband-dev, qt5-default, libqt5xmlpatterns5-dev, libarchive-dev, libasound2-dev, libjack-jackd2-dev | libjack-dev, liblo-dev, libpulse-dev, libcppunit-dev, liblash-compat-dev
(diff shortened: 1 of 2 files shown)


linux/debian/control
```diff
@@ -2,7 +2,7 @@ Source: hydrogen
 Section: Multimedia
 Priority: optional
 Maintainer: Sebastian Moors <smoors@users.sf.net>
-Build-Depends: debhelper (>= 4.0.0), libsndfile1-dev, libjack-dev, libasound-dev, libtar-dev, libqt4-dev, liblrdf-dev, cmake, librubberband-dev
+Build-Depends: debhelper (>= 4.0.0), libsndfile1-dev, libtar-dev, liblrdf-dev, cmake, librubberband-dev, qt5-default, libqt5xmlpatterns5-dev, libarchive-dev, libasound2-dev, libjack-jackd2-dev, liblo-dev, libpulse-dev, libcppunit-dev, liblash-compat-dev
```
trebmuh: Enforcing libjack-jackd2-dev i a bad idea. Building Hydrogen shouldn't depend on a specific version 1 or 2 of jack, then, we can put:

libjack-jackd2-dev | libjack-dev,

which will allow to build with both, preferably libjack2-dev.

PS: if I remember correctly, there is a thing in Debian which is allowing to build with libjack-jackd2-dev even if libjack-dev is asked.

theGreatWhiteShark:
Yes, you are right. Haven't thought about this one. 

I introduced libjack-dev as an alternative in the control file of the Debian package and wrote a hint in the package section of the INSTALL file.

mauser:
Hey ho,

thanks a lot, those changes are merged now to the master branch.

PR #594, opened 2018-04-28
update installation requirements

theGreatWhiteShark:
Since commit 033379f2814a0d96a6c9724701058a70aac75951 the Qt5LinguistTools package is introduced. To use it, two additional libraries have to be installed

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -116,7 +116,7 @@ installed on your system, and the development header files:
 	qt5-default libqt5xmlpatterns5-dev libarchive-dev
 	libsndfile1-dev libasound2-dev liblo-dev libpulse-dev
 	libcppunit-dev liblrdf-dev liblash-compat-dev
-	librubberband-dev
+	librubberband-dev qttools5-dev qttools5-dev-tools
 
 	In addition, either the libjack-jackd2-dev or libjack-dev
 	package must be present. Which one to pick depends on whether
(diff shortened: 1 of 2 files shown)


trebmuh:
@theGreatWhiteShark there is no needs (tested on Debian Stretch) to ask specifically for `qttools5-dev` to be installed since the installation of `qttools5-dev-tools` will pull it automatically.

theGreatWhiteShark:
In Devuan Jessie I explicitly need to install both of them. Just asking for `qttools5-dev-tools` will not automatically install `qttools5-dev`. In addition, both packages need to be present in order compile Hydrogen.

trebmuh:
I can understand your point. That said, the debian/ folder and the documentation are about "debian". Hence, the fact it doesn't fit with how devuan work is a devuan issue which either needs another paragraph and folder for devuan, or report/act within devuan it is relevant.

theGreatWhiteShark:
Yeah, you are definitely right about the Devuan part. But it is probably more an issue of the difference between Jessie and Stretch.

According to the Debian package [page](https://packages.debian.org/stretch/qttools5-dev) the `qttools5-dev` package depends on the `qttools5-dev-tools` package in **Stretch** but not the other way around. In **Jessie** they seem to be both independent. 

Well, I like my operation systems well tested and compile the newest stuff I need by hand. And I know a lot of people handling it the same way. In the end Jessie is not that old and we ain't talking about maintaining compatibility with Squeeze. 

But as you wish. I can of course remove the `qttools5-dev-tools` entry from the debian file.

theGreatWhiteShark:
Dependencies were introduced with commit 5fa3fdd5b1d1d24808475b909ef337770e920351

PR #618, opened 2018-06-03
remove qt5_use_modules to fix build with Qt 5.11

schnitzeltony:
* it is not necessary - all libs are linked by TARGET_LINK_LIBRARIES
* was removed in Qt 5.11

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>


schnitzeltony:
I think travis failure was caused by some download problem. So: Ping?

Issue #630, opened 2018-07-20
Generation of the documentation fails


theGreatWhiteShark:
Hydrogen version *          :   a89f053d66ed009b3b9b6b8e8f4c1b707c2f0b17
Operating system + version  :  debian-based
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
The build of the documentation fails with the following error

`./build.sh d`

> Generating API documentation with Doxygen
> /bin/sh: 1: cd: can't cd to /home/XXX/hydrogen/docs
> CMakeFiles/doc.dir/build.make:57: recipe for target 'CMakeFiles/doc' failed
> make[3]: *** [CMakeFiles/doc] Error 2
> CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/doc.dir/all' failed
> make[2]: *** [CMakeFiles/doc.dir/all] Error 2
> CMakeFiles/Makefile2:74: recipe for target 'CMakeFiles/doc.dir/rule' failed
> make[1]: *** [CMakeFiles/doc.dir/rule] Error 2
> Makefile:184: recipe for target 'doc' failed
> make: *** [doc] Error 2

jeremyz:
closed by #631

PR #631, opened 2018-07-20
creating ./docs folder while compiling docs

theGreatWhiteShark:
As described in issue #630 the generation of the documentation using Doxygen via the build script `./build.sh d` was not working. The problem was located in *build/CMakeFiles/doc.dir/build.make* line 58 where the `CMakeFiles/doc` target tries to enter the *./docs* folder at the root directory of the repository. Since this folder is not created by either the build script or CMake the generation of the documentation fails.

To solve this issue I introduced a line in the *./CMakeList.txt* file creating the missing *./docs/* folder.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,11 +233,14 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 find_package(Doxygen)
 if(DOXYGEN_FOUND)
 	configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
+	file(MAKE_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/docs)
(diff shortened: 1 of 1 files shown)


jeremyz:
Hi @theGreatWhiteShark  !
I would create the directory in `${CMAKE_CURRENT_BINARY_DIR}/doc`.
and to do so I would use : 
`add_custom_target(doc_dir ALL COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/doc)`
then add `DEPENDS doc_dir` to the doc target.
so that the directory is always created at build time and not only at configure time.

jeremyz:
cmake is a dirty bitch a would say (`make_directory`  beeing deprecated for ages with no known alternative(that I know about)),
but as the guy who brought it to hydrogen, I should just shut my mouth ;)
I'll merge your patch and remove the then useless `file(MAKE_DIRECTORY ...`
and will alos change `CMAKE_CURRENT_SOURCE_DIR`to `CMAKE_CURRENT_BINARY_DIR`.

btw, I can't wait to find the time needed to review #627,
greetings to you @theGreatWhiteShark !

PR #632, opened 2018-07-21
Functional tests

elpescado:
Hi,

I have created "functional tests" that test audio & MIDI export in Hydrogen. I have created some test songs, and exported them to reference audio and MIDI files. Then I have written tests that load those test songs, perform export and compare exported files to reference files. If there is any difference, test fails:

zrzut ekranu 2018-07-21 o 22 26 17

# Details

* Audio files are compared using libsndfile, so result and reference files can be in different formats, and differences in metadata should not be significant. MIDI files are compared byte-by-byte.

* Audio export in Hydrogen is implemented by playing song file through special audio driver that writes audio file to disk instead of playing it on sound card, so this test should cover most of code used for regular playback. OTOH MIDI playback and MIDI export use 2 different code paths, so MIDI tests cover only MIDI export - MIDI playback is not tested here.

* I have added running test suite to travis script, so each build triggers running tests on Travis-CI.

# Problems

* I have found some problems with test song which uses mute groups. I have different results on my computer vs Travis build. I will investigate it later.

# Extras

* I have fixed audio export in Hydrogen CLI


jeremyz:
That is just awesome, 
you just realized what I had in mind when I started the src/test tree.
greetings to you !!
I will definitely review this as soon as possible,
and hopefully fix some audio rendering bugs (no there is none ;) !!

elpescado:
# Problems

I faced one problem while implementing this task. Exported data is saved in temporary files. After passed tests those files are deleted in order to keep things clean and tidy. If test fails, however, resulting files are kept, so after all tests finish, developer can compare them with reference files. The problem is on Travis CI. As builds in Travis are built on fresh VM/containers that are wiped after job finishes - so there's no way of getting exported audio/midi files of failed tests.

Fortunately, Travis provides [artifacts upload tool](https://docs.travis-ci.com/user/uploading-artifacts/), which I [configured](https://github.com/hydrogen-music/hydrogen/pull/632/commits/b5af061e6e9a07eeb8c00b6b2a6401d6ec2a2502).

Unfortunately, that tool only supports Amazon S3, which , again unfortunately, doesn't provide free service (but Amazon provides one-year free trial).

Alternatively, it's possible to run any script in `after_script:` section so that we can upload build artifacts using `scp`/`ftp` etc somewhere.

jeremyz:
honestly, I don't think that it is a problem,
these are the kind of test routines that should be run by every developer before pushing a patch to master or merging something.
a maybe easy thing that could be done is to use travis hook to send an email to the last commiter who broke the tests :  you did something bad, check with tests, fix it and next time check with tests before pushing. 
What is your feeling about that ?

elpescado:
> honestly, I don't think that it is a problem,
> these are the kind of test routines that should be run by every developer before pushing a patch to master or merging something.

That's true, but is sometimes happens (it happened to me, actually*) that tests pass on local machine but fail on Travis. Debugging without generated files is hard in that case.

> a maybe easy thing that could be done is to use travis hook to send an email to the last commiter who broke the tests : you did something bad, check with tests, fix it and next time check with tests before pushing.

I think Travis already does that out of the box.

---

 `*` - see [this test](https://github.com/hydrogen-music/hydrogen/pull/632/commits/ac1046c307cba5bd56a21cd56859b54667a67131) - [travis build](https://travis-ci.org/elpescado/hydrogen/jobs/358026909)

jeremyz:
I was about to merge this right now,
but I don't understand the purpose of the changes you made in Filesystem::tmp_file and Filesystem::tmp_dir. 
My last audit of the file paths used in H2 told me that it was all ok.
Changing these doesn't sound like safe to me.
Can you please elaborate.

elpescado:
[`Filesystem::tmp_file_path`](https://github.com/hydrogen-music/hydrogen/pull/632/files#diff-c3554f494bde100029ac5b2ed8c95519R493) - I changed that to preserve file extension. Before that, if you requested `test.wav`, you'd get something like `/tmp/test.wav.ulAdop` - I changed it to `/tmp/testulAdop.wav`. Those `XXXXX` are replaced by [`QTemporaryFile`](http://doc.qt.io/qt-5/qtemporaryfile.html) with unique random characters. This change is cosmetic and can be reverted if you consider it too risky.

I think I didn't change `Filesystem::tmp_dir`. If I did, consider that a bug:)

I started this branch a while ago, and I rebased it recently against current `master` and had to make some changes in order to keep up with changes in `master`, so maybe something changed during that rebase.

jeremyz:
thanks for the details,
I'll see what I will do about ```Filesystem::tmp_file_path``` because now the doc in ```filesystem.h``` is not in sync with the behaviour.

Issue #558, opened 2018-03-11
desktop file icon missing (in correct location)


dvzrv:
Hydrogen version *          : 0.9.7
Operating system + version  : Arch Linux
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
The [hydrogen.desktop](https://github.com/hydrogen-music/hydrogen/blob/master/linux/hydrogen.desktop#L29) file states `h2-icon` as its icon. However, after `make install`, the only file that is named that way ends up in `/usr/share/hydrogen/data/img/gray/h2-icon.svg` (additionally the same file gets installed as `icon.svg` to the same folder).

This then breaks the displaying of an icon when launching hydrogen through a XDG aware menu or launcher (as those accept locations such as `/usr/share/pixmaps` or `/usr/share/icons/hicolor/scalable/apps` for applications).
It would be nice if this could be addressed in a pull request and won't have to be dealt with downstream.

Apart from the location of the `.svg` file, it is usually good style to name it the same way as the application (i.e. `hydrogen.svg`).
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

mauser:
Hi! 

Could you re-test this with Hydrogen 1.0? This should be fixed with commit b5b9cb902e84917753899eb3d10d70b6df82dc24 .

mauser:
Closing this since it seems to work fine with the current git version of Hydrogen (1.0) . Just mention it if it should be re-opened because sth. is missing.

dvzrv:
The icon should actually be installed as `/usr/share/icons/hicolor/scalable/apps/hydrogen.svg` and the desktop file (i.e. `org.hydrogenmusic.Hydrogen.desktop`) should either not even mention an icon (and default to its own name) or specify hydrogen.
Also, the desktop file name should be simply `hydrogen.desktop`. It's not a dbus configuration ;-)

PR #636, opened 2018-08-06
Allow non-root to make install on linux

bratner:
Having a hard-coded /usr/share/pixmaps prevented me from installing to a nicely separated directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,6 +153,7 @@ ENDIF(APPLE)
 SET(H2_SYS_PATH "${CMAKE_INSTALL_PREFIX}/share/hydrogen")
 # TODO remove data from path, could be .config/hydrogen
 SET(H2_USR_PATH ".hydrogen")
+SET(H2_APP_ICON_PATH "/usr/share/pixmaps" CACHE STRING "Default system location of H2 application icon")
 
 
 SET(MAX_INSTRUMENTS 1000 CACHE STRING "Maximum number of instruments")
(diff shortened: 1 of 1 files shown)


mauser:
Hi! 

the file path is hard-coded because the [free desktop standard](https://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) uses it as one default (see also #558): 

 "Icons and themes are looked for in a set of directories. By default, apps should look in $HOME/.icons (for backwards compatibility), in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps (in that order). Applications may further add their own icon directories to this list, and users may extend or change the list (in application/desktop specific ways).In each of these directories themes are stored as subdirectories. A theme can be spread across several base directories by having subdirectories of the same name. This way users can extend and override system themes. "

From reading the standard, i can't see that changing the default path to "${CMAKE_INSTALL_PREFIX}/share/pixmaps" would be a valid option.But this is just from reading the standard, i don't have an idea what other applications are implementing and if there is "best practice".

bratner:
Hi,

Well I was expecting something of the sort. 
How about putting the path in a cmake variable with a default value?
One of those you can override on cmake invocation? Something like. APP_ICON_INSTALL_PREFIX ?

bratner:
@mauser I have tried to amend my previous commit and got into a bit of git hell. Anyways my last commit vs master is what i meant by my previous comment.
I can go into build settings in QT developer and modify the icon path. Now run/debug can perform "make install" without me being root.

mauser:
Hi! 

About the path: Maybe use sth. like "unix" in it to make it clear that this default path is not used for Mac OS X images or the windows installer?  

Maybe @jeremyz has a comment on this? He is more experienced with the cmake wizardry :)

bratner:
Hi @mauser !

Please take a look at the last commit and let me know if I properly understood your comment or not.

jeremyz:
I think it's all good!

CMakeLists.txt
```diff
@@ -153,8 +153,11 @@ ENDIF(APPLE)
 SET(H2_SYS_PATH "${CMAKE_INSTALL_PREFIX}/share/hydrogen")
 # TODO remove data from path, could be .config/hydrogen
 SET(H2_USR_PATH ".hydrogen")
-SET(H2_APP_ICON_PATH "/usr/share/pixmaps" CACHE STRING "Default system location of H2 application icon")
 
+#Changing this to user writable location will allow successful, non-root, deployment (e.g: make install).
+IF(NOT MINGW AND NOT APPLE)
```
elpescado: I don't think it's strictly necessary, since whole `INSTALL` block is enclosed in `IF(NOT MINGW AND NOT APPLE)` anyway.

bratner: Well it is exactly the reason I put the definition of a variable under the same conditions it will be used. Also thought it derives directly from @mauser 's comment above regarding making it clear that is not used under mac and win.

elpescado: I'm not saying it's wrong. I think either way is OK.

mauser: I'm sorry that i have caused some confusion here. What i meant that you can't tell from the name of that option that this option is only usable in a meaningful way on a linux system. But i guess most people which tinker with those option will have a look at the cmake script nontheless, so everything should be fine!

mauser:
Merged! Again, thanks a lot for the pull request!

Issue #639, opened 2018-08-12
Holes between instrument layers cause sampler to skip notes


theGreatWhiteShark:
After writing the humanizer Hydrogen showed an odd behavior. Occasionally it seemed to just skip notes of specific instruments, e.g. for the Open Hat in the GM kit or for the Kick in Millo_Multilayered2. But all other instruments of the drumkit were not affected. Setting the verbose level to warnings revealed that only notes of very specific velocities in a very narrow range were skipped. 

At first I thought this to be a bug in the pink noise of the tweaked humanizer. But it turned out to occur in both the white noise and the original implementation of the humanizer as well. 

After spending some hours delving into the code base I found the sampler itself to be the cause of the problem. For some instruments the individual layers are not perfectly aligned. Let's take the *Open Hat* of the default GMKit for example. 

> (I) Sampler::__render_note nLayer: 0. Note velocity: 0.796811. Layer velocity_start: 0. Layer velocity_end: 0.202899
> (I) Sampler::__render_note nLayer: 1. Note velocity: 0.796811. Layer velocity_start: 0.202899. Layer velocity_end: 0.394928
> (I) Sampler::__render_note nLayer: 2. Note velocity: 0.796811. Layer velocity_start: 0.398551. Layer velocity_end: 0.601449
> (I) Sampler::__render_note nLayer: 3. Note velocity: 0.796811. Layer velocity_start: 0.605072. Layer velocity_end: 0.793478
> (I) Sampler::__render_note nLayer: 4. Note velocity: 0.796811. Layer velocity_start: 0.797101. Layer velocity_end: 1
> (W) Sampler::__render_note NULL sample for instrument Hat Open. Note velocity: 0.796811.

There are tiny holes between the end of one layer and the start of the next. Whenever the velocity of a note falls into it, it will be skipped by the sampler. The latter just checks whether the note is between the start and the end of a layer.

The problem was just more pronounced since the pink noise is able to create smaller and more subtle fluctuations, even with a large scaling. But you can also reproduce it by setting Open Hat beats to a velocity of 0.8 in the former implementation and setting the humanizer of the velocity to a small, non-zero value.

thijz:
wow ... nice catch !

jeremyz:
fixed by the author of the issue, thanks a lot

PR #640, opened 2018-08-12
sampler: fixing holes in instrument layers

theGreatWhiteShark:
In some drumkits, like the default GM ones, separate layers of an individual instrument are not perfectly aligned. E.g. for the "Hat Open" of the GM Kit the third layer ends at a velocity of 0.793478 and the fourth one starts at 0.797101. But since the selection of a sample for a specific note only checks whether its velocity is *within* each layer, the note can fall between two layers causing the sampler to skip it.

In this patch the sampling will work as it always does. But in the rare cases the note falls between two layers another for loop checks which layer is the closest one to the note and uses it for the sampling. This event should in general occur only occasionally so having a separate chunk so code handling it might be the most efficient way. In addition, the lookup of the nearest layer also respects instruments, which layers are not sorted in ascending order.

This fixes issue #639

### DIFF
--- a/src/core/src/sampler/sampler.cpp
+++ b/src/core/src/sampler/sampler.cpp
@@ -328,6 +328,40 @@ bool Sampler::__render_note( Note* pNote, unsigned nBufferSize, Song* pSong )
 							break;
 						}
 					}
+
+					if ( !pSample ){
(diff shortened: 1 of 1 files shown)


trebmuh:
@mauser : ping

jeremyz:
maybe the common code section could have been factorized in a local static function

Issue #641, opened 2018-08-16
Setting Hydrogen::setSong twice with same value crashes Hydrogen


elpescado:
Hydrogen version *          : all
Operating system + version  : all 
Audio driver + version      : n/a

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Calling [`Hydrogen::setSong`](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/src/hydrogen.cpp#L1860) twice with the same value crashes Hydrogen. That's relevant code:

```
void Hydrogen::setSong( Song *pSong )
{
	assert ( pSong );
	setSelectedPatternNumber( 0 );

	Song* oldSong = getSong();
	if ( oldSong ) {
		delete oldSong;
		oldSong = NULL;
		removeSong();
	}
```
So if `oldSong==pSong`, `oldSong` gets deleted, but as both point to same object, `pSong` is also deleted and thus no longer valid.

AFAIK, no code in Hydrogen (currently!) does that, so that's not user-facing bug, but this issue caused me to waste some time on useless debugging.

Now, there are two ways to fix this, and I'm unsure which one would be preferable:

1. Skip silently (i.e. `if (pSong == oldSong) { return; }`, maybe with some warning log, so that setting the same song twice does nothing.
2. Forbid setting the same song (i.e. `assert(pSong != oldSong)` - that way Hydrogen would still crash, but at least it's obvoius what's going on.

What are your thoughts? Once decided, I'd happily fix this issue.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : 969c42d7faad4db21764012cc2b9508f676feca4

mauser:
Hi! 

I would go for the "skip silently" option, maybe by just adding the pSong != oldsong to the if clause. Have you found the issue with when writing automated tests?

mauser:
Closed after Pull request #646 has been merged.

PR #646, opened 2018-08-27
Hydrogen::setSong - check for setting same song twice

elpescado:
Fix for #641


mauser:
Merged, thanks a lot!

Issue #661, opened 2018-09-26
H2 segfaults when switching to next song while receiving MIDI


thijz:
Hydrogen version *          : be8ecd073b167e569d7a61b881d3a9505014f5eb
Operating system + version  : ubuntu 16.04
Audio driver + version      : jack

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
when you use a playlist you can hit F5/F6 to jump to the next /previous song in the playlist
if you hit F5/F6 while you are sending MIDI to H2 (via a controller interface of soft midi controller) H2 segfaults
this is 100% reproducible

mauser:
Hm.. it segfaults here also without any midi activity while switching songs :smile:

mauser:
I've fixed one rather obvious bug when removing a song, but i'm not sure if this was also your problem which is described in this issue. Could you re-test this problem with the current master?

thijz:
just retested with the latest code : problem remains
will gather some more data and post it here

thijz:
running h2 in gdb gives me this when h2 segfaults : 


> Thread 12 "hydrogen" received signal SIGSEGV, Segmentation fault.
> [Switching to Thread 0x7fffc4e1e700 (LWP 12518)]
> 0x0000000000505f4a in H2Core::Song::get_instrument_list (this=0x0) at /home/tvanseve/Code/prive/hydrogenMay2018/src/core/include/hydrogen/basics/song.h:130
> 130				return __instrument_list;

elpescado:
Changing song involves unloading old drumkit and replacing it with one freshly loaded from new song. Maybe Hydrogen receives MIDI message while no drumkit is loaded?

Best regards,
Przemek

thijz:
Ok i'm in trouble :-(
My BCR controller is acting up and 2 knobs just jitter more and more : they randomly jump up/down 1 without me touching the knobs and thus send midi to H2 at random times. 
This is what made me log this issue, but it has gotten wors, a lot worse) ... as a consequence it is now nearly impossible to switch between songs using F5/F6.
To make matters worse i have a gig this weekend and i have no time to get a new BCR + config it properly (i have a highly customized config on the BCR) ... 
Would there be an easy fix for this issue ?

I am trying to see if i can temporarily disconnect the BCR from H2 during a switch to the next song, but it's not that simple to do so all help is highly appreciated !

mauser:
[...]

PR #663, opened 2018-10-02
Fix MIDI change pattern crash

elpescado:
Hi,

I had a quick look at #661. I managed to reproduce this error:

```
* thread #10, stop reason = EXC_BAD_ACCESS (code=1, address=0x50)
    frame #0: 0x0000000100035fcc hydrogen`H2Core::Song::get_instrument_list(this=0x0000000000000000) at song.h:130
   127 	
   128 			InstrumentList* get_instrument_list()
   129 			{
-> 130 				return __instrument_list;
   131 			}
   132 			void set_instrument_list( InstrumentList* list )
   133 			{
Target 0: (hydrogen) stopped.
```

As I supposed, this happened when MIDI message arrived when song hasn't been loaded yet.

This fix is modular, three commits are independent and can be cherry-picked individually.

a6169d3 skips note on/off MIDI messages if no song is loaded - and this commit alone fixes this issue.

502f556 skips all MIDI messages if no song is loaded. I didn't manage to reproduce situation when this fix is necessary, however I imagine CC message might, for example, modify instrument volume, and when no song is loaded there are no instruments Hydrogen would crash.

62b9b2a makes pointer to `Song` atomic, so that reading/writing song pointer should be synchronized across threads. On current hardware it shouldn't incur performance penalty, as is should use atomic CPU instructions.

### DIFF
--- a/src/core/include/hydrogen/hydrogen.h
+++ b/src/core/include/hydrogen/hydrogen.h
@@ -22,6 +22,7 @@
 #ifndef HYDROGEN_H
 #define HYDROGEN_H
 
+#include <atomic>
 #include <stdint.h> // for uint32_t et al
 #include "hydrogen/config.h"
 #include <hydrogen/midi_action.h>
@@ -238,7 +239,7 @@ public:
 private:
 	static Hydrogen* __instance;
 
-	Song*	__song; /// < Current song
+	std::atomic<Song*>	__song; /// < Current song
 
 	void initBeatcounter(void);
 
--- a/src/core/src/IO/midi_input.cpp
+++ b/src/core/src/IO/midi_input.cpp
@@ -220,6 +220,11 @@ void MidiInput::handleNoteOnMessage( const MidiMessage& msg )
 	MidiActionManager * aH = MidiActionManager::get_instance();
 	MidiMap * mM = MidiMap::get_instance();
 	Hydrogen *pEngine = Hydrogen::get_instance();
+	Song *pSong = pEngine->getSong();
+	if ( pSong == nullptr ) {
+		ERRORLOG( "No song loaded, skipping note" );
+		return;
+	}
 
 	pEngine->lastMidiEvent = "NOTE";
 	pEngine->lastMidiEventParameter = msg.m_nData1;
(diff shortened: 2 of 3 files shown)


thijz:
Thanks for this fix @elpescado  !
You saved my life !  ;-)

If i understand correctly the second fix ( 502f556 ) blocks _all_ midi messages : note on/off and CC, right ?
I only implemented this fix and it solves my problem.  When i turn a knob on my midi controller (sending CC messages to H2) _and_ i hit F5/F6 at the same time Ii see the error messages "No song loaded, skipping note" and H2 does not crash :-)

I have not yet tested fix3, but will do so after the gig i have this weekend

elpescado:
502f556 blocks all MIDI messages - Note on/off, CC, Program Change, SysEx etc. I am not that familiar with MIDI implementation, and unfortunately currently I don't own any MIDI hardware, so I fear that I may block too much - maybe there are MIDI messages that don't require song to be loaded.

thijz:
> maybe there are MIDI messages that don't require song to be loaded

You could be right, i didnt think of that.
I guess all midi messages that refer to a component of a song (like instruments) should be blocked.  Other messages that only affect 'global' things (master volume, humanize, etc) should not be blocked.

lets investigate some more later.

mauser:
@elpescado: The pessimistic approach to blocking midi messages is absolutely ok, since every kind of midi message can be mapped to a midi action which may modify internal state. 

In general: There are multiple threads around (Midi Driver, Audio Driver, User interface, OSC Server..). Currently there is only the audio engine locking mechanism for synchronisation. I have thought about a central, real-time capable event-mechanism when working on OSC but i would like to tear apart the AudioEngine from the Hydrogen class before that. In my opinion that would be sth. for the next release of Hydrogen, since it would require a lot of testing and performance profiling. 

I'm absolutely ok with the first two commits, but can we keep the std::atomic commit out of this pull request? The semantics are currently not clear to me.. which benefits would we get without using load/store ? I would prefer to discuss that separately :)

mauser:
Ah, and of course.. thanks a lot for helping saving @thijz and proposing those fixes :smiley:

mauser:
Hi! 

 502f556 is part of the master branch, i've cherry-picked it from your branch. 

Ah,ok.. so std::atomic was not meant to protect all properties of the song, but only the pointer itself when a  new song is being created.

PR #664, opened 2018-10-14
Controlling multiple note properties + bug fixes in pattern editor

theGreatWhiteShark:
Lately I revisited some old songs for which I set the velocity
properties of the individual notes by hand. Now I want them all to
have a common offset and to leave the rest to the humanizer (apart
from some highlights). But changing all values separately is way too
tedious. Therefore, I implemented a way the properties of all notes in
one pattern (velocity, probability, pan etc.) can be set at once.

When holding the SHIFT key and moving the mouse wheel over the ruler
of the pattern editor all notes are affected by the changes. When
pressing the CTRL key too, the change is not +/- 0.05 but +/- 0.01 for
all of them. Holding the SHIFT key and dragging a property of one of
the notes all others are set to exactly the same value.

In addition, I fixed some bugs, which in the ruler:
- Pressing CTRL did only had an effect on the mouse wheel if no other
  modifier key was pressed
- Adjusting the **pan** and **lead and lag** was not working using the
  mouse wheel since the could did not change the proper
  variables. Instead of assigning the old and new values to *__pan_R*
  or *__oldPan_R*, new variables called *pan_R* and *old_pan_R* were
  defined and used. The **startUndoAction** function did, however,
  only look at the former variables, was unaware of the new
  assignments, and reverted the changes. This was fixed by moving to
  the *__* type of variable names. In addition, the panning was
  rewritten in a more readable (less obscure) way.
- A number of property changes did not feature the print of the
  applied changes as a status bar message. I added messages for all of
  them (mouse wheel and mouse click + movement)
- The status bar message was adjusted for tiny window sizes. When
  working in a tiling window manager Hydrogen, as well as its status
  bar, can get quite tiny quite quickly. To still properly see all
  changes, I moved the numerical value in front of the message
  followed by the description of the changes

What is missing is a nice way to revert the changes. This should in my
opinion be done in a batch way be reverting all applied changes at
once. But I didn't had the time to look into this matter yet.


elpescado:
Hi,

I gave a quick look at your changes because I'm busy right now to do a thorough review, and it looks good to me. I like that your code is well documented. That's what caught my attention:

# General remarks

You store changes in `std::list`. Nowadays, when choosing a STL container, it's advised to stick to `std::vector` (or `std::array` if number of elements is known at compile time). That's because `std::vector` stores elements in contiguous chunk of memory ("spatial locality") which is very cache-friendly. I'll add some interesting articles/talks about that matter later.

Next, 

```
for ( auto it = __propertyChangesStack.begin(); it != __propertyChangesStack.end(); ++it ){
    ...
}
```

could be written in C++11 as:

```
for (auto& change : _propertyChangesStack) {
    ...
}
```

saving a few keystrokes.

# Undo implementation

When working on velocity automation, I noticed that undo/redo handling in Hydrogen is not entirely correct. When handling user action, Hydrogen (1) performs that action and then (2)pushes `QUndoCommand` to undo that action on `QUndoStack`. But `QUndoStack::push` executes `redo` method - so every action is performed twice. That's not incorrect, as most (all?) actions are idempotent (can be executed more than once and end result is the same), there's some unneeded work done.

That two items should not make noticeable performance problems, however. At least on modern hardware.

# Excess redrawing

I think performance issues are because there's too much redrawing. `SE_editNotePropertiesAction::redo` calls `undoRedoNotePropertiesEditAction()` for each `NoteProperties`, and `undoRedoNotePropertiesEditAction()` calls `updateEditor()`. So that if you undo action that modifies 12 notes, panel is redrawn 16 times, I think.

elpescado:
There's an excellent talk by Chandler Carruth at [CppCon 2014](https://youtu.be/fHNmRkzxHWs?t=2160), and another talk by Bjarne Stroustrup [GoingNative 2012](https://www.youtube.com/watch?v=YQs6IC-vgmo) about `std::list` vs `std::vector`.

elpescado:
There's an excellent talk by Chandler Carruth at [CppCon 2014](https://youtu.be/fHNmRkzxHWs?t=2160), and another talk by Bjarne Stroustrup [GoingNative 2012](https://www.youtube.com/watch?v=YQs6IC-vgmo) about `std::list` vs `std::vector`.

theGreatWhiteShark:
Ah, thanks a lot. Also for the video. I will do some profiling and polishing of the code during this week.

theGreatWhiteShark:
Ah, thanks a lot. Also for the video. I will do some profiling and polishing of the code during this week.

theGreatWhiteShark:
The undo/redo of the multiple note edits now works flawlessly (at least at my end).

While the undo/redo for the wheelEvent will be the plain undo/redo of the changes triggered by the user, for the mouseMoveEvent they will be grouped together. All actions between the pressing of the mouse key and its release will be considered as one action.
    
But apart from the physical release of the mouse key, two events will be also considered to be the end of one group and the onset of the next: the pressing of Shift key and the movement of the cursor to a different note. The latter ensures that all changes applied to a single note can be undone or redone in a single action and that the user can edit multiple individual notes without releasing the mouse button. The pressing or releasing of the shift key allows the user to seamlessly switch between single note and multiple notes editing and at the same time to undo/redo all changes in a proper way.

@elpescado moving the `updateEditor()` function call outside of the function reapplying all changes did the job. The new implementation is as fast as the old one, fully documented, and more consistent. Also I ended up sticking to the `std::list` since this way I can easily push at the front and pop at the back. But I enjoyed the videos. Thanks a lot!

theGreatWhiteShark:
I finally had some time to look into the video and changed the linked list into vectors. Thanks a lot for the reference :)

theGreatWhiteShark:
@cme I have a little trouble rebasing this PR since a lot has changed in the `NotePropertiesRuler`. Do you think it is worth the effort (after all you are doing further refactoring of the PatternEditor) or shall we just drop this (or using only parts of it)?

I would vote for dropping since #973 will also cover the feature in here when setting the noise variance to zero. Thus, we might just tackle it later on.

cme:
It seems a shame to lose the work, I didn't think NotePropertiesRuler had changed too much. But I agree that there's probably a lot of overlap with #973 and implementing that would supersede most of this.

theGreatWhiteShark:
Okay, I will close it for now and use it as a basis for #973 . I'm a little bit short on time right now and it's probably best to introduce it all at once properly. Having the UX change too often is also a nuisance.

Issue #411, opened 2016-09-26
Midi exported file has an offset of +24 on Midi notes


jduemes:
_(Note: message edited by @trebmuh for better readability using the so-called "Insert code" tag)_

Hi Folks,

I manually edited the GMKit drumkit.xml file adding midi notes to the instruments. This way I connect (via Jack audio connection Kit) Hydrogen to a keyboard(1) and - as expected - Hydrogen sends the notes to the keyboard.

Now, if I export my song to a Midi file, open this Midi file with Drumstick MIDI Player connected to that same keyboard, it plays very different notes.

I did a test connecting Drumstick MIDI Player to QMidiRoute and QMidiRoute to the keyboard. Then changing the "Note Offset" option on QMidiRoute to -24. Surprise !!! It plays exactly the same notes as using Hydrogen. As I understand the Hydrogen Midi export is buggy.

(1) I used a keyboard but the same happens if I use another midi synthetizer like Qsynth (a Fluydsynth GUI).

```
$ hydrogen --version
0.9.7-'2a5e0cb'
```

```
$ uname -a
Linux myhostname 4.7.4-1-ARCH #1 SMP PREEMPT Thu Sep 15 15:24:29 CEST 2016 x86_64 GNU/Linux
```

Can somebody confirm this or point me an option or workaround to get a correct midi file ?

Cheers,
Jeff

mauser:
Hi! 

I've tried to reproduce the problem, but i couldn't find any issues with the export. Attached is a screenshot of a minimal example, just a fresh song with GMKit and the result in Bitwig. The same note is shown in Hydrogen and Bitwig. 

Nevertheless, the hydrogen midi out / midi export handling can be confusing: #307 has more informations about possible problems. Does your problem persist with every song (even new songs)`or is it affecting only some songs?  

![hydromidiexport](https://cloud.githubusercontent.com/assets/202461/19047612/f0d638f0-89a2-11e6-97a3-0f240843da37.png)

elpescado:
I vaguely remember encountering the same problem. As far as I remember, MIDI defines note 60 to middle C, but does not define "middle C", so different software use C3, C4 and C5 as "middle C" (C3 being most common), and I'm not sure whether Hydrogen is consistent with this.

elpescado:
Hydrogen seems to offset notes by 36 and 60 in different places (60 - 36 = 24 which explains two octave difference):

[...]

PR #666, opened 2018-10-14
Fix MIDI output

elpescado:
I have been looking at #411 and tried to fix Issues with incorrect MIDI values. It seems that Hydrogen sends wrong MIDI note (24 tones lower than it should) - for example it sends 12 instead of 36. I fixed that and I hope I hadn't break anything else.

# Acceptance criteria

1. Instrument should send MIDI Note On message with pich set to MIDI note of 
instrument being played. That Note On message should be followed by
Note Off message with the same pitch.

*Eg. playing instrument with note 36 should play note 36 (C3).*

2. When "Use output note as input note" is **not** enabled,
*n*-th instrument should respond to NoteOn messages
with pitch set to 35+*n*.

*Eg. Note On with pitch 47 (B3) should trigger 12th instrument*

3. When "Use output note as input note" **is** enabled,
each instrument should respond to Note On messages with pitch
set to MIDI note assigned to that instrument.

*Eg. Note On with pitch 56 (G#4) should trigger instrument that
has MIDI note set to 56.*

4. When exporting MIDI file, exported MIDI notes should have
correct pitches (i.e. the same pitches that are assigned to
respective instruments).

*Eg. every note from instrument with MIDI note 38 should
have pitch 38 (D3) in output MIDI file.*

<sub>That case should be covered by automated tests</sub>

# Test results

Hydrogen provides 4 MIDI drivers: ALSA, CoreMIDI, Jack and PortMIDI. Unfortunately, I've been only able to test CoreMIDI, so I need your help! Please write your test results along with driver name in comment, and I'll update the table:

|    | ALSA | CoreMIDI | Jack | PortMIDI |
|---|---|---|---|---|
| Case 1 | ✓ | ✓ | ✓ | |
| Case 2 | ✓ | ✓ | ✓ | |
| Case 3 | ✓ | ✓ | ✓ | |
| Case 4 | ✓ | ✓ | ✓ | |

In my testing I've been using [MidiPipe](http://www.subtlesoft.square7.net/MidiPipe.html) utility:

zrzut ekranu 2018-10-14 o 20 47 12

I think there are similar utilities for other platforms.

---
<sub>Note names assume C5 = 60</sub>

### DIFF
--- a/src/core/include/hydrogen/basics/note.h
+++ b/src/core/include/hydrogen/basics/note.h
@@ -44,6 +44,9 @@
 #define LEAD_LAG_MIN            -1.0f
 #define LEAD_LAG_MAX            1.0f
 
+/* Should equal (default __octave + OCTAVE_OFFSET) * KEYS_PER_OCTAVE + default __key */
+#define MIDI_DEFAULT_OFFSET     36
+
 namespace H2Core
 {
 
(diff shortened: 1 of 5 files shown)


jduemes:
I tested your branch (hydrogen-git untagged.7aae0a8a6ad8f7e6f06e.r3325.d1ccfeb2-1)
Jack and Alsa: Check positive for all 4 cases.
```
$ uname -r
4.20.0-arch1-1-ARCH
```

thijz:
i will test when i find the time
great to see that you are working on thes @elpescado  !

PR #673, opened 2018-11-16
Documentation for JackAudioDriver and other functions/objects

### LINKED ISSUE #602: Lost or repeated beats with Non-daw
Hydrogen version * :1.0.0-beta1
Operating system + version : AVlinux 2018.4.2.
Audio driver + version : jackdmp version 1.9.12 tmpdir /dev/shm protocol 8
Non version 1.2.0
Using hydrogen with NON-daw i get lost\double sample triggering.
debug shows:
```
(I) JackAudioDriver::relocateBBT ...
(W) JackAudioDriver::updateTransportInfo Frame offset mismatch; triggering resync in 2 cycles
```

Reproduce:
Start NonSession> start Non-daw and hydrogen, play. If hydrogen is in song mode - mismatch occurs on loop trigger in Non, if in pattern mode - every time pattern loops.
Swing setting does not affect the issue
The same issue I had with 0.9.7, 
 
I'm not 100% sure if its hydrogen issue, but I had same with ardour5&hydrogen and if I remember it right didn't have with qtractor&hydrogen.

theGreatWhiteShark:
Since I wanted to understand how the transport in Hydrogen and its interaction with the JACK server is working anyway, I thought I just document all the stuff I understand to enable future contributors to get started more easily. I mean, the state of documentation is quite dire. But it turned out to be way more than expected...

Note: One of the reasons to look into this is also to fix #602. Indeed, I found some things going wrong in the JACK driver and will fix them during the next weeks. ~But I most likely will base this future fix on this pull request. Not because I want to push this one with all force, but because it would be way to tedious to start from the non-documented version of the driver all over. So, sorry in advance.~

I also removed some deprecated functions and object, which were defined but not used anywhere in the code anymore and fixed some minor bugs, like mismatching variable and initialization types.

Most notable changes:
- I renamed _src/core/include/hydrogen/JackAudioDriver.h_ to _src/core/include/hydrogen_/jack_audio_driver.h. I know this is a dick move but without I could not tell Emacs to jump between source and head file using a single button and wouldn't be done documenting by the end of next week. _I would also propose to name all audio driver headers consistent with their source files._
- I introduced a new file _src/core/include/hydrogen/config.dox_ to document the macros. I couldn't get `Doxygen` to read in config.h.in and the new file is ignored by `cmake` anyway.
- I updated _Doxyfile.in_. I haven't test the documentation for older version but I shouldn't have broken anything.
- I changed a lot of preprocessor check in headers from `#ifdef FLAG` to `#if defined(FLAG) || _DOXYGEN_`. This way I could `doxygen` make to document stuff for which a macro has to be defined. Beforehand only the fallback objects of the drivers were documented
- I fixed the existing documentation. But a lot of warnings (all of the same nature) remains. This is because I can not get `doxygen` to expand the `H2_OBJECT` macro. No idea why. But I think this might be a bug in their code instead of my settings.

Apart from this, none of the actual code is touched. Just comments and Doxygen-style documentation.


src/core/include/hydrogen/audio_engine.h
```diff
@@ -58,17 +74,37 @@ class AudioEngine : public H2Core::Object
...
 
-	/* Mutex locking and unlocking
+	/** Mutex locking of the AudioEngine.
 	 *
-	 * Easy usage:  Use the RIGHT_HERE macro like this...
+	 * It passes the address of the #__engine_mutex object to
+	 * _pthread_mutex_lock()_ (pthread.h) to lock the AudioEngine
+	 * and transfer ownership of the #__engine_mutex to the
+	 * calling thread. Only this very thread and unlock() the
+	 * engine again.
```
jeremyz: I guess can instead of and

src/core/include/hydrogen/IO/TransportInfo.h
```diff
@@ -32,19 +32,64 @@ class TransportInfo : public H2Core::Object
...
+	 */
 	unsigned m_status;
 
+	/** 
+	 * Current transport position in number of frames since the
+	 * beginning of the song.
+	 *
+	 * A __frame__ is a single sample of an audio signal. Thus,
+	 * with a _sample rate_ of 48000Hz 48000 frames will be recorded
+	 * in one second and with a _buffer size_ = 1024 1024
```
jeremyz: a ',' is missing ;)

jeremyz:
that's some serious work you've done !
either you change something regarding my comments,
or I merge it like that, it is really up to you ?

theGreatWhiteShark:
Sure, I will incorporate your remarks

PR #675, opened 2018-11-20
Nsm rework

Houston4444:
Hi, this is a NSM Implementation review.

* Switch nsm_client from core to GUI, because GUI is responsible of opening songs
* Add .h2song extension to file path
* Hide splash screen when launched with NSM
* Give Hydrogen infos to daemon before making gui, but start hydrogen's nsm thread only when GUI is loaded, this way, on Session Manager, hydrogen appears as "open" just after launch and as "ready" only when GUI is ready. 
* Add "switch" capability, whit it, switching nsm session also switch hydrogen song without close/relaunch hydrogen
* Add "dirty" capability, with it the session manager know if songfile has been modified or not
* Add "optional-gui" capability which allow session manager to show/hide main window
* Disable actions "New", "Open", "Open Demo", "Open Recent", "Save as" when launched from NSM as wanted by NSM dev

Notes:
I am the author of [Ray Session](https://github.com/Houston4444/RaySession), alternative to Non Session Manager, I did test these features on the 2 ones.

I didn't report a bug, but I saw that sometimes with your NSM implementation, Hydrogen doesn't open the existing file but an empty song. I seems to not appear with this patch.

Because nsm_client is launched in another thread, and qt widgets doesn't support to be modified in another that this one created them, I used Qt signals to launch open, save, show/hide gui. Maybe you won't like this.

Regards.


elpescado:
Hi,

It seems that this breaks [Mac build](https://travis-ci.org/hydrogen-music/hydrogen/jobs/457455254). As far as I understand, NSM depends on OSC, so it should be disabled in builds without OSC.

Houston4444:
Oh shit, I don't have any Mac, so I can't test this. 
 As you said NSM depends on OSC. 
Non-Session-Manager and Ray Session are both linux only softwares.

Houston4444:
That was breaking build without OSC support. 
The travis Mac build seems to not load liblo-dev, maybe it doesn't exists.
Now it should works.

trebmuh:
@Houston4444 can you please resolve the conflicting 2 files here, and then I'll be able to test this?

mauser:
Hi! 

thanks for the pull request and your work on Hydrogen's NSM code! 

I'm ok with most of the pull request, but currently i don't not understand the need to move the nsm_client to the gui. In my opinion, session management is something which should be in the core, so it don't has to be re-implemented for every UI (like the QT Gui or the CLI). I'm not a user of session management software, but i could imagine that using hydrogen cli with nsm would be interesting if you only want to use Hydrogen as a "player" for h2song files. 

If the synchronisation issues between the core and the gui are the only issue, then in my opinion this should be better fixed by extending/fixing our event queue (or the communication between core and ui in general), maybe we can think together about a better solution.

Houston4444:
Hi.
Thanks for your thanks.
I understand your opinion, I have done it this way because it seems to be the simplest solution,it has been done this way in Ardour (NSM implementation is managed by gui in ardour).

I'll maybe say mistakes (this PR is 6 months old, I can't remember all), but in Hydrogen, songs are opened via GUI, and I don't think there is a written way for core to signal GUI that a song is open. If you decide to manage NSM via core, this should be implemented I think.

> If the synchronisation issues between the core and the gui are the only issue

No, it's not the only one issue. 
As I said "dirty", "optional-gui", and "switch" NSM options can't be implemented in core as I said. Switch is a very nice option for user to gain time at session switch.
GUI has to show user if song has been saved from session manager.
"dirty" state (song contains unsaved changes or not) is managed by GUI in hydrogen.

GUI has to know if this instance is under NSM to disable actions as "New", "Save as", "Open". This limitations clearly informs user that their use will break the session management.

A session management in the core would have the advantage too to can use the same OSC server for NSM implementation and Hydrogen OSC controls.

I am ok to rework NSM implementation in core if I have all the tools I need to communicate from core to gui:
core is under_nsm
song is open
song is saved

Hide splash screen at startup is cool too (Session Manager is saying that hydrogen is opening).
It's certainly possible even from core.

sub26nico:
I've tested this PR with RaySession and Ardour and it works fine, all patterns, songs, Jack Sync, connections, etc..are recalled without missing or bug.

theGreatWhiteShark:
Superceded by #792. 

Thanks a lot for the discussion, testing, and various contributions to the topic!

PR #678, opened 2018-12-09
Fixing bugs in the JACK audio driver

theGreatWhiteShark:
1. the *beat*, *bar*, and *tick* information got misinterpreted as the
   current transport position propagated by the external JACK timebase
   master. But a "tick" is an arbitrary quantization of the beat and
   the last one passed at the current transport position will be
   provided by the BBT information of the JACK server. Therefore it
   almost always lags behind current transport and thus triggering a
   warning described in issue #602 .

   I rewrote the code in such a way the if clause probing for a
   discrepancy between the current transport position + the
   `bbt_frame_offset` and the JACK server transport position is only
   entered after relocations of the JACK transport position. It will
   set the current transport position to the one of the server and
   resets `bbt_frame_offset` to zero. The corner case of an external
   JACK timebase master changing the tempo using the BBT information
   and some relocations taking place afterwards will be covered by
   keeping track of the tick size upon during tempo change and
   remapping the position of the relocation to our internal version
   with a different speed.

   I'm not 100% sure what the *beat*, *bar*, *tick* information are
   actually provided for. Everything works flawless at my end and no
   one of the jack-devel mailing list could/wanted to give me any
   example or use case my implementation would not be able to handle.

2. The internal transport position and the `bbt_frame_offset` variable
   were not properly reset after a relocation by the user took place.

### DIFF
--- a/src/core/include/hydrogen/IO/TransportInfo.h
+++ b/src/core/include/hydrogen/IO/TransportInfo.h
@@ -76,21 +76,22 @@ public:
 	 * minimum duration of a Note as well as the minimum distance
 	 * between two of them.
 	 * 
-	 * It is calculated by the sample rate * 60.0 / ( #m_nBPM *
+	 * It will only be used by the JackAudioDriver and is
+	 * calculated by the sample rate * 60.0 / ( #m_fBPM *
 	 * Song::__resolution ). The factor 60.0 will be used to convert
-	 * the sample rate, which is given in second, into minutes (as the
-	 * #m_nBPM).
+	 * the sample rate, which is given in second, into minutes (as
+	 * the #m_fBPM).
 	 */
-	float m_nTickSize;
+	float m_fTickSize;
 	/** Current tempo in beats per minute. */
-	float m_nBPM;
+	float m_fBPM;
 
 	/**
 	 * Constructor of TransportInfo
 	 *
 	 * - Sets #m_status to TransportInfo::STOPPED
-	 * - Sets #m_nFrames and #m_nTickSize to 0
-	 * - Sets #m_nBPM to 120
+	 * - Sets #m_nFrames and #m_fTickSize to 0
+	 * - Sets #m_fBPM to 120
 	 */
 	TransportInfo();
 	/** Destructor of TransportInfo */
@@ -98,7 +99,7 @@ public:
 	/** Displays general information about the transport state in
 	    the #INFOLOG
 	  *
-	  * Prints out #m_status, #m_nFrames, and #m_nTickSize.
+	  * Prints out #m_status, #m_nFrames, and #m_fTickSize.
 	  */
 	void printInfo();
 };
--- a/src/core/include/hydrogen/IO/jack_audio_driver.h
+++ b/src/core/include/hydrogen/IO/jack_audio_driver.h
@@ -551,7 +551,7 @@ protected:
 	 *                          m_transport.m_fTickSize ) ) 
 	 * \endcode
 	 * - __ticks_per_beat__ :  the output of
-	 * Hydrogen::getTickForHumanPosition() with the __bar__ member
+	 * Hydrogen::getPatternLength() with the __bar__ member
 	 * supplied as input argument.
 	 * - __valid__ : to ( _JackPositionBBT_ | _JackBBTFrameOffset_
 	 * ), transport states defined
--- a/src/core/include/hydrogen/IO/jack_audio_driver.h
+++ b/src/core/include/hydrogen/IO/jack_audio_driver.h
@@ -52,37 +52,19 @@ class Song;
 class Instrument;
 class InstrumentComponent;
 /**
- * JACK (Jack Audio Connection Kit) server driver. 
- *
- * __Timebase Master__:
- *
- * The timebase master is responsible to update additional information
- * in the transport information of the JACK server apart from the
- * transport position in frames (see TransportInfo::m_nFrames if you
- * aren't familiar with frames), like the current beat, bar, tick,
- * tick size, speed etc. Every client can be registered as timebase
- * master by supplying a callback (for Hydrogen this would be
- * jack_timebase_callback()) but there can be at most one timebase
- * master at a time. Having none at all is perfectly fine too. Apart
- * from this additional responsibility, the registered client has no
- * other rights compared to others.
- *
- * After the status of the JACK transport has changed from
- * _JackTransportStarting_ to _JackTransportRolling_, the timebase
- * master needs an additional cycle to update its information.
+ * JACK (Jack Audio Connection Kit) server driver.
  *
  * __Transport Control__:
  *
- * Each JACK client can start and stop the transport or relocate its
- * current position. The request will take place in cycles. During the
- * first the status of the transport changes to
+ * Each JACK client can start and stop the transport or relocate the
+ * current transport position. The request will take place in
+ * cycles. During the first the status of the transport changes to
  * _JackTransportStarting_ to inform all clients a change is about to
  * happen. During the second the status is again
  * _JackTransportRolling_ and the transport position is updated
- * according to the request. But only its raw representation in frames
- * in the _frame_ member. The current timebase master, if present,
- * needs another cycle to update the additional transport
- * information. 
+ * according to the request. The current timebase master (see below),
+ * if present, needs another cycle to update the additional transport
+ * information.
  *
  * Such a relocation request is also triggered when clicking on the
  * timeline or the player control buttons of Hydrogen. Internally,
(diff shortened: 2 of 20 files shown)


trebmuh:
@theGreatWhiteShark this seems to have conflicting files. Could you please at a look at this?

theGreatWhiteShark:
Sure. 

This pull request did not depend on the documentation, which was merged earlier on. Therefore it needed to be properly rebased.

trebmuh:
@elpescado @jeremyz @mauser : would anyone of you review this one and merge it if fine please?

elpescado:
I have no experience with Jack, so I am not able to thoroughly evaluate this pull request, but from pure coding point of view it looks OK. I can't merge, I'm just a humble contributor;)

trebmuh:
( _I forgot earlier to ping @pvint as well_ )

theGreatWhiteShark:
If it helps I also can try to explain the tweaks and what in my opinion needed to be fixed in a QA session via the mailing list.

mauser:
Hi @theGreatWhiteShark! 

i suppose there is currently no special jack expert in our team, and even though trebmuh has requested a review from me, i don't think that i have the technical understanding of the jack internals :) But i just saw your other "jack documentation" pull request and will start from there.

How can the effect of this pull request be tested? Is it "just" the issue described in #602 or does it affect other behavior?

theGreatWhiteShark:
Hi mauser,

http://jackaudio.org/files/docs/html/transport-design.html provides the basic introduction of how transport is working with JACK. Apart from this I would recommend the [API reference](https://jackaudio.github.io/api/). As with everything, the JACK system is not that complicated once you understood how it works. But they clearly lack some nice and easy accessible introductory-level documentation.

Yes, the biggest issue are the lost beats described in issue #602 (as far as I recall). I used mainly Non-Timeline to test the behavior in the presence of another timebase master and the big problem is that the beat, bar, tick information is not precise but rounded. This will cause a mismatch in H2. I actually have no idea what this information can be used for since it seems quite redundant given all the other ones in the JACK transport information. I also asked for some clues on the jack-devel mailing list but no one provided some insights.

mauser:
Hi! 

i started to test the branch today, and at least for my limited use cases, it works really fine! I don't get any lost beats and the Hydrogen behaves as expected. 

@thijz: I believe you're using Hydrogen a lot with jack sync, could you give this branch a try?

mauser:
Hi,

i spent some time today playing around with some Jack Applications and this pull request. When using non-timeline, Hydrogen as a slave application (or Level 1, or whatever it is called now :smiley) worked as expected. Hydrogen picked up the tempo from the master and relocating the transport worked well, i could set it to any position in the song and Hydrogen acted as expected.   

When using Hydrogen as a timebase master, non-timeline did not pick up the tempo which i set in Hydrogen. I'm not sure if this is due to non-timeline or a problem with Hydrogen. 

I wanted to test it as well with Bitwig and Ardour5, but the later wasn't even able to start on my system. The jack transport in Bitwig seems to be quite limited. I could not get Hydrogen to pick up the tempo, but Start/Stop with Bitwig as Master seemed to work (most of the times..). I can see if i can do further tests with other applicatons..

theGreatWhiteShark:
Hey mauser,

I will double check the code and its behavior soon. After all, it was both one of my first peace of code written in C++ and I was just starting to introduce myself in Hydrogen.

It's quite a shame I don't have any time left anymore when development is rolling again in the project. But I will probably have more time in a month and I promise to get my pull requests straight as soon as possible :)

Also, I never tried having two instances of Hydrogen working together as timebase master and the other one as normal JACK client

theGreatWhiteShark:
There is still a conceptional flaw in the JACK timebase design of Hydrogen: Clicking the "J.MASTER" button in the GUI will result in Hydrogen getting registered as timebase master. But as soon as another application does register itself (which is nothing H2 has any influence on), H2 won't take "any" notice. At least the button stays active. As a consequence when running two H2 instances simultaneously they both tell the user they are timebase master while there can only be one. I will (try to) take care of it

theGreatWhiteShark:
There is still something off. When I start two Hydrogen instances and change the transport position in an arbitrary JACK client, the Hydrogen instance which is not JACK timebase master sort of freezes and just outputs artifacts. I'll look into it

theGreatWhiteShark:
Okay. It took me a while but I do now understand what's going on in the code. I also found some flaws in the code I wrote a year ago. I'll fix them.

But there still an open design question: Right now Hydrogen will always use the tempo provided by the timeline. If it is timebase master, it will propagate its value to all JACK clients. But if it is the listening to an external timebase master ifself, it will effectively discard the speed information. This doesn't seem right. I would suggest to disregard the (user-defined) timeline tempo in the presence of an external JACK timebase master. If the user is not happy about this, all she has to do is to press the "J.MASTER" button. I would also turn the timeline tempo thingy grey/make it unresponsive to highlight this change in state and would add some additional information to the manual.

@mauser what do you think?

mauser:
Hi! 

> I would suggest to disregard the (user-defined) timeline tempo in the presence of an external JACK timebase master. 

That makes sense to me. Maybe you could add an tooltip to the greyed-out button which informs the user that the timeline won't be available if the Hydrogen is not timebase master. There could be also a ERRORLOG message with the same intention. A message box might be a little too much.

theGreatWhiteShark:
Okay, I'm done rewriting the JACK support. What a mess.

There are basically two problems:
1. The handling of tempo in H2 is all over the place. Literally. H2 has to basically run a full cycle in order to adapt to the changes due to the Timeline.
2. JACK's frame-based handling of transport is inherently incompatible with the tick-based one of Hydrogen.

Whenever a tempo marker is passed, the tick size will be changed, and the internal transport position will be rescaled. (the current tickNumber is preserved)
```
frame_{post} = frame_{pre}* oldBPM/newBPM
```
But due to the factor in the second term, the rescaling is depending on the current transport position. When waiting  another cycle before rescaling we have
```
frame_{post} + bufferSize != (frame_{pre} + bufferSize) * oldBPM/newBPM
```
So, to keep two H2 instances in sync (one as timebase master and the other one as normal JACK client - as placeholders for other applications), one has to exactly know during which cycle the change in tempo occurred. This gave me quite some headache since there are quite a number of edge cases involved that have to be dealt with separately.

**former implementation**
The code in `master` deals with 1. by introducing *Humantime* positions -> JACK does just wait one or two (in case of relocating by mouse press) cycles until it adopts to changes. But this is a) quite dirty and b) makes it hard or impossible sync to H2 due to 2. 

Also whenever H2 is a normal JACK client and another application is timebase master, it misinterprets the bar-beat-tick information send by the JACK server as current transport position. This causes the lost beats in various issues.

**patch**
I tried to touch as little code outside of the JACK driver as possible - apart from renaming and refactoring. Now, JACK solely uses the current transport position and speed and does not wait till things settle during the remainder of the cycle in the audio engine. It, instead, predicts what the audio engine will do and takes appropriate measures. 

If we could make the Timeline frame-based instead of tick-based, it would simplify both the JACK audio driver and the overall handling of tempo by a margin. But this is for another PR.

@mauser, @trebmuh, @thijz could someone of you test this with MIDI input? There is the *realtime* position stuff in the code base, which is related to handling MIDI while the transport is not rolling. But since I never used H2 with MIDI input, I couldn't test (with resonable effort) if everything still works fine at this end.

## Known bugs
There are a couple of things, which are not possible with our current design.

**Launching two H2 with timebase master enabled and starting JACK transport in a region of different tempo (with respect to current speed) will make them go out of sync.**
Unfortunately the JACK API does not provide a way to check whether a application is still the timebase master. It just won't call the corresponding callback anymore. Therefore, the H2 launched first will only notice during the second transport cycle that it is not timebase master anymore and problem 2. kicks in.

**Changing tempo in timebase master will make clients go out of sync**
Due to 2. H2 will broadcast the speed it will use during the *next* cycle in the timebase information. Else, two H2 could not be able to be in sync at all. But when changing the speed by hand, we could only look into the future by storing it and wait one cycle till we adopt it. But this would impact behavior other than the JACK driver and I didn't want to do this for this edge case.

**Activating/Deactivating the timebase master causes jumps in the transport position**
When registering H2 as timebase master, JACK does not increment the current transport position by the buffer size but decrement it (??). This will be recognized as a relocating and resets the frame offset required to adjust the internal frame position to the one JACK provides. But this seems more like a issue of JACK itself

In all of the three cases above a click at an arbitrary position on the Transport ruler makes all clients go into sync again.

mauser:
Perfect! I'm starting to test this and will report back!

mauser:
Hi! 

I've finally merged this pull request after testing again with Bitwig, Ardour5 and non-timeline, and as far as i can see, everything works fine! 

Concerning the open issus with Hydrogen beeing the timemaster: I suppose those are minor issues, compared to the role of Hydrogen as a recipient of tempo/transport information. It seems that there are not many use cases where it really makes sense to use Hydrogen as a timebase master..

theGreatWhiteShark:
Perfect! I couldn't get much of these applications running on my (old) system. Only Qtractor. But it has major issue with the JACK timebase support itself and was not a proper tool for testing. But I upgraded quite recently and will make some tests as well just to be sure. :)

trebmuh:
:+1:

theGreatWhiteShark:
I done some testing and in general it's working.

When using Hydrogen in combination with Ardour one gets different results depending on which one is the timebase master (TBM). Ardour as TBM send the speed information always precisely on time. Hydrogen, on the other hand, does send the tempo change a little bit premature (partly due to #864 and partly because a small tweak I still have to do. But I will tackle the first one first). But everything is fully reproducible and you will only notice when changing TBM.

Non-Timeline does send the tempo information prematurely too. So I guess no one would ever notice. But I will fix it together with #864 nevertheless.

PR #688, opened 2019-01-19
Enable HiDPI scaling

hfiguiere:
Enable HiDPI

https://doc.qt.io/qt-5/highdpi.html

### DIFF
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -142,7 +142,10 @@ static void setApplicationIcon(QApplication *app)
 
 int main(int argc, char *argv[])
 {
-	try {	
+	try {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
+		QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+#endif
 		QApplication* pQApp = new QApplication( argc, argv );
 		pQApp->setApplicationName( "Hydrogen" );
 		pQApp->setApplicationVersion( QString::fromStdString( H2Core::get_version() ) );


hfiguiere:
FYI the build fail on TravisCI on Linux  because a 5 year old Ubuntu doesn't have Qt 5.6 but 5.2

hfiguiere:
Added a version check to allow build on TravisCI

mauser:
> > FYI the build fail on TravisCI on Linux because a 5 year old Ubuntu doesn't have Qt 5.6 but 5.2
> 
> Maybe we could switch to [Ubuntu 16.04 LTS](https://docs.travis-ci.com/user/reference/xenial) on Travis CI.

Done. This gives us QT 5.5.1. It seems that currently no newer version of Ubuntu is supported on Travis.

mauser:
Hi everyone, 

i don't own a device where this option could make a difference. Could anyone with such a device test this PR and report back?

mauser:
Hi! 

I will merge this now so it can appear in 1.0-beta2, maybe we can get then better test-feedback from our windows users with 4k screens.

mauser:
... and last but not least: Thanks a lot for this pull request and for tackling that problem!

cme:
It looks like Qt has some awkward behaviour with this. Windows picks default scaling factors for users which include fractional ones such as 1.25 and 1.5. If we declare the app to be HighDPI aware, Qt will round this to the nearest integer by default, which will give a display that feels slightly too small (at 1.25 rounded down) or much too large (rounding 1.5 up to 2.0) which is the case encountered in #914. 

It's disappointing that this didn't get found by the beta or RC's. It's hard to know how much uptake those got, but I'd have thought someone would have run across this. But Windows is a minority platform I guess. :/

Setting the rounding to pass-through a non-integer value makes the UI look terrible because Qt does nearest-neighbour scaling. In this case it would be preferable to use Windows' own scaling for DPI-unaware apps since it at least anti-aliases.

So a reasonable fix would seem to be to check for a non-integer scaling factor and *only* setting the app as hidpi aware when the scaling factor is an integer or close enough to not be noticably bulky.

I think there's also an argument to be made that Qt should round to the integer scaling factor giving the nearest *area* rather than just the nearest integer scaling factor, but that's a whole other issue....

hfiguiere:
Maybe this could be used:

https://doc.qt.io/qt-5/qt.html#HighDpiScaleFactorRoundingPolicy-enum

It require Qt 5.14

elpescado:
https://lists.qt-project.org/pipermail/development/2019-September/037434.html

cme:
So, it seems like a sensible fix would be to:
  - pick a more sensible default (floor, or prefer floor) and
  - add a user preference to allow the user to pick between the default, ceiling, or passthrough.

@elpescado , I guess there might be some intersection with any resolution to the issues in #916. If I implement the fix above, would there be anything in #916 that might conflict with or supersede it?

elpescado:
I think that #916 is unrelated issue, so I do not anticipate any conflicts.

BTW, AFAIR macOS only uses integer scaling.

PR #706, opened 2019-03-28
fix: mute midi on & off events when instrument is muted

### LINKED ISSUE #704: Muting an instrument does not mute midi events
Hydrogen version *          : 0.9.7
Operating system + version  : Ubuntu 18 LTS
Audio driver + version      : pulseaudio 11.1

This is related to https://github.com/hydrogen-music/hydrogen/issues/266 but not exactly.

Muting an instrument using the `M` button mutes the audio but not the MIDI events.

I am referring to this `M` button : 
image

Midi events are still sent to the `Hydrogen` MIDI output.

I am referring to this MIDI output : 
image

Is this expected? If yes, is there a known workaround so that MIDI events are not sent for a muted instrument?

julienmoumne:
This fixes #704 for my specific requirement, I've successfully built and manually tested this patch.

I have left a TODO in the code because I don't have the full picture.

Also I am not fluent in cpp so some refactoring must be done.

### DIFF
--- a/src/core/src/sampler/sampler.cpp
+++ b/src/core/src/sampler/sampler.cpp
@@ -156,8 +156,12 @@ void Sampler::process( uint32_t nFrames, Song* pSong )
 
 	while ( !__queuedNoteOffs.empty() ) {
 		pNote =  __queuedNoteOffs[0];
+		// todo should we apply the same logic as in Sampler::__render_note ?
+//        bool isMutedForExport = (pEngine->getIsExportSessionActive() && !pInstr->is_currently_exported());
+//        bool isMuted = isMutedForExport || pNote->get_instrument()->is_muted() || pSong->__is_muted || pMainCompo->is_muted();
+		bool isMuted = pNote->get_instrument()->is_muted();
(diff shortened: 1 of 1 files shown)


mauser:
> You might like to, for the sake of completeness, look at [MIDI export](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/src/smf/smf.cpp) and implement this change also there.

Good point! 

Apart from that, i would be ok with the pull request! As a small sidenote: Since we switched to C++11, you might want to use `nullptr` instead of `NULL`. But this is just a minor thing..

julienmoumne:
thanks for the review

I'm trying to fix the export as you mentioned

Before doing so, I tried exporting the `GM kit demo #3` (Project > Export Midi File) and importing it into rosegarden,

here is what I have 

image

I can't make it export more than one note as can be seen on the picture so I can not start fixing the code.

Are we talking about the same midi exporting feature?

mauser:
Hello Julien,

Thanks for the pull request! Since i couldn't merge it automatically, i've included it in commit b734c4caae71783f5c719f03de609450f4fbbfa2.

I will see about the midi export.. I'm currently not even sure if everyone would expect that muted instruments are not exported..

PR #713, opened 2019-04-04
Moving INSTALL.txt to INSTALL.md so it is rendered in github site

sdavtaker:
The INSTALL file can be rendered in the github page if the extension of the file is changed to .md
I tested in my forked repository and it seems to work fine.


mauser:
Thanks! It seems that some sections need some layout-editing, but currently it seems that is good enough to be read and understood.

Issue #682, opened 2018-12-22
Unable to see and edit song mode in any of the available hydrogen versions for Windows 10


pseudomoomin:
Hydrogen version *          : I have tried all of the available versions
Operating system + version  : Windows 10 64 bit professional
Audio driver + version      : ASIO

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

see the title

pseudomoomin:
picture of the GUI I am getting http://prntscr.com/lyccf3

theGreatWhiteShark:
This is strange. Does this happen after startup? If not, can you give a step-by-step guide to reproduce the issue.

In addition, can you run hydrogen with the command line arguments `hydrogen --verbose=DEBUG` and paste the output in here?

trebmuh:
@pseudomoomin : don't you have a menu such as:

![hydro](https://user-images.githubusercontent.com/8705846/55124913-41616180-5108-11e9-96dc-13bd3521ab62.png)

pseudomoomin:
> 
> 
> This is strange. Does this happen after startup? If not, can you give a step-by-step guide to reproduce the issue.

Yes after startup it is immediately like this.

Olli:
I have the same problem as @pseudomoomin but in Debian Linux. 
The debug output of mine is https://gist.github.com/Olli/1b26d1dc958d9348fb600416c2fe317f
I hope it helps.
For me I did build hydrogen from the source. Before that I completely uninstall it from /usr/local/ and removed the .hydrogen user files.

theGreatWhiteShark:
Hi,

@Olli, thanks to you log I was able to reproduce the issue.

The problem is the following: Somehow H2 was not properly installed on your system and the default configuration file in _/usr/[local/]/share/hydrogen/data/hydrogen.default.conf_ is not present. When starting up H2 for the first time the config file in _~/.hydrogen/hydrogen.conf_ gets initialized with some improper parameters.

Changing `<maxBars>0</maxBars>` to `<maxBars>400</maxBars>` in the _hydrogen.conf_ file should fix the issue. @pseudomoomin: I'm not perfectly sure were this file is located on Windows systems but this should solve the issue for you too.

Olli:
Hm the missing hydrogen.default.conf was something I also noticed but I thought "Never mind" ... 
[...]

PR #722, opened 2019-04-13
Fixing bug in config recreation

theGreatWhiteShark:
As reported in issue #682 the rendering of H2 does not work properly if the hydrogen.default.conf file was not properly installed.

If neither the system-level config file nor the one in the user directory are not present, H2 will recreate the user config using the default parameters set in the `Preferences::Preferences()` constructor. Since the parameters `maxLayers` and `maxBars` are not set in there, they will be initialized with `0`. This causes H2 to not be properly displayed and the user has no way of changing it but to manually rewrite her config file.

In addition introduced some documentation and made the naming of the `m_iMaxLayers` variable consistent throughout the code base.

It is possible that other parameters, which are not present in the Preferences constructor either might cause additional bugs? Should we/I do some preventive measures and set ALL private parameters in the constructors or would this be an overkill?

### DIFF
--- a/src/core/include/hydrogen/Preferences.h
+++ b/src/core/include/hydrogen/Preferences.h
@@ -460,11 +460,15 @@ public:
 	bool				useLash();
 	void				setUseLash( bool b );
 
-	void				setMaxBars( int bars );
-	int				getMaxBars();
+	/** @param bars Sets #m_iMaxBars.*/
+	void				setMaxBars( const int bars );
+	/** @return #m_iMaxBars.*/
+	int				getMaxBars() const;
 
-	void				setMaxLayers( int layers );
-	int				getMaxLayers();
+	/** @param layers Sets #m_iMaxLayers.*/
+	void				setMaxLayers( const int layers );
+	/** @return #m_iMaxLayers.*/
+	int				getMaxLayers() const;
 
 	void				setWaitForSessionHandler(bool value);
 	bool				getWaitForSessionHandler();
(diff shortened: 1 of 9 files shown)


mauser:
Hi! 

thanks a lot for the pull request / bugfix! 

> In addition introduced some documentation and made the naming of the m_iMaxLayers variable consistent throughout the code base.

I will change that to m_nMaxLayers for consistency :) But yes, as you already mentioned in your other pull request, i will try to expand the coding style part of the DEVELOPERS document.

> It is possible that other parameters, which are not present in the Preferences constructor either might cause additional bugs? Should we/I do some preventive measures and set ALL private parameters in the constructors or would this be an overkill?

I haven't checked the code, but i would suspect that the same effect may happen to other parameters and that those bugs can be hard to spot :-/  Setting all members of the class would be ok for me, in my opinion this is good practice, not overkill :) In addition, it seems that there are also some public members which are beeing adressed directly and not via getters/setters. At least at the moment i can't see the logic behind that..

mauser:
Ah, one additional thing. Could you set the tabwidth of your editor to 4? I will add that to the developers file.

theGreatWhiteShark:
> Could you set the tabwidth of your editor to 4? I will add that to the developers file.

YES! Absolutely! I specifically increased it to 8 for H2 to have the same tabwidth as used in the overall code base and I do detest it.

Okay, I will open an issue about getter/setter functions and the initialization of the members and assign myself to keep an eye on this matter.

PR #724, opened 2019-04-14
Driver classes

theGreatWhiteShark:
Although I do not have a strong opinion on what is the best naming schema/coding standard I find that at times the code base of H2 is quite confusing. 

It took me some minutes to figure out how to access the audio driver. Not only is it a _global_ object in _hydrogen.cpp_ while everyting else is either a singleton or one of its member, the function returning `m_pAudioDriver` is called `getAudioOutput()` and former is of class `AudioOutput`. I rewrote parts of the code and moved the audio and midi driver inside the `Hydrogen` instance, renamed the midi ones to `m_pMidiDriverInput` and `m_pMidiDriverOutput`, wrote corresponding get* and set* functions, and renamed the corresponding classes including the files they are located in.

Honestly, I do not enjoy rewriting existing code and would rather tackle some patches or other projects, like the Humanizer. But IMHO the inconsistency of the code base, the quite thin coverage of documentation, and the lack of a hosted version of the code's DOxygen doc is the main reason why not that many people do contribute to this project. E.g. `pHydrogen` and `pEngine` are both used to represent the same thing in different parts of the code. I do not want to dedicate my time in rewriting the code base but if we would come up with a unified coding style for the overall project (maybe by elaborating the /DEVELOPERS file) I would commit to changing the style of the parts of code I do come in contact with. 

Since such changes are a nuisance to developers who do not keep up with all commits of the projects, I added a NEWS.md file. It might help the project to scale better.

I also do not see why the functions controlling the audioEngine are defined globally in the _hydrogen.cpp_ file and are not part of the `Hydrogen` class. Since they do some heavy processing in terms of realtime audio, the overhead of calling them as member functions should be negligible. But I did not change their scope in the code. It's just puzzling me. Is it a historical thing or do I miss something?


mauser:
Hi! 

First of all, there are different issues contained in this pull request. I will try to adrdess them seperately.

1. Use of auto

I'm not sure if it makes sense to use `auto` everywhere,especially when talking about these changes:

`Song* pSong = pEngine->getSong();`

vs. 

`auto pSong = pEngine->getSong();`

If we would use it everywhere, the code gets (in my opinion) harder to read if every variable is declared "auto".

But it makes a lot of sense for me with more complicated type definitions, like `std::vector<DrumkitComponent*>`
 
That's just my opinion, and maybe a good starting point for a discussion :) 

2. About the NEWS.md: Is this intended to be sth. like Release Notes for developers? Maybe we could put it to the wiki, like https://github.com/hydrogen-music/hydrogen/wiki/New-features-in-Hydrogen-1.0. From my point of view, i think that users would associate different type of information in a file called "NEWS".

3.

> I also do not see why the functions controlling the audioEngine are defined globally in the hydrogen.cpp file and are not part of the Hydrogen class. Since they do some heavy processing in terms of realtime audio, the overhead of calling them as member functions should be negligible. But I did not change their scope in the code. It's just puzzling me. Is it a historical thing or do I miss something?

Good question! This has also puzzled me. I've been not looking much into those audio-related things in the past, but i remember that one argument was that the globally accessed data had to be there *before* the class Hydrogen was initialized completely, otherwise deadlocks could occur. I'm am not sure if this is still the case.

Nontheless, my understanding would be that the parts which you moved into the Hydrogen class would better fit into the AudioEngine class (which, confusingly, also exists in parallel to the global audioEngine_* functions.). The AudioEngine class would then encapsulate the interaction with the audio backends.

4. About the renaming itself: "MidiDriverInput" sound like sth. which produces Input for Midi Driver. Wouldn't InputMidiDriver be a better name? Maybe a native english speaker could comment on this.. 

For the future: I suppose that multiple smaller pull request are easier to handle :) In addition, maybe you can reach more developers via our development mailing list (especially if it is sth. very basic, like this pull request). Not everyone is following all pull requests.

theGreatWhiteShark:
Hey,

Yes, first of all sorry for mixing many things in this pull request. It sort of grew naturally while looking through the code. Plus it was a quite rainy Sunday.

Regarding the use of **auto**: It's of course a matter of personal change. For more complicated types I personally do like them to be written out explicitly to make the code more clear and only used it for "simpler" ones, which can be derived by the variable's name. By I'm fine with dropping all of these. They might be a hindrance for people new to C++.

@ 2. Okay, yes. That's a better way of doing it. I may have a too strong background in package development. But your right. The average H2 user probably does not care for changes in the internals.

@ 4. I choose those names to point out their similarity. But I do not have a strong opinion on the choice of wording.

Well, I'm subscribed to the mailing list. But it's so quite I thought this to be the better place. 
Sure, I send a mail during the next days regarding the 3. point. The `AudioEngine` class is again a different topic. In most parts of the code the local variable holding the instance of the `Hydrogen` class is called `pEngine` and the method for accessing the audio driver was already a member of the `Hydrogen` class in the first place. Quite confusing

theGreatWhiteShark:
superseded

Issue #670, opened 2018-11-12
Playback-Track: Wrong speed for 48000 Hz


desperateCoder:
Hydrogen version *          : 1.0.0-beta1
Operating system + version  : 3.16.0-7-amd64 #1 SMP Debian 3.16.59-1 (2018-10-03) x86_64 GNU/Linux
Audio driver + version      : PulseAudio 5.0

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Hi,
I was struggling over getting my playback to get tight to my grooves (see issue "playback offset").
I finally was able to find out, what went wrong, and i think it's a bug:
The recording guy gave me his exported files at 48000 Hz. This leads to faster playback somehow, like 5 to 10 BPM faster (maybe see issue #37 ?)

what i did to resolve: 
`sox myFancyFile.wav -r 44100 fancyFileWithRightTempo.wav`
this changes the Sample rate down to 44100 Hz, which gave me the constant right speed. This took me a couple of weeks to find out and drove me really crazy, since my old recordings ran like a charm (all @ 44100 Hz)

By the way, this happened also with the current master, i built HEAD and had the same result (except the fancy new colors :D )

Maybe you guys have a look at that?
A quick win would be just showing a warning dialogue as a low hanging fruit, but in long term it would be nice if H2 would play it just like VLC does.

Thanks in advance and sorry for complainig so much :D

trebmuh:
@desperateCoder : so, if I get it right, you're asking if Hydrogen could resample the playback-track?

desperateCoder:
@trebmuh Not at all, it just should respect the sample rate of the playback as it plays, even if the drum-track has a different rate. Resampling could be a possible solution, but who am I to tell you what would be the best :grin:

All I'm saying is that a playback not matching the project sample rate results in pure chaos. 

Did this clarify my problem?

theGreatWhiteShark:
You have a valid point. I encountered this problem before too. But with another software and I would really liked to have some warning popping up.

Luckily checking the sample rate without converting it is something `libsndfile` can handle. We actually already do this in line 166 of the [sample.cpp](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/src/basics/sample.cpp). I will introduce it to the playback-track as well including a pop-up for mismatches.

trebmuh:
> Did this clarify my problem?

Yes it does, thank you.

[...]

PR #727, opened 2019-04-22
Adding documentation for EventQueue and Playback Track

theGreatWhiteShark:
Update:

Since H2 does support resampling the playback track, I removed most of the changes and just kept the documentation and the checking for appropriate file formats.

---

This pull request implements the feature request in #670 

Whenever a playback track is added with a sample rate, which does not match the one of the H2 engine, a warning pop up will be shown with the following content:

> Sample rate of the loaded track does not match the one of the audio driver!
>
>Unfortunately, Hydrogen does not support converting the sample rate itself. Please consider using tools like SoX instead."

In order to check whether a loaded sample does have the same sample rate as the underlying audio driver, a new argument called `checkSampleRate` has been introduced to the `load*()` functions of `Sample`, `Drumkit`, `Instrument`, `InstrumentLayer`, and `InstrumentList`. This option is turned off by default and is currently only used for the playback track.

In principle this can be enabled for all samples the user manually loads into H2. But for short ones as for drumkits this might be an overkill. Or maybe not. I also thought about setting up a popup warning window, which give the user the option to suppress all future popup windows regarding the sample rate using a boolean variable stored in the preferences. Any thoughts regarding this matter?

For now the warning message is generic and does not include the sample rate of the loaded file since the Event class only uses numeric data to transport information. If there is a need for propagating text messages to the GUI in other parts of the code, it might be a good idea to have addition support for QStrings in the Event.

I added an `EventType` called `EVENT_MISMATCHING_SAMPLE_RATE` and the corresponding handler `mismatchingSampleRateEvent()`. I also updated the `QFileDialog` window used to select files for the playback track to a) show detailed information about the files and b) only display the files supported by **libsndfile**. This will prevent users to load e.g. mp3 files and wonder why it does not work.

I also changed the type of `EventQueue::__write_index` and `EventQueue::__read_index` from `int` to `unsigned int` because this better reflects their intended use. And, finally, I added a bunch of documentation.

### DIFF
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -324,7 +324,7 @@ int main(int argc, char *argv[])
 		}
 
 		if ( ! drumkitToLoad.isEmpty() ){
-			Drumkit* drumkitInfo = Drumkit::load_by_name( drumkitToLoad, true );
+			Drumkit* drumkitInfo = Drumkit::load_by_name( drumkitToLoad, true, false );
 			if ( drumkitInfo ) {
 				pHydrogen->loadDrumkit( drumkitInfo );
 			} else {
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -324,7 +324,7 @@ int main(int argc, char *argv[])
 		}
 
 		if ( ! drumkitToLoad.isEmpty() ){
-			Drumkit* drumkitInfo = Drumkit::load_by_name( drumkitToLoad, true, false );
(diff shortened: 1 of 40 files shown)


trebmuh:
> This will prevent users to load e.g. mp3 files and wonder why it does not work.

:+1:

theGreatWhiteShark:
@mauser Well, I think we can keep the documentation. 

I'll throw out all the `checkSampleRate` stuff and push an updated version during the next days.

mauser:
Perfect, thanks a lot!!

PR #729, opened 2019-04-28
Fix layout of INSTALL.md and cover build.sh arguments

theGreatWhiteShark:
The INSTALL.md file was not properly rendered.

I also added some notes about the different command line options of the *build.sh* script. AFAIK there is no hint at all regarding the generation of the documentation and the build using `ccache`. At least I was not aware of it when creating the first pull request.

P.S. I have no idea by [setext headings](https://github.github.com/gfm/#setext-headings) are not properly displayed by Github (anymore). But it's the same on GitLab.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,16 +5,18 @@
 BUILDING AND INSTALLING HYDROGEN
 ================================
 
-Contents:
+Content:
 
-1. System Requirements
-2. Download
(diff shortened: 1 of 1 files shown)


mauser:
Hey ho,

thanks a lot, merged!

trebmuh:
:+1: very nice!

PR #731, opened 2019-05-03
Multiple osc instances

### LINKED ISSUE #723: Segmentation fault when another Hydrogen instance is already running
It's not the first time that I compiled various branches, moved the ~/.hydrogen and /usr/local/share/hydrogen folder in order to find a working version of H2. In the end there was another instance already running on a different workspace from an earlier session.

Instead of reporting a segmentation fault, H2 should tell the user that there is already a H2 instance running and it is not allowed to run a second one.

I'll try to fix it myself.

theGreatWhiteShark:
This fixes #723.

Beforehand it was not possible to start two instances of H2 at the same time since both of them tried to start up an OSC server using the same port. This caused the starting of the second one to result in a segmentation fault.

Now, whenever H2 is about to start up the OSC it uses the port specified in the preferences. If this, however, is not possible, it will let the **liblo** library, which provides the implementation of the OSC server, choose an available port number by itself. This way arbitrary many instances of H2 can be started.

In order to let the `PreferencesDialog` know about the new port, I had to make some changes: 
In `PreferencesDialog::PreferencesDialog()` the settings of the user were reloaded from her ~/.hydrogen/hydrogen.conf file before setting up the PreferencesDialog. But this will discard all the changes to the `Preferences` singleton, which are not saved yet, like the port of the OSC server.

The only reason I could think of why this behavior would be desired is that it ensures consistency between the values shown and the ones in the config file. But I think overwriting the current state with a historic one is the wrong approach of ensuring this.

A way more convenient (and also a more timely one) would be to store the changes in the Preferences as soon as the `PreferencesDialog` gets closed by the user. I am actually surprised myself I have to press ctrl-s to store these changes. But I really like the fact that the dialog is created anew each time it is called. I implemented the automatic save of `Preferences` in the destructor of `PreferencesDialog`.

Apart from the general benefits I think this setting is most convenient for the OSC server itself: If the port is occupied and the user does not care about the actual number, she does not have a look at the PreferencesDialog and the new number will not be changed (if not done expicitly). But if she takes a look at it, the H2 will use the same port number she starts it up again.

Two notes:
- I will add the documentation to all stuff related to OSC in a separate pull request
- This does not allow H2 to run perfectly with many different instances in parallel yet. E.g. the Osc port number should be moved to the Song to ensure reproducible sessions and all instances should use individual temporary folders instead of sharing */tmp/hydrogen*. But I do not plan to pursue this topic any further.

### DIFF
--- a/src/core/include/hydrogen/osc_server.h
+++ b/src/core/include/hydrogen/osc_server.h
@@ -80,7 +80,7 @@ class OscServer : public H2Core::Object
 		 * \param pPreferences Pointer to the Preferences
 		 * singleton.
 		 */
-		static void create_instance(H2Core::Preferences* pPreferences);
+		static void create_instance(H2Core::Preferences* pPref);
 		/**
 		 * Returns a pointer to the current OscServer
 		 * singleton stored in #__instance.
@@ -138,8 +138,8 @@ class OscServer : public H2Core::Object
 	private:
 		OscServer(H2Core::Preferences* pPreferences);
 
-		lo::ServerThread*				m_pServerThread;
-		H2Core::Preferences*			m_pPreferences;
+		lo::ServerThread*		m_pServerThread;
+		H2Core::Preferences*		m_pPref;
 		static std::list<lo_address>	m_pClientRegistry;
 };
 
--- a/src/core/src/osc_server.cpp
+++ b/src/core/src/osc_server.cpp
@@ -242,21 +242,37 @@ int OscServer::generic_handler(const char *	path,
 	return 1;
 }
 
-
-
-OscServer::OscServer( H2Core::Preferences * pPreferences )
+OscServer::OscServer( H2Core::Preferences * pPref )
 	: Object( __class_name )
 {
-	m_pPreferences = pPreferences;
-	int port = m_pPreferences->getOscServerPort();
+	m_pPref = pPref;
+	
+	int port = m_pPref->getOscServerPort();
 
 	m_pServerThread = new lo::ServerThread( port );
+	
+	// If there is already another service registered to the same
+	// port, the OSC server is not valid an can not be started.
+	if ( !m_pServerThread->is_valid() ){
+		
+		delete m_pServerThread;
+		
+		// Instead, let the liblo library choose a working
+		// port on their own (nullptr argument).
+		m_pServerThread = new lo::ServerThread( nullptr );
+		
+		// Get the chosen port number and store it in the
+		// preferences.
+		port = m_pServerThread->port();
+		
+		m_pPref->setOscServerPort( port );
+	}
 }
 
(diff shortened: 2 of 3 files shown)


mauser:
Hi! 

Thanks a lot for figuring out that issue with the OSC api and finding a way to auo

The changed behavior of the preferences dialog save handling gave me some headache :) It seems to me that this little detail can give us some trouble, if there is code which modifies the preferences instance outside of the `PreferencesDialog`. I can only guess, but maybe this was original intention of reloading the values from the config file. In my opinion a well measured use of `const` and a clean API would be a better way to solve that initial problem, but i'm pretty sure that this code sits there since the first versions of Hydrogen.. 

Anyhow... From my point of view, the changed preferences behavior could lead to hard to find bugs and might confuse people. For a "real" multi-instantiable version of hydrogen, one must be able to run hydrogen with different config file locations on startup, like `hydrogen -c /tmp/hydrogen.conf` or sth. like that.

theGreatWhiteShark:
Yes, it gave me quite some headache too. I had already set up an `update` function for the `PreferencesDialog` and a corresponding event to trigger it when I realized it is always created anew with a fresh version of `Preferences`. This took me by surprise and I did not want to rewrite too much of the original code while making a pull request on a completely different topic.

Well, but we already established that the `Preferences` class does not properly initialize every member in its constructor. Maybe we can come up with some cleaner implementation and API of both the class itself and its GUI counterpart and I do implement at some point in the future.

theGreatWhiteShark:
Hey @mauser 

How about I add an `-o, --oscPort` command line option to resolve this issue?

In case Hydrogen is unable to start an OSC server, it will print an error message via `std::cout` and a popup and shuts down afterwards.

mauser:
Hi! 

I have fixed this issue in a similar, but different way (based on your initial code from this pull request). Instead of handling the temporary port in the preferences dialog, an error dialog is displayed which informs the user that the original osc port is busy and that a different port has been chosen temporarily. The new port will be displayed on the command line. 

The port number in the preferences dialog stays the same. 

I've chosen this approach because of two reasons: 
- the fallback mechanism for sound and midi driver works similar
- there is no need to change the preferences handling. 

However, i also see the charm of having a multi-instantiable Hydrogen, especially for use as a plugin. But i guess that's long way to go :(

mauser:
Ah, forgot to add: Thanks for the initial pull request and your code!

theGreatWhiteShark:
Ah, well, having multiple H2 instances is quite an edge case. In the end it was just me wondering why I can not start up the application while having forgotten another instance was already running on a different workspace :)

By the way in newer versions of *liblo* there will be an assertion instead of a segmentation fault https://github.com/radarsat1/liblo/issues/81

PR #732, opened 2019-05-05
Adding documentation and tweaks to OSC server

theGreatWhiteShark:
I added a bunch of doumentation to the OscServer class and its members.

In addition, removed the argument from the Constructor and `create_instance()` function as well as the `m_pPreferences` member. The former is used to pass a fresh copy of the `Preferences` instance and the latter is used to access `Preferencs::m_nOscServerPort`. Both can be done more readable and consistent with the overall code base by using direct calls to `Preferences::get_instance()`.

I also remove some member functions which were neither implemented nor used
- `BPM_CC_RELATIVE_Handler()`
- `BPM_FINE_CC_RELATIVE_Handler()`
- `SELECT_NEXT_PATTERN_CC_ABSOLUTE_Handler()`
- `SELECT_NEXT_PATTERN_RELATIVE_Handler()`

as well as the unused `OscClientInfo` class.

### DIFF
--- a/src/core/src/osc_server.cpp
+++ b/src/core/src/osc_server.cpp
@@ -827,7 +827,7 @@ void OscServer::start()
 	m_pServerThread->start();
 
 
-	INFOLOG(QString("Osc server started. Listening on port %1").arg( m_pPreferences->get_instance()->getOscServerPort() ));
+	INFOLOG(QString("Osc server started. Listening on port %1").arg( m_pPreferences->getOscServerPort() ));
 }
 
 OscServer::~OscServer()
--- a/src/core/src/osc_server.cpp
+++ b/src/core/src/osc_server.cpp
@@ -246,17 +246,18 @@ int OscServer::generic_handler(const char *	path,
 
 
 
-OscServer::OscServer() : Object( __class_name )
+OscServer::OscServer( H2Core::Preferences* pPreferences ) : Object( __class_name )
 {
-	int port = H2Core::Preferences::get_instance()->getOscServerPort();
+	m_pPreferences = pPreferences;
+	int port = m_pPreferences->getOscServerPort();
 
 	m_pServerThread = new lo::ServerThread( port );
 }
 
-void OscServer::create_instance()
+void OscServer::create_instance( H2Core::Preferences* pPreferences )
 {
 	if( __instance == nullptr ) {
-		__instance = new OscServer();
+		__instance = new OscServer( pPreferences );
 	}
 }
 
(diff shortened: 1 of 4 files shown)


mauser:
Hi! 

> The former is used to pass a fresh copy of the `Preferences` instance and the latter is used to access `Preferencs::m_nOscServerPort`. Both can be done more readable and consistent with the overall code base by using direct calls to `Preferences::get_instance()`

The OSC server class is quite new. The passing of the Preferences instance (not a copy of it) was intended in this way, so that this new class does not rely on Preferences beeing a singleton. That would be one my personal goals for the post-1.0 time (not only with Preferences, but with the use of the Singleton pattern in Hydrogen in general).. The current design with nearly everything beeing a singleton brings up a lot of problems with instantiation behavior and timing. Did the passing of the preferences instance did some trouble or was it just for readability? I'm perfectly ok with all the other changes..

theGreatWhiteShark:
Okay, good idea.

Is there a reason you just passed the instance of the `Preferences` and not the one of the `MidiActionManager`?

Would you like me to implement this scheme of passing the instances in other classes I touch or is this a thing we should reserve for later?

src/core/src/osc_server.cpp
```diff
@@ -822,7 +827,7 @@ void OscServer::start()
 	m_pServerThread->start();
 
 
-	INFOLOG(QString("Osc server started. Listening on port %1").arg( m_pPreferences->getOscServerPort() ));
+	INFOLOG(QString("Osc server started. Listening on port %1").arg( m_pPreferences->get_instance()->getOscServerPort() ));
```
mauser: Is this intentional?

mauser:
Perfect, thanks a lot for the update and the documentation! 

> Is there a reason you just passed the instance of the Preferences and not the one of the MidiActionManager? 

Good point, i don't think that there is an technical difference at the moment. Maybe the use of `Preferences` class directly makes it clear that there are no two instances used, but the original from the Preferences singleton. 

> Would you like me to implement this scheme of passing the instances in other classes I touch or is this a thing we should reserve for later?

I suppose it would be more useful after the 1.0 release, along with some other larger refactorings (concerning `AudioEngine` and the `Hydrogen` class)

src/core/src/osc_server.cpp
theGreatWhiteShark: Nope, that one slipped through. Just pushed a fix

theGreatWhiteShark:
Okay, nice.

> I suppose it would be more useful after the 1.0 release, along with some other larger refactorings (concerning AudioEngine and the Hydrogen class)

+1. This certainly needs some proper planning.

mauser:
Merged!

PR #733, opened 2019-05-09
porting stats.py to Python3

### LINKED ISSUE #730: Remove python2 dependencies
data/i18n/stats.py is python2, [which is soon very EOL](https://pythonclock.org/) and needs to be ported to python3 (only).

theGreatWhiteShark:
Fixes #730

### DIFF
--- a/data/i18n/stats.py
+++ b/data/i18n/stats.py
@@ -1,5 +1,9 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
+## Python3 script checking the coverage of the translation files.
+
+import os
+import re
 from xml.dom import minidom, Node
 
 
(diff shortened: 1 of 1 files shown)


mauser:
Hi! 

i'm not a python expert, but from looking at the script, it seems to me that it would also run fine with python2 (os and re should be present there as well). Wouldn't it be ok to leave the first line to 
`#!/usr/bin/env python`  ? I would expect that on a python3 only system, `#!/usr/bin/env python` would also work and point to the python3 binary. Or is this a wrong expectation?

theGreatWhiteShark:
Hey Mauser,

Uh, I'm no python expert myself. But I just found [this](https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take) stackoverflow question giving some insights. 

Since the script is neither intended to be run by the user and just prints some info to stdout, I would vote for getting rid of the sheband altogether. On the other hand, the script still works perfectly fine with python2 from the command line.

mauser:
Hi! 

ok, regarding to the post you mentioned, the use of env in combination with python should be fine for cases in which the script is compatible to python2 and python3. I would like to try it this way to allow users who are still using python2 to work with hydrogen. We have a lot of users and developers who are using older distros..

PR #734, opened 2019-05-12
Adding documentation in audioEngine

theGreatWhiteShark:
Most of the new parts focus on the `audioEngine_updateNoteQueue()` function.

In addition, I removed the `Hydrogen::__get_selected_PatternNumber()` since it was doing exactly the same as `Hydrogen::getSelectedPatternNumber()`


mauser:
Merged!

PR #735, opened 2019-05-26
Class name warnings and documentation for the Object class and logging macros

theGreatWhiteShark:
Unfortunately I could not get Doxygen to work properly with the `H2_OBJECT` macro (although it gets expanded properly). This results in as many warning messages as classes in the code derived from the `Object` class (quite a number).

To deal with the issue, I explicitly introduced the class name `m_sClassName` and its public getter `className()` to every class inheriting from the `Object` class. This is of course way more verbose and not really good style. But the warning messages were quite annoying and were covering an actual bug in the header of `Sample` (fixed).

Since both the private class name and its public getter are inherit from `Object`, it would be enough to only define the members explicitly for all classes having static member functions and to directly hand over the name of the class when creating the object. But I would vote against it since it makes the code less consistent.

In addition, I also added a bunch of documentation for the `Object` class and the logging macros.

### DIFF
--- a/src/core/include/hydrogen/IO/AlsaAudioDriver.h
+++ b/src/core/include/hydrogen/IO/AlsaAudioDriver.h
@@ -37,8 +37,9 @@ typedef int  ( *audioProcessCallback )( uint32_t, void * );
 
 class AlsaAudioDriver : public AudioOutput
 {
-	H2_OBJECT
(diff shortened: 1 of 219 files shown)


theGreatWhiteShark:
superseded by #1319

PR #737, opened 2019-06-02
Doc structure

theGreatWhiteShark:
NOTE: This pull request depends on #735 . I know this is really bad style. But since I had to touch almost all source files in both of them, it would be hell of a nuisance to resolve possible merge conflicts.

The documentation of the source code lacks clear structure. A differentiation into inside and outside the `H2Core` namespace is just not enough for so much classes and functions. Also, all the global `audioEngine_*` functions do not show up in the sidebar at all. 

To resolve this problem, I introduced a couple of groups to bundle classes and functions related to a particular topic. They can now be accessed via the `Modules` node in the sidebar.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,14 +231,19 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 #set(CMAKE_AUTOMOC ON)
 
 
+# Building the developer/API documentation.
 find_package(Doxygen)
 if(DOXYGEN_FOUND)
 	configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
-	add_custom_target(doc_dir ALL COMMAND ${CMAKE_COMMAND}
-		-E make_directory ${CMAKE_CURRENT_BINARY_DIR}/docs)
+	add_custom_target(doc_dir
+	  ALL
+	  COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/docs)
+	add_custom_target(doxygen_dir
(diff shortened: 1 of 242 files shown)


mauser:
Hi! 

I've just started to browse through the latest pull requests. I'm aksing myself if all of those additional information would make more sense in a dedicated part of our wiki (where you have also the possibility to work with grafics, like architecture diagrams). I have the impression that more people are looking at those places then at the doxygen information.

Of course this makes no sense for information which is directly related to specific funtions, but the parts which you put under doxygen/* would be better suited for our wiki (or a md page at our github repo) in my opinion.

theGreatWhiteShark:
Okay. For a little context I would just briefly sketch the direction I intend to go after this contribution.

I do think documentation has to be provided online. Else most people would just miss them (I also just stumbled over them since another contributor pointed me towards writing some for the humanizer code). Therefore I think it would be a great idea to serve the HTML version of the Doxygen doc along with the H2 web page. After all it's a static one and shouldn't be hard to always serve the latest version from Github (incl. the latest version of the manual). E.g. CI with Travis and Github Pages and letting the DNS of the webpage just pointing to the result. The details on this one are on a different page but I think all those documentation I wrote recently are of no use when only to be found locally.

Regarding the wiki: honestly, I'm not a big fan of it myself. As far as I understood it, it is just a separate git repo containing a bunch of flat files and we have to keep up linking them all together manually and enforce some structure on them ourselves. Therefore I often struggle when I have to find information in them. In addition, it holds information about various stuff while its nice to have all information required for debugging and development in one place.

I think for the time being with the Doxygen doc not being served online, one could duplicate the information into the wiki. But, on the other hand, one would have to maintain two versions and it would not be nice to remove one of them with people having already adapted to them.

Well, but in the end I do not care that much about where to find what information. It's all about empowering people to be able to contribute to this project. :)

PR #740, opened 2019-07-11
Update INSTALL.md

alkurop:
Let mac users know that MacPorts is a good installation option


mauser:
Merged, thanks a lot!

Issue #657, opened 2018-09-13
macOS build for hydrogen 1.0.0-beta1


robmckinnon:
Hydrogen version  : 1.0.0-beta1 - 690a82bbbc979c7676f37fda51af8b346206d465
Operating system + version  : macOS Sierra 10.12.6
Audio driver + version      : jack 0.125.0_2

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

For reference, in case it helps others:

I've managed to build `1.0.0-beta1` on **macOS Sierra 10.12.6**, using [Homebrew](https://brew.sh) installed dependencies, with the following steps (note: may be incomplete and not all steps necessary):
```
brew install jack
brew install tar
brew install libtar
brew install libarchive
brew install libsndfile
brew install qt

export Qt5XmlPatterns_DIR=/usr/local/opt/qt/lib/cmake/Qt5XmlPatterns
export Qt5Xml_DIR=/usr/local/opt/qt/lib/cmake/Qt5Xml
export Qt5Widgets_DIR=/usr/local/opt/qt/lib/cmake/Qt5Widgets
export LIBSNDFILE_PATH=/usr/local/opt/libsndfile/include
export CMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk
export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk
export PATH="/usr/local/Cellar/jack/0.125.0_2/bin:/usr/local/opt/libarchive/bin:/usr/local/opt/libsndfile/include:$PATH"
export LIBARCHIVE_INCLUDE_DIRS=/usr/local/opt/libarchive/lib/pkgconfig:/usr/local/opt/libarchive/lib:/usr/local/opt/libarchive/include
```

Then:

```
git clone git@github.com:hydrogen-music/hydrogen.git
cd hydrogen
git checkout 1.0.0-beta1
./build.sh c && ./build.sh m
```

At this point failures to compile, so:
```
cp /usr/local/opt/libsndfile/include/sndfile.hh ./src/core/include/hydrogen/sndfile.hh
cp /usr/local/opt/libsndfile/include/sndfile.h ./src/core/include/hydrogen/sndfile.h
./build.sh c && ./build.sh m
[...]

PR #744, opened 2019-07-16
updating required packages for MacOS

theGreatWhiteShark:
I found myself unable to compile H2 with `brew` and the packages listed in the INSTALL.md. Using the MacPorts equivalents everything works fine.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -94,7 +94,7 @@ libraries and development header files installed on your system:
 - GNU g++ compiler (>=4.0, 3.x might work)
 - cmake (>=2.6)
 - libsndfile >=1.0.18
-- zlib and libtar *OR* libarchive (OS X: You will probably need to build libarchive from source)
+- zlib and libtar *OR* libarchive
 - OS X: Xcode
 - At least one of the following audio and midi driver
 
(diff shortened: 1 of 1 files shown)


mauser:
Hi! 

as a sidenote: our own travis build uses brew, which compiles fine.

Maybe we could list the requirements for both systems? I feel that this is not the first time that we switch between macports and brew (or other systems..) :)

theGreatWhiteShark:
Well, for some reason building it with `brew` is quite an hassle. Maybe it's because I do use a different OS version than Travis but I'm not able to build it with either the instructions in the [wiki](https://github.com/hydrogen-music/hydrogen/wiki/Building-Hydrogen-from-source-(MAC-OSX)) or in #657 . But with MacPorts I'm able to do it without any problems.

To be honest, I do not want to look into how to fix the `brew` building process. Learning about proprietary operation systems just feels like a waste of time and there are a number of more important issue to be fixed.

But I will put the content of the issue and some further notes in the wiki so we can close it.

mauser:
Ok, i'm fine with the wiki solution. In my opinion sometimes brew works better, and sometimes MacPorts. Both systems can sometimes be found in a state where libraries don't fit to each other.. The some goes for mxe. But maybe this makes it clear why i want to keep both instructions, if possible :)

Issue #751, opened 2019-08-11
Segmentation fault while saving drumkit


theGreatWhiteShark:
Hydrogen version *          : 6eb975e
Operating system + version  : Devuan ASCII
Audio driver + version      : jackd 1.9.12

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Not quite sure since when but when selecting _Drumkits > Save_ on current master, H2 will just crash with a segmentation fault. It's a little bit irregular and usually happens during the next action. So, it's best to observe when saving while playback is rolling.

I was able to boil it down to four different lines of code (uncommented everything (apart from the actual saving) works fine).

- //src/core/src/basics/drumkit.cpp *Drumkit::save* in line 218 `QFileInfo fi( image );`
  If the input string `image` is not set, because the image field of the current drumkit is empty, this action results in an error

- //src/core/src/basics/drumkit.cpp *Drumkit::save* in line 230 `pDrumkit->set_components(pCopiedVector); delete pDrumkit;`
  Seems like there was not a deep copy made from the current drumkit components since`sampler::__render_note` asserts on the drumkit component the next time it will be called.

- //src/core/src/basics/drumkit.cpp *Drumkit::save* in line 254 `save_samples( dk_dir, overwrite );`
- //src/core/src/basics/drumkit.cpp *Drumkit::save* in line 254 `save_file( Filesystem::drumkit_file( dk_dir ), overwrite );`
 Not quite sure yet what is happening in here

theGreatWhiteShark:
For the record: it's actually  the copying of the instruments causing the segmentation fault in the second point

Below you can see the address of all instruments, their instrument components, and the corresponding `__related_drumkit_componentID` field including its value. The latter will be used in `sampler::__render_note()`
`pMainCompo = pEngine->getSong()->get_component( pCompo->get_drumkit_componentID() );`
and causes the segmentation fault.

> [orig inst component] 0 : [inst]: 0x556f72c6de90 : [comp]: 0x556f63c3a960 : [id]: 0 : 0x556f63c3a968
> [orig inst component] 1 : [inst]: 0x556f72e26000 : [comp]: 0x556f63c3b5b0 : [id]: 0 : 0x556f63c3b5b8
> [orig inst component] 2 : [inst]: 0x556f72de71e0 : [comp]: 0x556f63c3d1e0 : [id]: 0 : 0x556f63c3d1e8
[...]

PR #752, opened 2019-08-11
Fixing copy constructors & check image path when saving drumkit

theGreatWhiteShark:
else the construction of a QFileInfo will cause trouble.

Fix of ~the trivial first part of~ #751

### DIFF
--- a/src/core/src/basics/drumkit.cpp
+++ b/src/core/src/basics/drumkit.cpp
@@ -70,7 +70,9 @@ Drumkit::Drumkit( Drumkit* other ) :
 	__instruments = new InstrumentList( other->get_instruments() );
 
 	__components = new std::vector<DrumkitComponent*> ();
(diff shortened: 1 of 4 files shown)


theGreatWhiteShark:
Now the drumkit can be saved without Hydrogen crashing.

But I seemed to introduce a memory leak. I will look at this during the next couple of days.

theGreatWhiteShark:
Turns out the memory leakage was not introduced by me tweaking the copy constructors but was already present in H2.

in `MainForm::action_instruments_saveLibrary()` all calls to `Drumkit::load()` create a new Drumkit on the heap via `Drumkit::load_from()` or `Legacy::load_drumkit()`. These objects are now cleaned up manually and there is no memory leakage anymore when saving the drumkit.

mauser:
Merged, thanks a lot!

Issue #755, opened 2019-08-18
[Feat Req] More 'pattern Size' options in pattern editor panel (16ths-meters)


oddtime:
This is a simple problem: how to set values like 'five sixteenths' for a pattern size?
I didn't find any open issue on it (nor I know any way of doing it, without writing directly in song file)

I have just modified file src/gui/src/PatternEditor/PatternEditorPanel.cpp and built H2 again.
I have no experience in c++/Qt but it seems to work: now the new pattern size menu shows entries from 1 to 64 (instead of 1-32) and they are considered to be 16ths instead of 8ths.
Pros: can set 16ths-meter sizes!
        minimal code modification
Con: meters more difficult for the user to understand

What do you think about it? Would you suggest different ways?

theGreatWhiteShark:
Could you share the code by e.g. making a pull request?
https://github.com/hydrogen-music/hydrogen/wiki/How-to-contribute-code-using-Github

oddtime:
It is really an elementary alternative to the code. If you wish I can see how to do the the pull req....
this will take a while

If I am not wrong these were the mods:
in file `src/gui/src/PatternEditor/PatternEditorPanel.cpp`
in line 618 "8" comes "16"
in line 802 "8" comes "16" (from eighths to sixteenths)
in line 115 "32" comes "64"
in line 863 "32" comes "64" (items of pattern size menu)

oddtime:
@theGreatWhiteShark are these changements enough to justify a pull req?

oddtime:
The previous changes were applied to v1.0.0.beta1 but are inconsistent with the newest repository.

Just made pull request #756 to show the code.

theGreatWhiteShark:
> are these changements enough to justify a pull req?

Well, I'm not part of Hydrogen developer team and not authorized to merge pull request myself, but, sure. As long as there is a problem present in the current implementation, it's worth to point it out and to provide a solution/alternative. Even if it is just a single line of code.

oddtime:
Ok!
Question, I am working on tuplets as well. Should I clone H2 repo for each thread? Or same clone can be used for different pull requests?

theGreatWhiteShark:
With `git` it's best to keep those threads inside individual branches.

This tutorial describes some best practices to be used in bigger projects (like this one) https://nvie.com/posts/a-successful-git-branching-model/

oddtime:
Pull request #999 closes this issue

PR #756, opened 2019-08-19
16ths-meters for pattern size menu

oddtime:
For simplicity I am opening a pull request related to issue #755

Basically here the pattern size menu shows integers from 1 to 64 (instead of 1-32) and they are considered to be 16ths instead of 8ths.
It allows to set meters like 5/16.

Variable `nEighth` should be renamed `nSixteenth` but I have not done it yet

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditorPanel.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorPanel.cpp
@@ -112,7 +112,7 @@ PatternEditorPanel::PatternEditorPanel( QWidget *pParent )
 	__pattern_size_combo = new LCDCombo(pSizeResol, 4);
 	__pattern_size_combo->move( 34, 2 );
 	__pattern_size_combo->setToolTip( trUtf8("Select pattern size") );
-	for ( int i = 1; i <= 32; i++) {
+	for ( int i = 1; i <= 64; i++) {
 		__pattern_size_combo->addItem( QString( "%1" ).arg( i ) );
 	}
 	// is triggered from inside selectedPatternChangedEvent()
@@ -615,7 +615,7 @@ void PatternEditorPanel::selectedPatternChangedEvent()
 
 		// update pattern size combobox
 		int nPatternSize = m_pPattern->get_length();
-		int nEighth = MAX_NOTES / 8;
+		int nEighth = MAX_NOTES / 16;
 		__pattern_size_combo->select( (nPatternSize / nEighth) - 1 );
 	}
 	else {
@@ -799,7 +799,7 @@ void PatternEditorPanel::patternSizeChanged( int nSelected )
 		return;
 	}
 
-	int nEighth = MAX_NOTES / 8;
+	int nEighth = MAX_NOTES / 16;
 
 	if ( !m_bEnablePatternResize ) {
 		__pattern_size_combo->select( ((m_pPattern->get_length() / nEighth) - 1), false );
--- a/src/gui/src/PatternEditor/PatternEditorPanel.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorPanel.cpp
@@ -615,8 +615,8 @@ void PatternEditorPanel::selectedPatternChangedEvent()
 
 		// update pattern size combobox
 		int nPatternSize = m_pPattern->get_length();
-		int nEighth = MAX_NOTES / 16;
-		__pattern_size_combo->select( (nPatternSize / nEighth) - 1 );
+		int n16th = MAX_NOTES / 16;
+		__pattern_size_combo->select( (nPatternSize / n16th) - 1 );
 	}
 	else {
 		m_pPattern = NULL;
@@ -799,15 +799,15 @@ void PatternEditorPanel::patternSizeChanged( int nSelected )
 		return;
 	}
 
-	int nEighth = MAX_NOTES / 16;
+	int n16th = MAX_NOTES / 16;
 
 	if ( !m_bEnablePatternResize ) {
-		__pattern_size_combo->select( ((m_pPattern->get_length() / nEighth) - 1), false );
+		__pattern_size_combo->select( ((m_pPattern->get_length() / n16th) - 1), false );
 		QMessageBox::information( this, "Hydrogen", trUtf8( "Is not possible to change the pattern size when playing." ) );
 		return;
 	}
 
-	m_pPattern->set_length( nEighth * ( nSelected + 1 ) );
+	m_pPattern->set_length( n16th * ( nSelected + 1 ) );
 
 	m_pPatternEditorRuler->updateEditor( true );	// redraw all
 	m_pNoteVelocityEditor->updateEditor();


theGreatWhiteShark:
Could you add some notes in https://github.com/hydrogen-music/hydrogen/blob/master/src/core/include/hydrogen/config.dox about `MAX_NOTES` telling that it must be a multiple of 16? Doesn't it? I'm not entirely sure when looking at `PatternEditorPanel::patternSizeChanged()` but it feel like a good idea to have it this way.

The length of a `Pattern` only seem to be set in `PatternEditorPanel::patternSizeChanged()` itself. This should be fine.

trebmuh:
> What do you think about setting enumerator and denominator independently

:+1:

oddtime:
> Could you add some notes in https://github.com/hydrogen-music/hydrogen/blob/master/src/core/include/hydrogen/config.dox about MAX_NOTES telling that it must be a multiple of 16? Doesn't it?

From what I understand MAX_NOTES, which is the number of ticks in a 4/4 bar, is set = 192 because it allows triplets and 64ths resolution... 192=64*3. 
https://github.com/hydrogen-music/hydrogen/blob/master/src/core/include/hydrogen/config.dox states that it's set by cmake... Are you suggesting something like the following (except capital)?
```
/**
 * Maximum number of notes. Set by __cmake__ to 192 (currently).
*IT HAS TO BE MULTIPLE OF 64 AND 3
 */
#define MAX_NOTES       	@MAX_NOTES@
```

> What do you think about setting enumerator and denominator independently.

Yeah I thought the same, but we need an additional widget and I don't know how to make it cause I have never used Qt/c++. So if someone wants to do it I am happy, I would do it now if I could.
Setting 16 as fixed denominator was easier :)

oddtime:
Maybe we could have 5 differents widgets in Pattern panel:
size_numerator
size_denominator {2,4,8,16,32,64}
resolution, {4,8,16,32,64}
tuplet_numerator 
tuplet_denominator

By the way, the current Maximum resolution = 48 ticks/quarter = 192ticks/(4quarters) allows even 64T and 128T which are not present in the RES. menu!

theGreatWhiteShark:
> From what I understand MAX_NOTES, which is the number of ticks in a 4/4 bar, is set = 192 because it allows triplets and 64ths resolution... 192=64*3.
> https://github.com/hydrogen-music/hydrogen/blob/master/src/core/include/hydrogen/config.dox states that it's set by cmake... Are you suggesting something like the following (except capital)?

So, this is a variable any user can set using the `cmake` command during compilation or by manually editing the [CMakeFile](https://github.com/hydrogen-music/hydrogen/blob/master/CMakeLists.txt#L161).

Good point with the triplets! I think it's best to write something like "This number was set to 196 since it's a multiple of 64 * 3 (the highest resolution possible; mind the triplets). Only change the value with extreme care since it could result in unexpected behavior of the application." People should have the freedom to change this kind of parameters as they like. But we need to inform them that this might be a very bad idea.

theGreatWhiteShark:
> I don't know how to make it cause I have never used Qt/c++

Don't worry. Fortunately, it's not that difficult (especially the Qt and GUI part). It's just one year since I was at the very same spot as you are in right now. ;) 

> Maybe we could have 5 differents widgets in Pattern panel

I think we have to be clever in terms of UI/UX on this one. After all, we want to improve things and just changing the resolution of a Pattern might be to cumbersome with the enumerator/denominator approach. There are a number of pros and cons that need to be considered.

But this is definitely out of the scope of this pull request and we should shift the discussion to the developer mailing list instead. 

One short hint: when something is worth to be bind to a keyboard shortcut, it's also worth to being represented by a widget. Changing the resolution and pattern length IMHO definitely are.

oddtime:
Ok, if the independent enumerator/denominator (widgets or not) is generally approved by the dev team I can try to implement it but I don't guarantee it soon.
In the meanwhile let's see if the feature is worth the job.

theGreatWhiteShark:
First of all, I'm just a random contributor and not part of the dev team. So, I don't have a say in what's being merged and what's not.

Secondly, I'm not sure myself if an independent enumerator/denominator would really improve things or if there might be an even better way to restructure the panel. I think we it would be a good idea to discuss the pros and cons on the mailing list

oddtime:
Ok! I shall ask soon on the mail list
cheers

oddtime:
Some progress:
Added a denominator widget
background_res-new2

here an example that shows how it works
https://youtu.be/_bpkhZsO0jU

Explanation (if the video is not clear).
Denominator act as a (inverse) unit for the pattern size display (you can consider this as the numerator).
When changing the denominator, numerator display is automatically updated, while the pattern size is unchanged.
When changing the numerator widget, pattern size is changed.

Problems:

1. numerator display is constrained to integer numbers, sometimes that causes inaccurate indication (like a 9/8 bar that should be shown as 4.5/4). This problem is actually already present in hydrogen and occurs if one puts a pattern size in the .h2song (with text editor) that is not integer multiple of the duration of a eight note (in ticks).

2. I make an example for describing it: consider a give pattern with size of a 9/8 bar. Now select "4" for the denominator. Numerator widget is automatically updated and shows 4 (instead of 4.5 or a warning message). Now if I want to set a 4/4 size using the numerator widget it won't update the pattern size, unless you select a temporary different value (eg. 3) and then select again 4.

I thought of these possible solutions:

1. set warning text in numerator display.

2. change the pattern size numerator widget to show any value (decimal too).

oddtime:
Solution 2 (widget deactivated but shows a value instead of "err")

screenshot-0b644973

oddtime:
Any preferences?

cme:
I have a couple of questions, I hope that's okay. I wasn't able to work out how it actually worked from watching the video, and had to read the description to understand what was going on.

1. Why doesn't the pattern size update immediately on changing the denominator? It's better to show the user the result of their actions immediately. That way they can work out how things actually work by exploring, and don't get nasty surprises later on when they change the denominator, nothing happens, they forget what they were doing and then later change the numerator.

2. Is this the right UI for this? (Sorry, this is a big question)
Having a menu for this has always irritated me slightly, as a textual entry box with spin buttons or similar seemed like a much better idea. Only those don't currently exist in the Hyrdogen LCD style widget set. :/ A text entry box style widget or spin box might be better served by decimals or proper fractions. Alternatively, a dialog box with spin boxes for both?

3. Does this get saved? The songfile currently only has pattern length in ticks. So it would be possible for the user to set their pattern size in one denomination, then get confused when they next reload to see that their pattern is shown as a different length afterwards.

oddtime:
Hi! Colin, thanks for your feedback. Answers below in blue 

I have a couple of questions, I hope that's okay. I wasn't able to work out how it actually worked from watching the video, and had to read the description to understand what was going onSorry, did you see the description at https://github.com/hydrogen-music/hydrogen/pull/756 ?
Why doesn't the pattern size update immediately on changing the denominator? It's better to show the user the result of their actions immediately. That way they can work out how things actually work by exploring, and don't get nasty surprises later on when they change the denominator, nothing happens, they forget what they were doing and then later change the numerator.
Actually it does update immediately. Maybe there was a problem in video encoding
Is this the right UI for this? (Sorry, this is a big question). Having a menu for this has always irritated me slightly, as a textual entry box with spin buttons or similar seemed like a much better idea. Only those don't currently exist in the Hyrdogen LCD style widget set. :/ A text entry box style widget or spin box might be better served by decimals or proper fractions. Alternatively, a dialog box with spin boxes for both?
This can be discussed, I had no experience with widgets so I didn't modified it. However no problem with the question, I agree.
Does this get saved? The songfile currently only has pattern length in ticks. So it would be possible for the user to set their pattern size in one denomination, then get confused when they next reload to see that their pattern is shown as a different length afterwards.No, with this way you continue to set the pattern size in ticks. There is not a new variable saved, just a new denominator widget allowing new meters.If someone wants to help adding the "denominator" property to each pattern I am fine... I am not able to do it right now.But still it is something new that is useful I think, even without.If you don't want get confused you can just leave denominator "8" and continuing using hydrogen as before...By the way we could discuss even about the way the grid is displayed, I mean numbers appearing in the pattern editor  (e.g 6/8 is shown as 3/4, or a 12/8 is even more confusing); patterns could store this information as well... Can we think to this feature later?

Another thing I thought,:there exists some uncommon meters used in contemporary classical scores like 4/3 or 11/12 (i.e. Pierre Boulez, but also in some modern jazz even if jazz musicians don't know this notation in most cases and they uses bpm variation, which is much more difficult to deal with)Example: a bar in 11/12 is a 4/4 bars (subdivided in triplets) minus one 8th-note triplet. It would be fantastic to add this feature as well, in my opinion!We can do it with multiples of 3 because the actual MAXRESOLUTION is capable to handle triplets precisely.No with other prime numbers, but this is another problem and we could start with multiples of 3 only.

oddtime:
Sorry I misunderstood point 1
The pattern size numerator widget does change but the not the pattern size. You would need to store the denominator property in the pattern if you want to really update the size after denominator widget change.


 

    Il sabato 14 novembre 2020, 15:39:40 CET, Colin McEwan <notifications@github.com> ha scritto:  
 
 


I have a couple of questions, I hope that's okay. I wasn't able to work out how it actually worked from watching the video, and had to read the description to understand what was going on.
   
   -    
Why doesn't the pattern size update immediately on changing the denominator? It's better to show the user the result of their actions immediately. That way they can work out how things actually work by exploring, and don't get nasty surprises later on when they change the denominator, nothing happens, they forget what they were doing and then later change the numerator.

   -    
Is this the right UI for this? (Sorry, this is a big question)   
Having a menu for this has always irritated me slightly, as a textual entry box with spin buttons or similar seemed like a much better idea. Only those don't currently exist in the Hyrdogen LCD style widget set. :/ A text entry box style widget or spin box might be better served by decimals or proper fractions. Alternatively, a dialog box with spin boxes for both?

   -    
Does this get saved? The songfile currently only has pattern length in ticks. So it would be possible for the user to set their pattern size in one denomination, then get confused when they next reload to see that their pattern is shown as a different length afterwards.

oddtime:
Another thing I thought: there exist some uncommon meters used in contemporary classical scores like 4/3 or 11/12 (for example Pierre Boulez, but also in some modern jazz even if jazz musicians don't know this notation in most cases and they uses bpm variation, which is much more difficult to deal with)
Example: a bar in 11/12 is a 4/4 bars (subdivided in triplets) minus one 8th-note triplet. It would be fantastic to add this feature as well, in my opinion!
We can do it with multiples of 3 because the actual MAXRESOLUTION is capable to handle triplets precisely.No with other prime numbers, but this is another problem and we could start with multiples of 3 only.

oddtime:
Hi! @cme thanks for your feedback.

>Why doesn't the pattern size update immediately on changing the denominator? It's better to show the user the result of their actions immediately. That way they can work out how things actually work by exploring, and don't get nasty surprises later on when they change the denominator, nothing happens, they forget what they were doing and then later change the numerator.

The pattern size numerator widget does change but the not the pattern size. You would need to store the denominator property in the pattern if you want to really update the size after denominator widget change.

> Is this the right UI for this? (Sorry, this is a big question)
Having a menu for this has always irritated me slightly, as a textual entry box with spin buttons or similar seemed like a much better idea. Only those don't currently exist in the Hyrdogen LCD style widget set. :/ A text entry box style widget or spin box might be better served by decimals or proper fractions. Alternatively, a dialog box with spin boxes for both?

This can be discussed, I had no experience with widgets so I didn't modified it. However no problem with the question, I agree.

> Does this get saved? The songfile currently only has pattern length in ticks. So it would be possible for the user to set their pattern size in one denomination, then get confused when they next reload to see that their pattern is shown as a different length afterwards.

No, with this way you continue to set the pattern size in ticks. There is not a new information saved in h2song file, just a new denominator widget allowing new meters. If someone wants to help adding the "denominator" property to each pattern I am fine... I am not able to do it right now.
But still it is something new that is useful I think, even without.
If you don't want get confused you can just leave denominator "8" and continuing using hydrogen as before...
By the way we could discuss even about the way the grid is displayed, I mean numbers appearing in the pattern editor  (e.g 6/8 is shown as 3/4, or a 12/8 is even more confusing); patterns could store this information as well...
Can we think to this feature later?

oddtime:
What about this idea for the gui?
screenshot-dab08713

oddtime:
screenshot-1173c8c7

News: pattern_size_combo widget substituted with this LCDwidget, editable with text entry.
Type is double so it is possible to enter odd 16th patterns.

I have added the property `denominator` to the pattern class, Default set = 4, which is actually read from the Size widget, but not saved into songfile, nor imported from file.
With that we may use a custom denominator without confusion.

Also, I wasn't able to create a dialog with multiple inputs, for num/den.

cme:
This seems pretty good to me! I was going to suggest allowing decimals to set the size.

Also, I *really* like anchoring it as a proportion of quarter-notes which I think works really well with the fixed (for the moment) 48 TPQN

oddtime:
Hello, please may I ask your help:?
I'm trying to publish the code but I don't want to make a mess with git / github.

This new branch includes the new pattern size widget:  https://github.com/oddtime/hydrogen/tree/customPatternSize

This current pull request is based on my github master branch (probably was not the best way...).

Do I have to create another pull req or can I rebase the current one to the new branch...?

cme:
I don't think there's any way to change the source branch on a pull request (the target branch can be changed easily enough though). I think creating a new pull request is fine -- just reference this one in the message for the new PR, and close this one. This will still be visible here for reference, and we can still comment on this even after it's closed.

oddtime:
Ok, I am closing the current Pull Request and opening #999

PR #759, opened 2019-08-21
Tuplets

oddtime:
Similar to Pull Request #364, applied to v1.0.0beta.

I am new to github, C++, Qt... sorry for any errors, it was a hard job. I hope to help to develop the tuplet feature which is important for many users. 
Thank to all the people behind such a fantastic drum machine.
I am happy to receive feedback, opinions and suggestions.

**Changes made** (compared to #364):
- correct grid visualization
- position of tuplet notes is explicitly rounded, for best approximation (since implicit int cast rounds always down). In my opinion that makes the inexact tuplets a bit more pleasant to hear, at least for quintuplets.
- 'Resolution' is scaled according to the notation rule for the tuplet number (See [https://en.wikipedia.org/wiki/Tuplet]() )
- fixed fillNotes function
- Update velocity randomize

**Still to solve** (help is welcome)
 - velocity randomize function: I can't figure out why the same changes that work for the fillnotes function, are not working for velocity randomize function. So I added a line that does the job but you cannot undo the action.
In file  `src/gui/src/PatternEditor/PatternEditorInstrumentList.cpp` line 457:
`pNote->set_velocity(fVal);`

**Idea for perfect tuplets**
The problem of discrepancy between the ideal exact (non integer) and the rounded (integer) position of notes - due to finite resolution = 48 ticks/quarter which is not multiple of 5,7,9 - may be solved via the leadlag note property...
It would be simpler for me if some H2 expert told me
-how lead lag value acts,
-where in the code,
-how to mod the code so that a wanted leadlag is set together with a new note.

Note: the first two commits are present in my previous pull request too (about pattern size menu, very simple)...

### DIFF
--- a/src/core/include/hydrogen/Preferences.h
+++ b/src/core/include/hydrogen/Preferences.h
@@ -453,6 +453,16 @@ public:
 	bool			isPatternEditorUsingTriplets();
 	void			setPatternEditorUsingTriplets( bool value );
 
+	bool			isPatternEditorUsingQuintuplets();
+	void			setPatternEditorUsingQuintuplets( bool value );
+
+	bool			isPatternEditorUsingSeptuplets();
+	void			setPatternEditorUsingSeptuplets( bool value );
+
+	bool			isPatternEditorUsing9tuplets();
+	void			setPatternEditorUsing9tuplets( bool value );
+
+
 	bool			isFXTabVisible();
 	void			setFXTabVisible( bool value );
 	
@@ -672,6 +682,9 @@ private:
 	float					mixerFalloffSpeed;
 	int						m_nPatternEditorGridResolution;
 	bool					m_bPatternEditorUsingTriplets;
+	bool					m_bPatternEditorUsingQuintuplets;
+	bool					m_bPatternEditorUsingSeptuplets;
+	bool					m_bPatternEditorUsing9tuplets;
 	bool					m_bShowInstrumentPeaks;
 	bool					m_bIsFXTabVisible;
 	bool					m_bShowAutomationArea;
@@ -945,12 +958,37 @@ inline void Preferences::setPatternEditorGridResolution( int value ) {
 	m_nPatternEditorGridResolution = value;
 }
 
+inline bool Preferences::isPatternEditorUsingQuintuplets() {
+	return m_bPatternEditorUsingQuintuplets;
+}
+
+inline bool Preferences::isPatternEditorUsingSeptuplets() {
+	return m_bPatternEditorUsingSeptuplets;
+}
+
+inline bool Preferences::isPatternEditorUsing9tuplets() {
+	return m_bPatternEditorUsing9tuplets;
+}
+
+
 inline bool Preferences::isPatternEditorUsingTriplets() {
 	return m_bPatternEditorUsingTriplets;
 }
 inline void Preferences::setPatternEditorUsingTriplets( bool value ) {
 	m_bPatternEditorUsingTriplets = value;
 }
+inline void Preferences::setPatternEditorUsingQuintuplets( bool value ) {
+  m_bPatternEditorUsingQuintuplets = value;
+}
+
+inline void Preferences::setPatternEditorUsingSeptuplets( bool value ) {
+  m_bPatternEditorUsingSeptuplets = value;
+}
+
+inline void Preferences::setPatternEditorUsing9tuplets( bool value ) {
+  m_bPatternEditorUsing9tuplets = value;
+}
+
 
 inline bool Preferences::isFXTabVisible() {
 	return m_bIsFXTabVisible;
--- a/src/core/src/hydrogen.cpp
+++ b/src/core/src/hydrogen.cpp
@@ -2549,7 +2549,18 @@ void Hydrogen::addRealtimeNote(	int		instrument,
 	Preferences *pPreferences = Preferences::get_instance();
 	unsigned int nRealColumn = 0;
 	unsigned res = pPreferences->getPatternEditorGridResolution();
-	int nBase = pPreferences->isPatternEditorUsingTriplets() ? 3 : 4;
+	//int nBase = pPreferences->isPatternEditorUsingTriplets() ? 3 : 4;
+	int nBase;
+	if(pPreferences->isPatternEditorUsingTriplets())
+		nBase = 3;
+	else if(pPreferences->isPatternEditorUsingQuintuplets())
+		nBase = 5;
+	else if(pPreferences->isPatternEditorUsingSeptuplets())
+		nBase = 7;
+	else if(pPreferences->isPatternEditorUsing9tuplets())
+		nBase = 9;
+	else
+		nBase = 4;
 	int scalar = ( 4 * MAX_NOTES ) / ( res * nBase );
 	bool hearnote = forcePlay;
 	int currentPatternNumber;
--- a/src/gui/src/PatternEditor/DrumPatternEditor.cpp
+++ b/src/gui/src/PatternEditor/DrumPatternEditor.cpp
@@ -58,6 +58,9 @@ DrumPatternEditor::DrumPatternEditor(QWidget* parent, PatternEditorPanel *panel)
  , Object( __class_name )
  , m_nResolution( 8 )
  , m_bUseTriplets( false )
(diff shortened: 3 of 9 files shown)


theGreatWhiteShark:
To be honest, I don't like the graphical approach to tuplets because it neither gives true tuplets nor let you choose an arbitrary subdivision. Problem with the `leadlag` is, that is lives on the same tick-based grid as the notes drawn by hand (in *hydrogen.cpp* > `audioEngine_updateNoteQueue()`). The resulting precision would be not enough to allow a satisfying result. But I gave it some thoughts and I think one could rather easily implement true tuplets in Hydrogen.

# Proposal for true tuplets

Tuplets, as far as I understand the concept, are equally-sized integer subdivisions of a single note. E.g. a quarter can be split into a triplet, which take up exactly the space/time the quarter would have. Also, whenever I encountered it in `Tuxguitar` I used it as a decorator for a couple of notes and not necessary for the whole beat. 

Inspired by this handling and concepts of fractional calculus, I would add a property called e.g. *subdivisions = N* to each `Note` and a an operator on both core and gui side, which is able to handle this property.

Pros: It would require minimal changes in the core to implement it.
Cons: Songs with this property set would not be properly played in older versions.

## Core operator

I would leave everything until the very end of the processing of each note untouched. When it's about to add the bits and bytes of the buffer representing the note's sample of the instrument to the main buffer, I would take this buffer, extract the portion corresponding to the exact length of the note with respect to the current tempo, and extract the first 1/N (number of subdivisions) bits. Next, I would glue N-1 of them together and add the whole sample in the last remaining slot. This should give true triplets on the resolution of the samplesize.

One could of course also introduce randomization in time by dividing lead-lag noise level by N and simple adjusting the number of bits copied from the original buffer. 

Randomizing velocity and pitch is not that straight forward since we can not just multiply the buffer by a factor. This would disregard all nonlinearities involved in e.g. an LADSPA effect being applied on the sample. Instead, one needs to make multiple transformations of the original sample, which might be expensive.

Having a couple of vectors of length N holding the customization of the subdivisions of that single note, the user would have full control over whats happening (velocity, probability, pitch, leadlag, whether being played).

Pros: 
- True tuplets with temporal resolution of the samplerate
- Lead-lag behavior among them at the resolution of the samplerate (it's probably worth to implement the lead-lag using zero padding on a buffer level instead of constraining it to the tick-grid for all of Hydrogen. Well, someday...)
- (Arbitrary) number of subdivision would be possible. So, we are not restricted to triplets, quintuplets etc.

Cons:
- Having 103 subdivisions and pitch/velocity randomization might be too much to handle and cause Hydrogen to produce XRUNs.

## GUI operator

This one simply checks whether a note has a subdivision property > 1 and if so, retrieves the whole width the original note is supposed to allocate in both the pattern editor and property panel and fills it with those note circles/bars at all 1/N positions at a resolution of pixels. Maybe even with a different color (or the numbers 3, 5, 7, 9, N) indicating that the tuplet property is set. One would also need a new button/shortcut to activcate/adjust this feature.

romsom:
Hi,

Thanks @oddtime for your efforts!
Are you still working on this?
I gave this a try as well and generalized it a bit more by replacing the "isUsingTriplets" property by an integer divisionBase/subdivisions property, mainly as a task to familiarize myself a bit with the code base. As a result it is possible to add arbitrary subdivisions easily.
I'm still sorting out some problems with different rounding in different parts of the PatternEditor code, but things look promising.

I would love to see this in hydrogen, even playing around with my unfinished patched version was a lot of fun :)

On a sidenote: The concept of coarser granularity quantization seems to sound quite interesting and could be something worth implementing as well, but of course with arbitrary resolution (up to a certain point).
I have a few ideas and would like to implement them in hydrogen if it's possible.

Also thanks @theGreatWhiteShark for the insight.

I'm not sure if I understood your proposal entirely. Having notes only e.g., on the tuplet offbeats would still be possible, right?
I guess the note would need two properties then: subdivision n and position 0 <= p < n in the beat.
If we wanted to place more notes than one into the beat, all those notes would be positioned on the same tick?

Do you think it would also be possible to make the tick resolution dynamic as part of the song instead?
I suspect the new song format would be somewhat incompatible with the existing format.
Then we could have true tuplets on pattern level without manipulating the buffer and having to worry about the audio signal chain.

At least for my workflow having a grid with a certain resolution seems quite adequate. I think more in polyrhythms than in ornamenting certain notes, though.


One could determine the resolution needed from the least common multiple of the subdivisions used in the song. Maybe limit the value to a rather high value to avoid performance problems with songs that use a lot of prime subdivisions.

Just some thoughts.
Sorry if I missed something obvious. I'm new to the code base and still in the process of exploring it. :)

oddtime:
thanks @romson
looking forward to build your improved version. I didn't go ahead with the feature as the approximate graphical approach seemed to be not liked. @theGreatWhiteShark 's solution is interesting but I am not able to implement it currently and I should learn some c++

theGreatWhiteShark:
Hey

> Having notes only e.g., on the tuplet offbeats would still be possible, right?
I guess the note would need two properties then: subdivision n and position 0 <= p < n in the beat.
If we wanted to place more notes than one into the beat, all those notes would be positioned on the same tick?

Sure. Completely missed that one (maybe because I wouldn't be a user of this feature myself).

> Do you think it would also be possible to make the tick resolution dynamic as part of the song instead?

AFAIK this is something that has to be set during compile time. I'm not quite sure but maybe one can even set the tick resolution as a command line option during compilation. Problem with making this one dynamic would be that not just the resolution but also position(time) and speed would alter as well. It would require quite an amount of bookkeeping to keep all these things straight.

> At least for my workflow having a grid with a certain resolution seems quite adequate.

It's also dependent on tempo. For fast playback you may not hear any irregularities but that might change as speed decreases. Also, if you combine Hydrogen with other programs that support tuplets at samplerate resolution, the approximate solution would most probabily cause audible mismatches.

oddtime:
> But I gave it some thoughts and I think one could rather easily implement true tuplets in Hydrogen...

@theGreatWhiteShark  I still do not follow this implementation.
Tuplets are a way to write any rational note value, like 1/3 of whole note which has not its own dedicated musical symbol, but it is written as a 1/2 note with a "3" (or "3:2") over it that tells: multiply the note value times 2/3.
Would your proposal be able to represent this for example? Just asking because maybe I don't understand it
screenshot-3d74e86e

oddtime:
> Problem with the leadlag is, that is lives on the same tick-based grid as the notes drawn by hand

Are you sure that it doesn't live in a frame-based grid with sample rate resolution? As the effect of humanized timing?

in `Hydrogen.cpp`
```
if ( pSong->getHumanizeTimeValue() != 0 ) {
	nOffset += ( int )( getGaussian( 0.3 ) * pSong->getHumanizeTimeValue() * pHydrogen->m_nMaxTimeHumanize );
}

// Lead or Lag - timing parameter // Add a constant offset to all notes.
nOffset += (int) ( pNote->get_lead_lag() * nLeadLagFactor );
// [...]

Note *pCopiedNote = new Note( pNote );
pCopiedNote->set_position( tick );
pCopiedNote->set_humanize_delay( nOffset );
```

then in `Sample.cpp`
```
int noteStartInFrames = ( int ) ( pNote->get_position() * pAudioOutput->m_transport.m_fTickSize ) + pNote->get_humanize_delay();
```
If so an additional note member, say `m_time_offset` (like leadlag but automatically computed and fixed) could easily resolve the tuplet issue

oddtime:
@romsom can you show your version on github so to work on it?

romsom:
@oddtime Sure, here it is: https://github.com/romsom/hydrogen/tree/tuplet-generalization
Be warned though: It's not much more than a few experiments, but it should work (kind of) if I recall correctly. :)

oddtime:
I am having a look.
There were some changes in H2 master until now, so I try to merge manually all the new ideas in a new branch, which we can develop together of course.

oddtime:
A short video :)
Leadlag seems to work!
Here's a perfect sounding quintuplet (res combobox shows 1/4T but it is 1/4Q really)
https://user-images.githubusercontent.com/54219881/105991140-bbb3f180-60a3-11eb-8fb8-5390eb631d6f.mp4
That's just a rough experiment but looks promising!!

oddtime:
Closing and opening #1127

PR #761, opened 2019-08-25
fix setting of a new song

### LINKED ISSUE #760: LADSPA effects not activate when loading a song
Hydrogen version *          : 6eb975e
Operating system + version  : Devuan ASCII
Audio driver + version      : jackd 1.9.12

I add some LADSPA FX to a Song (the settings of how much FX added to each instrument is stored both in the *.h2song* and *drumkit.xml* file but the name/kind of the LADSPA FX is only stored in the *.h2song* file) and everything works fine. But whenever I restart Hydrogen or load the Song none of the FX is active. I have to add another one or at least open up the list of different FX to make the others work again.

As far as I can see right now the problem is that `audioEngine_setupLadspaFX()`, which takes care of activating the FX, is called from `audioEngine_setSong()` before the Song is actually loaded.

theGreatWhiteShark:
`__song` has to be set prior to the call to `audioEngine_setSong()` in order for the functions it calls to work properly.

Fixes #760


mauser:
Thanks a lot, merged!

Issue #736, opened 2019-05-28
Horizontal scrolling


theGreatWhiteShark:
Hydrogen version *          :  3b44914 
Operating system + version  :  Devuan ASCII + i3

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Is it just me and my window manager or does the horizontal scrolling of the pattern editor does not work?

Using the touchpad the content of the pattern editor moves slightly but always snaps back. The number on top and the scrollbar do not move at all

thijz:
just tested on latest code and everything seems to work as expected :

- grabbing and dragging the scrollbar
- 2-finger scroll om trackpad (both on scrolbar and on the pattern editor area) 
- clicking the left/right arrows next to the scrollbar
- works at any zoom level
- works in song/pattern mode looped and non looped

theGreatWhiteShark:
Hmmm. That's odd. Maybe it's my tiling window manager messing things up in Qt.

esbatl:
You mean like [this](https://youtu.be/MYqlL75TyjQ)? Bc if so, SAME :(

cme:
Yeah, I've always observed this behaviour on Hydrogen, on Windows, Mac OS X and Linux.

> just tested on latest code and everything seems to work as expected :
>     * grabbing and dragging the scrollbar
> [...]
>     * 2-finger scroll om trackpad (both on scrolbar and on the pattern editor area)

Interesting -- did you click in the scroll bar and then immediately use 2-finger scroll without clicking the pattern editor in between? That should work, but isn't very useful.

If you click the in the pattern editor, and *then* use 2-finger scroll to try and scroll horizontally, the behaviour I've observed on all Hydrogens to date is that the PaternEditorPanel will scroll but the scroll bar, velocity etc editors and ruler will *not* scroll.

I have a fix for this (tested OS X and Linux) and will get around to a pull request in the next couple of days.

mauser:
@theGreatWhiteShark : Could you re-test with the current master branch?

theGreatWhiteShark:
Hey, I'm AFK till next weekend. I'll test the patch soon.

theGreatWhiteShark:
Works splendidly. Thanks a lot!

PR #765, opened 2019-09-05
Fix scrolling

cme:
Fix pattern editor horizontal scrolling issues (bug #736). Scrolling (using cursor keys, touchpad gestures or wheel movements) in the Pattern Editor or any of its subcomponents would scroll that pane only, leaving it out of sync with the horizontal scrollbar and the rest of the Pattern Editor.


mauser:
Thanks a lot, merged!

PR #768, opened 2019-09-15
Fix copy constructor of `Sample`

### LINKED ISSUE #764: Sample playback and preview for multi layered instruments is broken
Hydrogen version *          : 3652dbf9ecea3aba2b2cd47a068eb47b95381953

When previewing sounds from a multi-layered kit (like _Tom 10" flam_ from the Flac_GSCW-1 kit) , the sound of the instrument seems to be messed up. The sample starts with the original sample (as one would expect), but after that ends, a different sample will be played. 

This is also shown in the sample wave display: 

grafik

When opening the sample editor, the correctly processed sample can be heard when using the "play original sample" functionality.

theGreatWhiteShark:
Since the underlying data is given in float and the number of floats instead of the corresponding number of bytes is handed to the `memcpy` function call, only 25% of the samples have been copied.

This fixes #764.

I'm a little puzzled how this could remain unnoticed.

### DIFF
--- a/src/core/src/basics/sample.cpp
+++ b/src/core/src/basics/sample.cpp
@@ -69,8 +69,13 @@ Sample::Sample( Sample* pOther ): Object( __class_name ),
 {
 	__data_l = new float[__frames];
 	__data_r = new float[__frames];
-	memcpy( __data_l, pOther->get_data_l(), __frames );
-	memcpy( __data_r, pOther->get_data_r(), __frames );
+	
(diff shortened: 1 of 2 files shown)


mauser:
Merged! 

Wow, that's a little bit creepy :smiley:  Maybe the copy constructor wasn't called earlier?

theGreatWhiteShark:
Seems like the broken copy constructor of `InstrumentLayer` did save us :laughing:

PR #773, opened 2019-09-28
OSC session management

theGreatWhiteShark:
I'm working on a proper reimplementation of the NSM (Non Session Management) since in its current state it is not working at all. But I thought instead of implementing all the required handlers in some NSM specific classes and introduce even more redundancy in the code, I created some more general OSC handlers taking care of all the basic functionality.

You can now use these five new messages to control Hydrogen via the command line
`oscsend localhost 19462 /Hydrogen/NEW_SONG s "$HOME/test.h2song"`
`oscsend localhost 19462 /Hydrogen/OPEN_SONG s "$HOME/test.h2song"`
`oscsend localhost 19462 /Hydrogen/SAVE_SONG`
`oscsend localhost 19462 /Hydrogen/SAVE_SONG_AS s "$HOME/test.h2song"`
`oscsend localhost 19462 /Hydrogen/QUIT_SONG`

This can become very handy in case we plan to revive the CLI.

## Notes about the implementation
I introduced two variable specifying whether Hydrogen is under session management and if the Qt5 GUI is active. If the latter is true, the new `Song` **must** be set via the GUI itself or it will freeze. To support this, another `Event` called `EVENT_UPDATE_SONG` was introduced. 

Shutting down the application is only handled by the GUI right now using the `EVENT_QUIT` `Event`. 

I also found a bug in the `Event` dispatcher `HydrogenApp::onEventQueueTimer()`. Since `HydrogenApp` was not itself added as an `EventListener` all its methods implementing the `EventListener` were called as many times as `EventListener`s enrolled.

### DIFF
--- a/src/core/include/hydrogen/nsm_client.h
+++ b/src/core/include/hydrogen/nsm_client.h
@@ -71,6 +71,13 @@ class NsmClient : public H2Core::Object
 
 		void shutdown();
 
+		/**
+		 * To determine whether Hydrogen is under Non session management,
+		 * it is not sufficient to check whether the NSM_URL environmental
+		 * variable is set but also whether. 
+		 */
+		bool m_bUnderSessionManagement;
+
 	private:
 		NsmClient();
 
--- a/src/core/include/hydrogen/event_queue.h
+++ b/src/core/include/hydrogen/event_queue.h
@@ -70,17 +70,33 @@ enum EventType {
 	 * decoupled from the creation and queuing of the corresponding
 	 * Note itself.
 	 *
+	 * In Director it triggers a change in the displayed number,
+	 * color, tag, and triggers Director::update(). In case the
+	 * provided value is 3, instead of performing the changes above,
+	 * the Director loads the metadata a the current Song.
+	 *
 	 * The associated values do correspond to the following actions:
 	 * - 0: Beat at the beginning of a Pattern in
 	 *      audioEngine_updateNoteQueue(). The corresponding Note will
 	 *      be created with a pitch of 3 and velocity of 1.0.
+	 *      Sets MetronomeWidget::m_state to
+	 *      MetronomeWidget::METRO_ON and triggers
+	 *      MetronomeWidget::updateWidget().
 	 * - 1: Beat in the remainder of a Pattern in
 	 *      audioEngine_updateNoteQueue(). The corresponding Note will
 	 *      be created with a pitch of 0 and velocity of 0.8. In
 	 *      addition, it will be also pushed by
 	 *      Hydrogen::setPatternPos() without creating a Note.
-	 * - 2:
-	 * - 3:
+	 *      Sets MetronomeWidget::m_state to
+	 *      MetronomeWidget::METRO_FIRST and triggers
+	 *      MetronomeWidget::updateWidget().
+	 * - 2: Signals MetronomeWidget to neither update nor setting
+	 *      MetronomeWidget::m_state.
+	 * - 3: Tells the Director that a new Song was loaded and triggers
+	 *      its Director::update().
+	 *      Sets MetronomeWidget::m_state to
+	 *      MetronomeWidget::METRO_ON and triggers
+	 *      MetronomeWidget::updateWidget().
 	 *
 	 * Handled by EventListener::metronomeEvent().
 	 */
@@ -91,7 +107,12 @@ enum EventType {
 	EVENT_PLAYLIST_LOADSONG,
 	EVENT_UNDO_REDO,
 	EVENT_SONG_MODIFIED,
-	EVENT_TEMPO_CHANGED
+	EVENT_TEMPO_CHANGED,
+	/**
+	 * Event triggered whenever the Song was changed outside of the
+	 * GUI, e.g. by session management or and OSC command.
+	 */
+	EVENT_UPDATE_SONG
 };
 
 /** Basic building block for the communication between the core of
--- a/src/core/include/hydrogen/event_queue.h
+++ b/src/core/include/hydrogen/event_queue.h
@@ -109,13 +109,21 @@ enum EventType {
 	EVENT_SONG_MODIFIED,
 	EVENT_TEMPO_CHANGED,
 	/**
-	 * Event triggered whenever the Song was changed outside of the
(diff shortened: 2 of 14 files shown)


trebmuh:
Nice to see !

Pinging @houston4444 as he might have some inside/comment about it.

trebmuh:
See #675 too

Houston4444:
Interesting !
I see that you implemented a SAVE_SONG_AS message, does it works under NSM ? if yes, does it switch on new saved song or not ? (it's forbidden by NSM, because it's very confusing user).

Have you the possibility to add ":dirty:" and ":switch:" NSM capabilities with this code ?
Thanks

theGreatWhiteShark:
Hey @Houston4444,

This is just a prerequesite I will use to implement the NSM support during the next weeks and I make it as general as possible so it can also be used in another context.

Jupp, the user won't be allowed to call "save as" via the GUI. Another thing that's still missing (afaik) is a function to easily restarting the JACK driver when switching the between sessions. So yes, I plan to add support for the `:dirty:` and `:switch:` capabilities too :)

Houston4444:
Really cool ! Thanks

mauser:
Hi! 

i've started to look into this pull request.. some remarks:

- You're using std::cout instead of INFOLOG (ERRORLOG, etc...). Is there are reason for this or can this be replaced with the corresponding *LOG macros? 

- Currently, "quit" is not implemented for the CLI client, but i think it might be very useful for it, especially for configuration which are using Hydrogen more or less as a "player" for h2songs only. Would look into this? Or is there a technical reason to not implement it currently?

- save_as is currently implemented in MidiActionManager, but i think there is no chance that this action can be triggered in any sensible way via midi (since you can't supply a filename). Wouldn't it be better suited in the core_action_controller class? I wrote that class when implementing common functionality which is used in the OSC server as well as in the UI. The initial idea was to avoid code duplication, but it could be also used for the implementation of actions which are not receivable via midi.

theGreatWhiteShark:
> You're using std::cout instead of INFOLOG (ERRORLOG, etc...). Is there are reason for this or can this be replaced with the corresponding *LOG macros?

Yeah, there was a reason. I had it all covered with `ERRORLOG` macros first but changed it later on. Unfortunately, I didn't wrote the reason down into a commit message. But I think it was due to development issues, like the NSM client asked to open for a song which doesn't exist and H2 failed with a segfault. Due to the time delay of the logger the corresponding messages would have been lost. Once H2 is set up `ERRORLOG` should be sufficient. And since it's send to std::cout instead of std::err, it smells even more like a debugging thingy.

> Currently, "quit" is not implemented for the CLI client, but i think it might be very useful for it, especially for configuration which are using Hydrogen more or less as a "player" for h2songs only. Would look into this? Or is there a technical reason to not implement it currently?

It's more a matter of strategy. Currently the main entry point is the `hydrogen` binary, which always comes with the Qt5 GUI. The `h2cli` and `h2player` do exist but are neither advertised nor mentioned in the documentation at all. In fact I only know about their presence because I had a look at the source code and I constantly forget that they are there. Using the OSC session management one could (at least on a very basic level) with H2 like via the GUI.

But I'm not sure about the plans for the CLI.  Will it at some point become a full fledged core you can control via OSC? How it be implemented? Maybe a `hydrogen --no-gui` flag to start just the core part (in the background) and `h2cli` won't setup H2 itself anymore but is able to control it via e.g. `h2cli stop` and `h2cli next` using the corresponding OSC commands under the hood. I don't know and I didn't want to go ahead without consulting anyone. I did my best to make it work with and without GUI but how to do the latter is still a matter of several design decisions.

> save_as is currently implemented in MidiActionManager, but i think there is no chance that this action can be triggered in any sensible way via midi (since you can't supply a filename). Wouldn't it be better suited in the core_action_controller class? I wrote that class when implementing common functionality which is used in the OSC server as well as in the UI. The initial idea was to avoid code duplication, but it could be also used for the implementation of actions which are not receivable via midi.

Yes, same thing for `open`. I but them there for consistency and to not scatter them all over the place. Shall I put all the session stuff in the `CoreActionController`?

mauser:
Hi! 

Ok, i will merge this pull request now and care for the ERRORLOGs. 

About h2cli/h2player: I can understand that confusion :) `h2player` is really old and outdated. I remember that when i started to tinker with Hydrogen, the binary didn't even compile. In my opinion, `h2player` was always a very exotic usecase. 

`h2cli` is a different thing. Even it is not mentioned in our documentation (well, most newer features are not mentioned there...) it is being used. People write bug reports about it or asking for more features, i know of at least of some people who are using in their headless setups or for easily exporting songs. I suppose it would be a good idea to add it to our Feature list. 
About the future of h2ci: It would be totally reasonable to discuss another design, especially in the light of the OSC support. Personally,  i absolutely dislike the current design of h2cli because of its code duplication.  Currently the only downside i can see (of the OSC client/server solution) would be the hassle of starting first the server, then client. In addition, the current main() Function is located in the UI, so there would be some refactoring needed. 

As a compromise for the current situation... i saw that you implented quit() via the EVENT_QUIT. Since the main loop of h2cli is also consuming events from the event queue, we could just add a case for EVENT_QUIT there. 

> Yes, same thing for open. I but them there for consistency and to not scatter them all over the place. Shall I put all the session stuff in the CoreActionController?

That would be great!

src/core/include/hydrogen/nsm_client.h
```diff
@@ -71,6 +71,13 @@ class NsmClient : public H2Core::Object
 
 		void shutdown();
 
+		/**
+		 * To determine whether Hydrogen is under Non session management,
+		 * it is not sufficient to check whether the NSM_URL environmental
+		 * variable is set but also whether.
```
mauser: Well, the end of the sentence seems to be missing :)

theGreatWhiteShark:
Alright. I'll have a look at the `h2cli` and make the quit work for it too.

trebmuh:
@theGreatWhiteShark can you update [the comment](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/include/hydrogen/nsm_client.h#L74) as @mauser suggested above? This sentence looks quite confusing as is.

theGreatWhiteShark:
I'm on it

Issue #786, opened 2019-11-12
Make auto audio driver display the selected driver


theGreatWhiteShark:
When using the `auto` selection in Preferences > Audio system the auto driver will select a proper driver based on the operation system and some heuristics (order to try the different ones).

But, unfortunately, it does not communicate its decision and the user is unable to view the selected driver in the Preferences dialog.

trebmuh:
Agreed.
Could be great to see something like "Auto (whatever_audio_driver_choosen)".

thijz:
there have been many questions asked about this 'auto' option.  would indeed be good to show the driver that was auto-selected

mauser:
Closed: the related pull request has already been merged.

PR #787, opened 2019-11-17
display choice of auto driver

theGreatWhiteShark:
When choosing the **auto** driver, the `__class_name` of the selected audio driver will be displayed in the Preferences menu.

preference-auto-driver

Solves #786


trebmuh:
Nice.

Issue #590, opened 2018-04-21
Non-Session: Does not save / link drumkit samples in NSM session dir


diovudau:
Hydrogen version *          : 0.9.7
Operating system + version  :  Arch Linux 
Audio driver + version      :  Jack2 1.9.12

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
NSM benefits from storing all non-program data like samples or hydrogen drumkits in the session dir.
To avoid redundancy it is recommended to link the data into the session dir on first use and from then on only use the link.

Other NSM aware programs like DrumkV1 already use this method.

Relevant section of the NSM Api:
http://non.tuxfamily.org/nsm/API.html#n:1.1.2.2.

1.1.2.2. External Files
Files required by the project but external to it (typically read-only data such as audio samples) SHOULD be referenced by creating a symbolic link within the assigned session area, and then referring to the symlink. This allows sessions to be archived and transported simply (e.g. with "tar -h") by tools that have no knowledge of the project formats of the various clients in the session. The symlinks thus created should, at the very least, be named after the files they refer to (some unique component may be required to prevent collisions)

theGreatWhiteShark:
Closed with #792

PR #792, opened 2019-12-26
Backup of NSM sessions.

### LINKED ISSUE #373: Save and Restore Input Mode
If I set the input mode to Instrument, there seems to be no way to save and restore this setting for next time. Particularly when using hydrogen under session management (NSM), this is a problem since manual intervention is required, defeating the point of session management.

theGreatWhiteShark:
### What's new?

When under session management (H2 is called by the (Non|New) Session Manager) the user configuration file is copied and the current drumkit is linked into the corresponding session folder. If one wishes to archive the session, there is no tedious and error prone copy&pasting / restoring of the $HOME/.hydrogen folder anymore. A simple archiving with dereferencing is enough, e.g. `tar -chf`. If present, H2 will load the `Preferences` and samples associated with the drumkit of the session from the session folder. This way it is fully self-contained and can easily be restored.

With this pull request Hydrogen is finally compliant with the NSM specification.

Fixes #373 and implements #590.

### What can go wrong/limitations?

The "Single pane"/"tabbed view" is not supported when switching between different session. H2 needs to be restarted to adapt to it.

Project > Save or Ctrl+S just saves the current `Song` while the event triggered by the "Save" button of the session manager will store both the latest changes of the `Song` and the `Preferences` (including everything related to the GUI). This discrepancy is unfortunately necessary to not loose any information/setting when switching sessions.

I get some redrawing/resizing issues when switching between sessions with different geometries stored in the settings. But I'm using `I3` as a window manager, which is not known for handling floating windows properly (moving the GUI to another workspace fixes potential glitches).


grammoboy2:
I did build your version from here: https://github.com/theGreatWhiteShark/hydrogen/tree/nsmBackup
Hydrogen still has a 'save as' function in nsm. This is against the nsm API. If I save as outside the session folder, it doesn't get reloaded. 

http://non.tuxfamily.org/nsm/API.html#n:1.

> 
> 1.1.1.4. Save As
> 
> This option MUST be disabled.
> 
> The application may, however, elect to implement an option called 'Export from Session', which creates a copy of the current file/project which is then saved in a user-specified location outside of the session path provided by NSM.

grammoboy2:
There is still the option 'open recent'. This is not according to the nsm api

> 1.1.1.2. Open
> 
> This option MUST be disabled.
> 
> The application may, however, elect to implement an option called 'Import into Session', creates a copy of a file/project which is then saved at the session path provided by NSM. 
> 
>

grammoboy2:
There is still a option 'open'

> 
> 1.1.1.2. Open
> 
> This option MUST be disabled.
> 
> The application may, however, elect to implement an option called 'Import into Session', creates a copy of a file/project which is then saved at the session path provided by NSM.

grammoboy2:
There is still a option 'open pattern'. 

> 1.1.1.2. Open
> 
> This option MUST be disabled.
> 
> The application may, however, elect to implement an option called 'Import into Session', creates a copy of a file/project which is then saved at the session path provided by NSM. 
> 
>

grammoboy2:
There is still a option 'open demo'. This is not right according to the nsm api.

I think this and all my previous comments also apply to the whole 'drumkits' menu.

grammoboy2:
I think all my previous comments also apply to the whole 'drumkits' menu.

grammoboy2:
'Save' opens a window where I can choose the location

> 1.1.1.3. Save
> 
> This option should behave as normal, saving the current file/project as established by the NSM open message.
> 
> UNDER NO CIRCUMSTANCES should this option present the user with a choice of where to save the file.

theGreatWhiteShark:
The pull request does not alter the default behavior of Hydrogen. Only when launched within a NSM session, `save as` will be replaced by `export from session`, `open` by `import to session` and so on.

You have to install the compiled software using `sudo make install` since `non-session-manager` will try to start the application using the system's `$PATH` variable.

When called from within NSM, the `NSM_URL` (or `NSM_PATH` .. something like this) variable will be set and Hydrogen will consider itself under session management and displays another set of menu options.

Opening  a drumkit works differently. The normal behavior is to just take the default settings of the drumkit and its name and write it inside the *.h2song file. When loading a song all samples will be loaded from the drumkit corresponding to the stored name.

In this pull request, the drumkit used will be symbolically linked into the session folder, samples will be read from there, and opening a new one will be replace link. This way you can bundle the session folder without any loss of information.

grammoboy2:
On 20-02-2020 12:58, theGreatWhiteShark wrote:
>
> The pull request does not alter the default behavior of Hydrogen. Only 
> when launched within a NSM session, |save as| will be replaced by 
> |export from session|, |open| by |import to session| and so on.
>
> You have to install the compiled software using |sudo make install| 
> since |non-session-manager| will try to start the application using 
> the system's |$PATH| variable.
>
> When called from within NSM, the |NSM_URL| (or |NSM_PATH| .. something 
> like this) variable will be set and Hydrogen will consider itself 
> under session management and displays another set of menu options.
>
> Opening a drumkit works differently. The normal behavior is to just 
> take the default settings of the drumkit and its name and write it 
> inside the *.h2song file. When loading a song all samples will be 
> loaded from the drumkit corresponding to the stored name.
>
> In this pull request, the drumkit used will be symbolically linked 
> into the session folder, samples will be read from there, and opening 
> a new one will be replace link. This way you can bundle the session 
> folder without any loss of information.
>
> —
>
I did build it with fakeroot dpkg-buildpackage

will try to build it via make

grammoboy2:
hydrogen: error while loading shared libraries: 
libhydrogen-core-1.0.0.so: cannot open shared object file: No such file 
or directory

Hm cannot start hydrogen

grammoboy2:
Ok. Have the version running.

When I hit 'new' and then 'save'. I do get a window where I can choose where to save a song.

theGreatWhiteShark:
If you see options like 'New' or 'Save as', your Hydrogen is still not under session management. https://github.com/theGreatWhiteShark/hydrogen/blob/nsmBackup/src/gui/src/MainForm.cpp#L288

Please be sure to install Hydrogen using `sudo`, to open it by pressing "Add Client to Session" inside the Non Session Manager, and to enter `hydrogen` in the appearing form.

Even if there's a problem with the pull request to should see (colored) log messages like

```
[Hydrogen] ...
```

in your terminal when starting NSM via the command line (using the `non-session-manager`). If you don't, NSM is not starting your newly compiled version of the program.

grammoboy2:
On 21-02-2020 21:53, theGreatWhiteShark wrote:
>
> If you see options like 'New' or 'Save as', your Hydrogen is still not
> under session management.
> https://github.com/theGreatWhiteShark/hydrogen/blob/nsmBackup/src/gui/src/MainForm.cpp#L288
>
> Please be sure to install Hydrogen using |sudo|, to open it by
> pressing "Add Client to Session" inside the Non Session Manager, and
> to enter |hydrogen| in the appearing form.
>
> Even if there's a problem with the pull request to should see
> (colored) log messages like
>
> |[Hydrogen] ... |
>
> in your terminal when starting NSM via the command line (using the
> |non-session-manager|). If you don't, NSM is not starting your newly
> compiled version of the program.
>
> —
>
Build in on a other computer. This are the nsmd messages.

I still do have 'new' and 'save as' options.

> [non-session-manager] Sending add for: hydrogen
> [nsmd] Process has pid: 10920
> [nsmd] Launching hydrogen
>
> Hydrogen 1.0.0-beta1 [Feb 22 2020]  [http://www.hydrogen-music.org]
> Copyright 2002-2008 Alessandro Cominu
> Copyright 2008-2018 The hydrogen development team
> Hydrogen comes with ABSOLUTELY NO WARRANTY
> This is free software, and you are welcome to redistribute it under
> certain conditions. See the file COPYING for details.
> [nsmd] Got announce from Hydrogen
> [nsmd] Client was expected.
> [nsmd] Process has pid: 10920
> [nsmd] The client "Hydrogen" at "osc.udp://127.0.0.1:14571/" informs
> us it's ready to receive commands.
> NSM: Successfully registered. NSM says: Howdy, what took you so
> long?[nsmd] Client "Hydrogen" replied with: OK in 1.403000ms
>
>

grammoboy2:
I think it would be good if 'new' would act as 'reset' with a confirmation dialog

> 1.1.1.1. New
> 
> This option may empty/reset the current file or project (possibly after user confirmation). UNDER NO CIRCUMSTANCES should it allow the user to create a new project/file in another location.

theGreatWhiteShark:
You still haven't installed the version of this pull request in such a way NSM is able to find it. When called by NSM Hydrogen runs the `nsm_open_cb` function https://github.com/theGreatWhiteShark/hydrogen/blob/nsmBackup/src/core/src/nsm_client.cpp#L144 . With the new version their is at least some output to stdout or stderr.

If you are unsure whether the installed version is the default one executed by your system, you could for instance add a message to std::cout at the beginning of the main.cpp file or alter the about text https://github.com/theGreatWhiteShark/hydrogen/blob/nsmBackup/src/gui/src/main.cpp#L160, reinstall, and call the application via the command line. You should see the altered/added messages.

> I think it would be good if 'new' would act as 'reset' with a confirmation dialog

Right now "New" is replaced by "Replace with new Song" and asks for consent before doing so

grammoboy2:
Ok, this is a bit confusing. On the other hand, it's good news that I
don't have the right version for some reason. :)

Could you tell me exactly how I should install the nsm version of hydrogen?

This is what I did:

sudo apt purge hydrogen hydrogen-data hydrogen-doc

git clone https://github.com/theGreatWhiteShark/hydrogen.git

cd hydrogen

mkdir build

cd build

cmake ..

make

sudo make install

sudo ln -s /usr/local/lib/libhydrogen-core-1.0.0.so
/usr/lib/libhydrogen-core-1.0.0.so

theGreatWhiteShark:
You forgot to checkout the git branch containing this patch.

``` bash
git clone https://github.com/theGreatWhiteShark/hydrogen.git
cd hydrogen
git checkout nsmBackup
./build.sh mm
cd build
sudo make install
```

AFAIK there should be no problem with the systems' Hydrogen being installed (in /usr/bin and /usr/lib) next compiled one (in /usr/local/bin and /usr/local/lib). Also I don't understand why you do need the last linking step.

grammoboy2:
Thanks, works now.

  * One issue which I found is that Hydrogen seems to autoconnect JACK.
    That's not allowed when running in a session manager afaik.

  * Nsm displays this message, when using the default drumkit:

[Hydrogen] Error: Symlink does not point to valid drumkit
[nsm_open_cb] relinking
[Hydrogen] Error: Unable to link [/home/user/NSM
Sessions/Test./nsmhydrogen1/Hydrogen.nQPIE/drumkit] to
[/home/user/.hydrogen/data/drumkits/GMRockKit][


  * If you open a pattern, it's not copied into the session. This means
    you end up with a empty pattern, when you're reloading the session.

  * When you open a drumkit and load a different one. The symlink
    doesn't change in the session folder. The other drumkit is reloaded
    when closing and open the session in nsm.

Still:

lrwxrwxrwx 1 user user   46 feb 24 22:10 drumkit ->
/home/user.hydrogen/data/drumkits/GMRockKit

but I've actually TR808... loaded now.


  * When I import a drumkit (download it) and then load it. Then hit
    close on the nsm gui, the new drumkit isn't loaded when relaunching
    the session.
  * it is also possible to import a song and a pattern from Internet.
    Couldn't test it, but might be tricky as well.
  * There is also a option 'save as' and 'export' in the drumkit menu. I
    couldn't test it, but might be tricky too.


Some issues, but very promising. Thanks! Will wait for your feedback and
I'm happy to do more testing on request.

theGreatWhiteShark:
I'll look into your comments, I promise. But it may take a couple of weeks

grammoboy2:
Ok, take your time.

grammoboy2:
I think Qtractor is also using symlinks, might be a interesting example
to look into.

theGreatWhiteShark:
> One issue which I found is that Hydrogen seems to autoconnect JACK.
    That's not allowed when running in a session manager afaik.

This is a default setting of Hydrogen the user can toggle. It's not really against the rules to use a application other than `JACKPatch` to establish persistent connections within a session but, yes, I also disable it when using NSM. In addition, I don't want to alter anything in settings menu because it should still very much feel like the same application.

> If you open a pattern, it's not copied into the session. This means
    you end up with a empty pattern, when you're reloading the session.

- [ ] I never opened a pattern till now. No idea why. :laughing: I will check it out.

> When you open a drumkit and load a different one. The symlink
    doesn't change in the session folder. The other drumkit is reloaded
    when closing and open the session in nsm.

- [ ] During development everything worked just fine. But you are right. Sometimes the symlink is off.

> When I import a drumkit (download it) and then load it. Then hit
    close on the nsm gui, the new drumkit isn't loaded when relaunching
    the session.

- [ ] I probably missed that one.

> it is also possible to import a song and a pattern from Internet.
    Couldn't test it, but might be tricky as well.

- [ ] It is? :astonished: Never heard about that feature.

> There is also a option 'save as' and 'export' in the drumkit menu. I
    couldn't test it, but might be tricky too.

- [ ] That's no problem. All settings regarding the session (also for the drumkit itself) are stored within the .h2song file within the corresponding folder in the session directory. Only the samples are loaded directly. Therefore the symlink is sufficient. But I will check again if everything is fine when renaming the drumkit.

- [ ] Another thing I realized during the past weeks is that the symlink does not seem to be updated properly when manually copying another .h2song in the session folder (e.g. using the command line). 

Thanks a lot for your feedback! I will look into these things

hernando:
@theGreatWhiteShark, thanks for this PR, I'm testing it to see if it fixes the problems I've observed in my sessions after upgrading Hydrogen and Ardour to their latest versions and it fact it does.
However, the PR seems to introduce a regression. Within the session, the audio settings related to audio outputs are not taken into account at all. I would like to have one output per instrument and no default connections, and none of these things work. Let me know if I can help you identifying the cause of the problem, I'm comfortable doing software development in Linux.

theGreatWhiteShark:
Hey @hernando, you are right. I can reproduce this behavior.

The idea is that every session got it's own configuration. E.g. if you wire up all the per-instrument tracks of one session, you don't want this configuration to be discarded when toggling the per-track output option in another session or in blank Hydrogen.

Technically the Preferences in the user home/sys folder loaded during startup and initialization of Hydrogen's core. As soon as the NsmClient is initialized, it either reads the preference file in the session folder or copies the current configuration in there if none is present (`NsmClient::copyPreferences)`. The session-level preferences file will be then used at all instances using the `PreferencesOverwritePath` in the `Preferences`.

At some point the utilization of the session-level preferences is jammed. I will take a look into it.

theGreatWhiteShark:
@hernando after all it was just a couple of lines which got lost during the rebase. Your audio settings should apply now properly

hernando:
Thanks for the quick fix, I will try right away.

hernando:
Now the configuration is taken into account, but there's still a problem. I don't get any sound from the per-instruments output. I've started a new NSM session, added Hydrogen, disconnected the main outputs from the system playback inputs and connected one instrument directly to the playback inputs. When I trigger the instrument I can see the meter in H2's mixer rising, but nothing sounds. On the other hand, the main outputs produce sound.

theGreatWhiteShark:
> Now the configuration is taken into account, but there's still a problem. I don't get any sound from the per-instruments output. I've started a new NSM session, added Hydrogen, disconnected the main outputs from the system playback inputs and connected one instrument directly to the playback inputs. When I trigger the instrument I can see the meter in H2's mixer rising, but nothing sounds. On the other hand, the main outputs produce sound.

Ah. I see. This issue already appeared once. Seems like something else got lost during rebasing... Sorry, I will take care of it

theGreatWhiteShark:
@hernando has the sound issue been resolved?

hernando:
Yes, I was about to reply. Thanks for the fixes!

trebmuh:
@theGreatWhiteShark : I'm really interested about NSM in Hydrogen. Is that a good time for me to test and report to you? If so, could you please resolve the conflicting file in this branch, then I can compile and test. Thanks.

theGreatWhiteShark:
Hey @trebmuh, I'm not quite there yet. There are still some open points on my todo list regarding the linking and archiving of the drumkits and the consistent behavior of the GUI. But they are getting less and less every day.

Don't worry. I'm super into this too. It's like the main feature I'm missing to use Hydrogen during recordings in a convenient way. I ping you as soon as I am done.

trebmuh:
Awesome. I'll wait for your ping then.

theGreatWhiteShark:
@trebmuh everything should be in place now. If you like you can start testing

trebmuh:
@theGreatWhiteShark : Thanks.
I've rebuild it and will test it soon.
I'll let you know how it goes.

Houston4444:
Hi !
I tested the @trebmuh build, All seems to work fine, thanks a lot !
Just a very little thing, would it be possible to send a '/nsm/client/is_clean' message to the server once hydrogen is ready (There is no reason to be dirty at start ;) ), because raysession and agordejo displays the client dirty state. In RaySession, stop the client from GUI displays an optional warning if client has dirty changes.

Is there the possibility to add optional-gui capability ? If so, please do it start hidden or remember itself the GUI state (I prefer start hidden).

theGreatWhiteShark:
> Just a very little thing, would it be possible to send a '/nsm/client/is_clean' message to the server once hydrogen is ready

Alright.

It's not explicitly covered in the API standard. Does the server expect to receive the `is_clean` message the moment the song and preferences are saved - there are no unsaved changes?

> Is there the possibility to add optional-gui capability ? 

AFAIK it's current not possible to disband the GUI of Hydrogen

Houston4444:
> Does the server expect to receive the is_clean message the moment the song and preferences are saved - there are no unsaved changes?

I don't understand this question. There is no specification about when exactly client should send theses messages. I think, that if client is capable to send dirtiness as soon as possible, it is better for user. But, don't worry, this is a trivial question.

theGreatWhiteShark:
@Houston4444 Okay. I made Hydrogen send the clean message as soon as the GUI is ready and can express dirtiness. This will happen slightly prior to the reply of Hydrogen with respect to the 'open' message.

Houston4444:
> This will happen slightly prior to the reply of Hydrogen

That's fine.
But the merge checks failed.

src/core/include/hydrogen/Preferences.h
```diff
@@ -1293,6 +1317,13 @@ inline QString Preferences::getH2ProcessName() {
 	return m_sH2ProcessName;
 }
 
+inline QString Preferences::getPreferencesOverwritePath() {
```
mauser, line 1320: I'm wondering if this should go into our `Filesystem` class, since this already manages other things like the temporary directory or the user/system dir. 

But it is great that you've integrated this option already, this will come handy for the configuration for the hydrogen LV2 plugin!

src/core/src/core_action_controller.cpp
```diff
@@ -434,90 +444,118 @@ bool CoreActionController::saveSong() {
...
+		Preferences::get_instance()->savePreferences();
+	}
+	
+	return true;
+
+}
+
+bool CoreActionController::saveDrumkit() {
+
+	return true;
```
mauser, line 520: This looks like a stub :) Is this intended?

src/core/src/nsm_client.cpp
```diff
@@ -20,101 +20,414 @@
...
+			}
+		}
+	}
 }
 
-void* nsm_processEvent(void* data)
-{
-	nsm_client_t* nsm = (nsm_client_t*) data;
+void NsmClient::printError( const QString& msg ) {
+	std::cerr << "[\033[30mHydrogen\033[0m]\033[31m "
```
mauser, line 392: Errorlog?

src/core/src/nsm_client.cpp
```diff
@@ -20,101 +20,414 @@
...
+			  << "Error " << msg.toLocal8Bit().data() << "\033[0m" << std::endl;
+}
+void NsmClient::printMessage( const QString& msg ) {
+	std::cerr << "[\033[30mHydrogen\033[0m]\033[32m "
```
mauser, line 396: Errorlog?

src/core/src/nsm_client.cpp
```diff
@@ -20,101 +20,414 @@
...
-	m_bUnderSessionManagement = false;
+	return ERR_OK;
 }
 
-void NsmClient::create_instance()
-{
-	if( __instance == nullptr ) {
-		__instance = new NsmClient;
+void* NsmClient::ProcessEvent(void* data) {
+	nsm_client_t* nsm = (nsm_client_t*) data;
```
mauser, line 419: pNsm :smiley:

src/core/src/preferences.cpp
```diff
@@ -1106,26 +1127,34 @@ void Preferences::setMostRecentFX( QString FX_name )
...
-		for ( unsigned j = 0; j < temp.size(); j++ ) {
-			if ( sFilename == temp[ j ] ) {
-				bExists = true;
-				break;
-			}
-		}
-		if ( !bExists ) {
-			temp.push_back( sFilename );
+	std::vector<QString> sTmpVec;
+	for ( auto ssFilename : recentFiles ) {
```
mauser, line 1149: Small optimisation: I suppose "auto&" or "const auto&"  would skip some copy constructor calls for QString

src/gui/src/MainForm.cpp
```diff
@@ -1347,50 +1481,27 @@ void MainForm::updateRecentUsedSongList()
...
-}
-
-
-
-void MainForm::openSongFile( const QString& sFilename )
-{
-	Hydrogen *pEngine = Hydrogen::get_instance();
-	if ( pEngine->getState() == STATE_PLAYING ) {
-		pEngine->sequencer_stop();
+#ifdef H2CORE_HAVE_OSC
```
mauser, line 1484: I'm wondering if we could get rid of many of those preprocessor magic if we would put the preprocesser macros inside the NSM class iso. of around its uses. Would this be possible? I did sth. similar with the Ladspa includes in my lv2 playground branch. This had the advantage that the compiler and static analysis tool always see as much as code as possible and would problems with variable renaming in those sections.

 Of course that might mean that NsmClient::get_instance() might be a nullptr or that the NsmClient class itself would have an "enabled" flag or sth. like that.

src/core/include/hydrogen/core_action_controller.h
```diff
@@ -98,9 +98,11 @@ class CoreActionController : public H2Core::Object {
 		 *
 		 * \param songPath Absolute path to the .h2song file to be
 		 *    opened.
+		 * \param bRestartDriver Whether or not to restart the audio
+		 *    driver after successfully loading the new song.
 		 * \return true on success
 		 */
-		bool openSong( const QString& songPath );
+		bool openSong( const QString& songPath, bool bRestartDriver );
```
mauser, line 105: This "restart drivers after song load" logic seems to be kind of strange to me and makes the song loading process even more complicated. But i guess there no easy way around, at least i also don't see a better solution at the moment :confused:

src/core/include/hydrogen/Preferences.h
theGreatWhiteShark: Sure. Moving it to `Filesystem` will keep the `Preferences` class more focused.

The reason I didn't put functions like `NsmClient::copyPreferences` and `NsmClient::linkDrumkit` in there in the first place is because - when under session management - Hydrogen will always be called by the NSM daemon. So, all verbose output has to be taken from it's log output (everything that goes to stderr) and there is no way to provide commandline flags for debugging. Using `NsmClient::printError` and `NsmClient::printMessage` one gets both a "[Hydrogen] " prefix required to identify the corresponding line of log and everything is visible since it's directed to stderr. On the other hand, our log messages are directed to stdout - even the ERRORLOG ones.

If either `copyPreferences` or `linkDrumkit` are required in other parts of Hydrogen too, we can put them in `Filesystem` and add some more logic for the logging. Putting them there in their current version might be more confusing then helpful

src/core/src/core_action_controller.cpp
theGreatWhiteShark: You are right. I decided against implementing this one :)

src/core/include/hydrogen/Preferences.h
theGreatWhiteShark: Small correction: the original `non-session-manager` only supports messages directed to stderr. The tweaked `nsm-legacy-gui` (it's fork) of the New Session Manager does display the ERRORLOGs as well.

src/core/src/nsm_client.cpp
theGreatWhiteShark: :+1:

src/core/src/nsm_client.cpp
theGreatWhiteShark: As pointed out above the log of the logger won't be visible when called by the NSM daemon since it's not directed to stderr. 

This is also the reason I introduced all this raw Qt-based file copying instead of using the corresponding functions from `Filesystem`. Problems wouldn't show up in the log of non-session-manager

src/core/src/nsm_client.cpp
theGreatWhiteShark: see above

src/core/src/preferences.cpp
theGreatWhiteShark: :+1:

src/core/include/hydrogen/core_action_controller.h
theGreatWhiteShark: Yeah. I added another commit in order to streamline the loading a little more. But yes. It's rather complicated and this adds another layer. I removed this argument and replaced it by some checks in the `CoreActionController` itself. Only if Hydrogen is under session management the audio driver will be restarted. 

Still, the most puzzling (and probably redundant) part for me is that the Song has to be loaded by the GUI. This leads to really awkward program flow as the NsmClient reads a song from disk and stores it into a variable. The GUI than takes this song and asks the `Hydrogen` class to set it. But when just omitting the GUI part, the application freezes. I haven't looked into it much but this core -> gui -> core part can most probably be simplified. (But this is not directly related to NSM so I did not attempted it in this PR).

src/gui/src/MainForm.cpp
theGreatWhiteShark: sure

mauser:
Hi! 

it seems that there is an include guard missing somewhere: 
```

In file included from C:\projects\hydrogen\src\core\NsmClient.cpp:31:
C:/projects/hydrogen/src/core/Nsm.h:79:10: fatal error: lo/lo.h: No such file or directory
   79 | #include <lo/lo.h>
      |          ^~~~~~~~~
```
See https://ci.appveyor.com/project/mauser/hydrogen/build/job/0fuq72a1dhxnmbie

mauser:
Hey! 

can i do anything to assist with this branch? Can it be merged or does it need testing or anything else? I'm not a NSM user, but offering my help anyway :)

theGreatWhiteShark:
I think after the feedback of the NSM users, your code review, and the rebasing we are finally good to go. And can merge it :)

PR #801, opened 2020-02-21
Fix CPPUNIT include-directories

rrrapha:


mauser:
Thanks a lot, merged!

PR #802, opened 2020-02-22
Fix Library Panel crash

tatokis:
Commit e9a1892dd0 caused a regression where having an empty song
list would result in a crash due to __song_item being uninitialised.

Previously QTreeWidget::isItemExpanded() would be called, which
would simply return false if nullptr would be passed to it.


mauser:
Hi! 

i've i understand your issue correctly, then it has been already addressed by @jeremyz in 9dfab9acc576e6bbe99e6f519d948d16094a6509. Anyway, thanks for the report and the fix!

PR #689, opened 2019-01-22
Install icon in prefix, in proper location, rename .desktop and appdata to proper id name

hfiguiere:

PR #815, opened 2020-03-22
Fix installation of icon

hfiguiere:
Regression caused by #689.

tl;dr INSTALL doesn't rename the file without RENAME


mauser:
Thanks, merged!

Issue #819, opened 2020-03-25
[beta2] translations broken


trebmuh:
It seems that the translation system is broken. I've tried to make a LibraZiK package for beta2, but when launching Hydrogen, it's always in English.

Note: I'm using a French locale.

Launching it with `hydrogen -VDebug` from the CLI, I've got those which looks interesting:

```
...
...
(I) Filesystem::info Internationalization dir   : /usr/share/hydrogen/data/i18n/
...
...
(I) ::int main(int, char**) Using QT version 5.11.3
(I) ::int main(int, char**) Using data path: /usr/share/hydrogen/data/
(I) ::int main(int, char**) Warning: no locale found: /usr/share/hydrogen/data/i18n/hydrogen.fr_FR
(I) ::int main(int, char**) Warning: error loading locale: /usr/share/hydrogen/data/i18n//hydrogen.fr_FR.qm
...
...
```

Note that:

```
$ ls -la /usr/share/hydrogen/data/i18n/hydrogen_fr*
-rw-r--r-- 1 root root 109417 mars  25 16:01 /usr/share/hydrogen/data/i18n/hydrogen_fr.qm
```

It looks to me that hydrogen is looking for a `hydrogen.fr_FR.qm` while there is only a `hydrogen_fr.qm` one.

trebmuh:
Note that if I rename  `/usr/share/hydrogen/data/i18n/hydrogen_fr.qm` for  `/usr/share/hydrogen/data/i18n/hydrogen.fr_FR.qm` as expected by Hydrogen, I've got my Hydrogen in French. Renaming it for `/usr/share/hydrogen/data/i18n/hydrogen.fr.qm` looks to be working as well.

mauser:
Hi! 

i've made a recent change 70e828afed2f737d4cb2fa193612bad712c95fdf to the CMakeLists file in the i18n dir. This has been more of a temporary nature for the windows build. Can you try if it works with an earlier commit?

trebmuh:
Hi @mauser , I've applied [jpcima's patch](https://github.com/hydrogen-music/hydrogen/pull/820) and it seems to work here. Do you still need me to try a commit before 70e828a?

mauser:
Hi,

the patch is applied, so i'm closing this issue. @trebmuh : Just re-open it if the problem persists.

trebmuh:
OK, I'll wait for beta3 to recheck this.

PR #820, opened 2020-03-25
Fix the search of translation files

jpcima:
#819 

This lets Qt locate the correct translation file, regardless of how is constructed the locale name.
(eg. "fr" vs "fr_FR")

The solution indicated in this comment is not sufficient.
https://github.com/hydrogen-music/hydrogen/issues/819#issuecomment-604038793


trebmuh:
I can confirm that it works here (Debian stable).

elpescado:
That looks good to me. Much better and cleaner.

mauser:
Thanks a lot, merged!

Issue #821, opened 2020-03-25
[beta2] Mac installer is missing translations


elpescado:
Hydrogen version *          : 1.0.0-beta2
Operating system + version  : macOS 10.13
Audio driver + version      : n/a

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Mac package is missing compiled translation files so Hydrogen is not localised (UI is in English regardless of system language settings).

![Zrzut ekranu 2020-03-25 o 22 19 23](https://user-images.githubusercontent.com/1619113/77586639-d5663480-6ee6-11ea-8fb6-2e498b3e93b3.png)

trebmuh:
@elpescado probably due to https://github.com/hydrogen-music/hydrogen/issues/819

elpescado:
I think that's unrelated.

As you can see in attached screenshot, there are source translation files (*.ts, files used and edited by translators) included in package, but there are no compiled translations (*.qm, those are actually used by application). Some time ago we removed compiled translations from git repository (which was a good thing) and compile them as part of build process. Somehow yet they didn't make their way to package file. (As a side issue, *.ts files are totally unnecessary in binary package and could be skipped).

I'll investigate it, but I cannot assign myself to this issue.

trebmuh:
assigned!

mauser:
When building with Linux, my build/data/i18n directory holds the following files:

```
drwxr-xr-x 3 sebastian sebastian 4,0K Mär 26 21:27 CMakeFiles
-rw-r--r-- 1 sebastian sebastian 2,5K Mär 26 21:27 cmake_install.cmake
-rw-r--r-- 1 sebastian sebastian  56K Mär 26 21:27 hydrogen_ca.qm
-rw-r--r-- 1 sebastian sebastian  53K Mär 26 21:27 hydrogen_cs.qm
-rw-r--r-- 1 sebastian sebastian 105K Mär 26 21:27 hydrogen_de.qm
-rw-r--r-- 1 sebastian sebastian 110K Mär 26 21:27 hydrogen_el.qm
-rw-r--r-- 1 sebastian sebastian 104K Mär 26 21:27 hydrogen_es.qm
-rw-r--r-- 1 sebastian sebastian 107K Mär 26 21:27 hydrogen_fr.qm
-rw-r--r-- 1 sebastian sebastian  85K Mär 26 21:27 hydrogen_gl.qm
[...]

PR #823, opened 2020-03-30
macOS: Deploy i18n files in .dmg image

elpescado:
Fix for #821.

Translation files were built but not included in mac .dmg image. This happened on Mac only, because Mac package is created using custom script.


elpescado:
Mac bundle after:

Zrzut ekranu 2020-03-30 o 21 30 09

mauser:
Thanks a lot, merged!

PR #835, opened 2020-04-05
Correct midi channel export (fixes #580)

### LINKED ISSUE #580: [FEATURE REQUEST] Export midi files with multiple channels
_(from http://hydrogen.popez.org/hcms/node/17532)_

> Is there a way to export a midi file that contains more than 1 channel?  I have combined kits that have drums and accompiniment instruments (bass) to create rhythm tracks to use through my DR-5.  I have placed all of the drums into channel 10 and the bass notes to channel 2 but when exporting a midi file the file only contains the drums.  The bass notes do not show on any midi file editing software I have.  Will this become a supported feature in future releases of Hydrogen?

gloryofrobots:
This PR fixes #580  

Currently H2 gives user an ability to set custom channel to a specific instrument however it exports all midi notes into 10-th drum dedicated channel.
This PR fixes it so midi export will write proper channel data for all notes.
As a positive side effect this will allow to use midi multitrack export with LMMS, because LMMS squishes all notes from one channel into one track. So for LMMS user will need to set each track to separate channel

Changes are very small and localized in one file smf.cpp and they can be quickly reviewed.

To test those changes you need some kind of a midi editor, midi viewer or midi friendly DAW. 

I can not find how to test this in Ardour it looks like this program omits midi channels entirely.

In Rosegarden after midi import the channels for each track will be on the left info panel under the name "instrument" (screenshot 1)

In Qtractor you need to double click on the track  after import and look at the track options    (screenshot 2)

rose

qtractor

### DIFF
--- a/src/core/src/smf/smf.cpp
+++ b/src/core/src/smf/smf.cpp
@@ -207,6 +207,7 @@ vector<char> SMF::getBuffer()
 
 const unsigned int TPQN = 192;
 const unsigned int DRUM_CHANNEL = 9;
+const unsigned int NOTE_LENGTH = 12;
 
 const char* SMFWriter::__class_name = "SMFWriter";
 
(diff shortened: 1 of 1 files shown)


mauser:
Hi! 

Thanks for the pull request! The feature is now part of our master branch.

trebmuh:
congratz @gloryofrobots !

PR #837, opened 2020-04-10
Osc patch

theGreatWhiteShark:
Some minor patches to #773 

1. All the session management actions have been moved from *midi_action* into *core_action_controller* since most of the corresponding actions do require a string input and wouldn't work with MIDI anyway.
2. The `h2cli` can now be terminated via OSC

``` bash
oscsend localhost 9000 /Hydrogen/QUIT
```

3. I wrote unit tests for all session management methods. (and I intend to do so for all future contributions)

---

Worth noting:

- I added a dependency to `QTest` to have a platform-independent sleep function.
- I had to set up a dummy OSC server to test the handler functions. Everytime I used our own `OscServer` class (either the one bootstrapped along with Hydrogen or a fresh one constructed in the fixture) I got a segmentation fault. I looked several hours into it but couldn't figure out what was going wrong (or what's the big difference between the dummy server and the real one. The segfault always occurred at the very end of the tests (after all runner finished) and it appear as soon as I was using the `send()` method of a `lo::Address` to send an arbitrary message to port Hydrogen is registered at.


mauser:
Thanks a lot, merged!

PR #847, opened 2020-04-23
Keyboard cursor-driven pattern, song and note properties editing

cme:
Cursor-key driven movement and pattern, song & note properties editing.

Aims to be obvious-to-discover for the user. Currently uses only the standard movement keys and Enter to place/clear notes and patterns.

Also sorts out tab ordering, and the annoying spurious movement of widgets that appears when tabbing between focused widgets.


mauser:
Hey,

i have to say that i like feature this a lot :) Thanks for your work!! 

Two questions came to my mind when using your branch:

1. I can see that when i press tab once, the song editor is selected. Then comes the pattern editor, then the note properties editor. After that, i can't figure out where the focus goes when i press tab. Are there more editable things? 

2. In my opinion, it is confusing the the "input focus square" is also displayed if i'm using the mouse. I would like it better if the focus box disappears after a mouse press and is only used for keyboard input (activated by tab-press).

But after all, i can imagine that this feature will be a benefit for many people! 

Last but not least some words to your code: It would be great if you could avoid if statements withouth braces. We have no detailled coding guidelines, but there are some smaller rules in **https://github.com/hydrogen-music/hydrogen/blob/master/DEVELOPERS** .  And yes, those guidelines are not followed in the current code at all places, but we're adressing that :smile:

jeremyz:
I like it a lot !
maybe kind of what @mauser said, hide cursor on mouse click,
show it up on handled key press (tab, enter, left, right, ...)

cme:
Thanks a lot for the feedback guys :)

>     1. I can see that when i press tab once, the song editor is selected. Then comes the pattern editor, then the note properties editor. After that, i can't figure out where the focus goes when i press tab. Are there more editable things?

After the pattern editor, the focus chain flows to the instrument editor where there are a couple of checkboxes that can be focused. I don't see much value in having these editable by keyboard, so it's probably better to set them to not focus. Will do that.

>     2. In my opinion, it is confusing the the "input focus square" is also displayed if i'm using the mouse. I would like it better if the focus box disappears after a mouse press and is only used for keyboard input (activated by tab-press).

I like this, as I had similar thoughts. My only worry though is that the usual user expectation for clicking something in a document would set the input focus to that location, particularly if they're editing with the keyboard and want to use the mouse to move to a nearby location.

After a little more thought, I propose:
  - mouse clicks should hide the input focus cursor
  - as @jeremyz suggests, cursor movement keys or tab would reveal the focus cursor
  - input focus should follow mouse clicks as currently, **but** with the cursor hidden.

This way
  - mouse-only operation is completely unaffected for existing users
  - keyboard + mouse operation is still possible (with the minor wrinkle that the new input focus cursor position won't show up until the next keypress. This seems very minor since the user knows where they just clicked!)


> Last but not least some words to your code: It would be great if you could avoid if statements withouth braces. We have no detailled coding guidelines, but there are some smaller rules in **https://github.com/hydrogen-music/hydrogen/blob/master/DEVELOPERS** . And yes, those guidelines are not followed in the current code at all places, but we're adressing that smile

Ah! I read that last year but didn't remember about the 'if' rule. Will update, thank you :)

cme:
Changes outlined above made, and also realised that there was indeed some fairly obvious keyboard behaviour that should be implemented in Select mode in the Song Editor:

  - Enter will start or end a select or move/copy, in the same way that pressing or releasing a mouse button would
  - The standard Selection keys (I think shift+arrow on all platforms, though Qt abstracts it) will begin a selection regardless of whether the input cursor is currently on an active pattern or not.
  - Escape will cancel a selection or move. Which is also useful when using the mouse (I realise now that I've actually wished for this behaviour before)

elpescado:
I'd keep `override` keywords. To get rid of warnings, I'd rather disable it with `-Wno-inconsistent-missing-override` (clang) and `-Wno-suggest-override` (gcc), or - better - add `override` to all overridden methods.

cme:
> I'd keep `override` keywords. To get rid of warnings, [...] , or - better - add `override` to all overridden methods.

I entirely agree, but that's going to be a separate pull request ;)

elpescado:
I *think* it might be done automatically:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html

mauser:
Hi! 

Tested it again, and it works like a charm! I'm just unsure about the default input mode. Should the input focus marker be activated by default or not? 

If the marker is shown by default, it adds a little bit more visual clutter to the UI and may confuse people (because it simply disappears after a click). 

If the marker is not shown by default, people may not be aware of this new feature. On the other hand, just by seeing the marker, one does not know how to use it or understand what its good for.

@cme, @elpescado : What is your opinion about the default behavior? My personal favourite would be to hide it as a default.

cme:
Hmm, the possibility for confusion on seeing something disappear that they wouldn't know how to get back is something that didn't occur to me before. I think now I'm leaning towards hiding it by default. As you say, the only way for the user to discover that it's an input cursor is to hit a key; which is the same condition that having it hidden-by-default will unveil it in the first place.

cme:
Done, since we all seem to be in agreement. Thanks.

Issue #844, opened 2020-04-20
Mac OS X Version of Hydrogen 1.0-beta2 does not support OS X 11.x


mauser:
Hydrogen version *          : 1.0-beta2
Operating system + version  : Mac OS X 11.6

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
I tried today to update the Hydrogen version on my old update, and i ran directly into an issue. When starting the application (via commandline, to get debug output..) an error messages appears which informs me that QT (at least the bundled version..) needs Mac OS X 10.12. 

@elpescado : I'm not sure if we can do anything about it, but maybe it is possible to use an older QT version? I remember that it worked when you initially set up the deployment process, i'm still using the Hydrogen image from that time.

elpescado:
Hi @mauser,

What's that message? Does Hydrogen start anyway, or is it fatal error?

We set `MACOSX_DEPLOYMENT_TARGET="10.10"` env in Travis in order to build binary compatible with 10.10, but that applies to Hydrogen binary itself. It seems that third-party libraries might have their own compatibility requirements. As I checked, we currently use Qt 5.13.2 which seems to [support macOS 10.12, 10.13, 10.14](https://doc.qt.io/archives/qt-5.13/supported-platforms.html). I've compiled compatibility information for various Qt versions:
```
Qt     macOS                                     Reference
-------------------------------------------------------------------------------------------------------------
5.8    10.9 10.10 10.11 10.12                    https://doc.qt.io/archives/qt-5.8/supported-platforms.html
5.9         10.10 10.11 10.12                    https://doc-snapshots.qt.io/qt5-5.9/supported-platforms.html
5.10              10.11 10.12 10.13              https://doc.qt.io/archives/qt-5.10/supported-platforms.html
5.11              10.11 10.12 10.13              https://doc.qt.io/archives/qt-5.11/supported-platforms.html
5.12                    10.12 10.13 10.14        https://doc.qt.io/qt-5.12/supported-platforms.html
5.13                    10.12 10.13 10.14        https://doc.qt.io/archives/qt-5.13/supported-platforms.html
5.14                          10.13 10.14 10.15  https://doc.qt.io/qt-5/supported-platforms.html
```
[...]

PR #849, opened 2020-04-26
#844 Create second Mac job to build packages for macOS 10.10..10.12

elpescado:
I've created second macOS job on Travis CI to build Hydrogen against Qt in order to support macOS releases 10.10, 10.11 and 10.12 - regular build requires 10.12 or newer.

Here are example builds: https://github.com/elpescado/hydrogen/releases/tag/legacy5

That build works on my Mac running 10.12. Unfortunately I no longer have any VMs with older macOS releases, so I'm unable to test it on those releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,14 @@ jobs:
   - os: linux
     dist: xenial
   - os: osx
-    env: MACOSX_DEPLOYMENT_TARGET="10.10"
+    env:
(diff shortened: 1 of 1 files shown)


mauser:
Hi! 

thanks a lot, it works perfectly!

trebmuh:
Awesome!

Issue #822, opened 2020-03-25
[beta2] Hydrogen on Retina Macs is pixelated


elpescado:
Hydrogen version *          : 1.0.0-beta2
Operating system + version  : macOS 10.14
Audio driver + version      : n/a

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Hydrogen on HighDPI (Retina) screen on Mac looks pixelated. Compare Hydrogen window to Finder window below it:

<img width="1023" alt="Screenshot 2020-03-25 at 22 42 27" src="https://user-images.githubusercontent.com/1619113/77588413-1c095e00-6eea-11ea-83dd-6a6d01fb4a62.png">

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

elpescado:
I'd like to investigate this issue and assign myself to this issue but I can't do that:(

trebmuh:
done

PR #850, opened 2020-04-26
#822 Set NSHighResolutionCapable to true in Info.plist

elpescado:
Fix for #822. It seems that one entry was missing in Hydrogen manifest (`Info.plist`), thus Hydrogen didn't advertise HiDPI capability.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -44,16 +44,18 @@ ENDIF()
 
 ADD_EXECUTABLE(hydrogen WIN32 MACOSX_BUNDLE ${hydrogen_SRCS} ${hydrogen_MOC} ${hydrogen_UIS_H} ${CMAKE_SOURCE_DIR}/windows/icon.rc)
 SET_PROPERTY(TARGET hydrogen PROPERTY CXX_STANDARD 14)
-add_custom_command(
-      TARGET hydrogen
-      POST_BUILD
-      COMMAND plutil -replace NSPrincipalClass -string NSApplication hydrogen.app/Contents/Info.plist
-      )
-add_custom_command(
-      TARGET hydrogen
-      POST_BUILD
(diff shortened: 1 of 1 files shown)


elpescado:
Mandatory screenshot:

Screenshot 2020-04-26 at 20 02 26

mauser:
Hi! 

currently the linux build fails because the `plutil` can't be found. I suppose that this custom command should be added only for OS X

elpescado:
Sorry about that! Should be fixed now.

mauser:
Perfect, merged! I wish i would have a newer mac with a high-res display to test it :smile:

cme:
1. I like the approach of using plutil to programmatically modify the Info.plist generated by cmake (cpack?). Much less pain than manually writing one (there's an old handwritten one that doesn't get used anymore). I'll probably try the same approach for adding associations for *.h2* files.
2. The high-DPI rendering is lovely, might make a motivating case for updating the icons & custom widgets to high dpi as well.

elpescado:
> * I like the approach of using plutil to programmatically modify the Info.plist generated by cmake (cpack?). Much less pain than manually writing one (there's an old handwritten one that doesn't get used anymore). I'll probably try the same approach for adding associations for _.h2_ files.

Thanks! The other approach was to copy Info.plist template from CMake, customise it and use it with [`MACOSX_BUNDLE_INFO_PLIST`](https://cmake.org/cmake/help/v3.7/prop_tgt/MACOSX_BUNDLE_INFO_PLIST.html) instead of default one. Using `plutil` seems to me much more elegant.

> * The high-DPI rendering is lovely, might make a motivating case for updating the icons & custom widgets to high dpi as well.

Thanks again! When I was working on HiDPI screen support I wanted to create high resolution assets but, unfortunately, I soon realised that it required too much work, so I decided to focus on most important things, notably font rendering. If someone is interested in picking that up, I might try to dig up my old branch and share my efforts.

Issue #702, opened 2019-03-18
Song: Time counter changes with BPM change


schattenform:
Hydrogen version *          : 0.9.7
Operating system + version  : Windows 10
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

The time counter in the top left corner starts counting from 0 when playing a track. But when the BPM changes during a song (lets say from 40bpm to 80bpm after 10 beats) the counter does not simply keep counting after the tenth beat but it changes and "pretends" that it has been 80 bpm from the beginning. So it might show 1min 30s when actually 2min have passed since starting at beat 1.

I would like the time counter to simply keep counting in real time independant from bpm changes. Is there an option for it? Is the current function on purpose? or is it an error?

Thank you.

mauser:
Hi! 

In my opinion this behavior is a bug,  label should show the correct time. Since this is not intended, there is no possibility to change that behavior.

theGreatWhiteShark:
But in order to get the real time, one has to take all the previous bpm markers into account. 

In case of the JACK driver these markers could very well be present in other JACK clients and Hydrogen does have no access to them at all. On the other hand, there is a variable called *usec*, a "free rolling" time, contained in the JACK transport information http://www.jackaudio.org/api/structjack__transport__info__t.html . I haven't used it yet but it might serve the purpose. But even if it does it won't reset when the pattern/song is looped and I don't see a way to implement this behaviour right now.

theGreatWhiteShark:
The problem in here is that transport in Hydrogen (for whatever reason) is *not* based on time but on ticks. When passing the tempo marker the tick size is rescaled and in some sense H2 internal time will be rescaled too.

With the current implementation one would either have to calculate the lengths of all previous patterns and combine them with all previously passed tempo markers or have to come up with a clever way to keep track of time.

theGreatWhiteShark:
Should be fixed with #855

PR #855, opened 2020-05-02
Calculated Timeline-aware elapsed time

theGreatWhiteShark:
Previously the time displayed in the time counter was solely based on the current position in frames.

But as soon as a BPM marker in the Timeline is passed, the size of a tick (the fundamental unit of position in Hydrogen) and effectively time itself will be rescaled. Therefore, relocating transport into a region of different tempo produces glitches in the time counter.

The pull requests fixes the this issue reported in #702 by introducing an overall time stored in the audio engine. Relocation does trigger a recalculation of its value, which takes all the tempo markers into account.

To have reproducible results regardless of the current transport position, I assumed the region before the first marker to have the same speed as set by the first marker itself.


theGreatWhiteShark:
@mauser do we have a policy concerning casts? During the process of revisiting the JACK driver stuff I had a bug (which was quite hard to track down) due to implicit casts in some equation. signed, unsigned, long, float. It's a lot of different things in here. I assume casting the arguments of a function call is a little bit over the top but it might be quite helpful to do so in calculations. What do you think?

PR #861, opened 2020-05-04
Proof-Of-Concept: Windows CI pipeline

elpescado:
Hi!

I've made Windows CI pipeline based on MSYS2 toolchain running on AppVeyor CI service.

# The good

* It works! It works automatically, builds every commit and checks every Pull Request, on four configurations (32 and 64 bit x Release and Debug)

Zrzut ekranu 2020-05-04 o 23 15 06

* Runs unit tests and reports progress in real time!

Zrzut ekranu 2020-05-04 o 23 05 48

* Makes installer .exe, runs it and verifies if some key files are present. Installer is stored on AppVeyor and can be fetched from there:

Zrzut ekranu 2020-05-04 o 23 07 45

[try it!](https://ci.appveyor.com/project/elpescado/hydrogen/builds/32599222/job/8gdiaetm40k23wkl/artifacts)
 
# The bad

* Building installer required some changes to CMake scripts, and that might break other things such as cross compile script. I've split this Pull Request into two commits, on only runs test and is fairly non-intrusive (`.appveyor.yml` configuration, use `pig-config` on Windows and some code in test framework to report test results to AppVeyor). The second commit makes installer, and is more intrusive.

# The ugly

* I don't have Windows, so I can't really test it;) It runs on AppVeyor build workers and on Wine, though. I NEED YOUR HELP!

* Test suite fails on 32-bit builds. I think that's problem with test suite itself, not Hydrogen. Currently test suite expects exported audio files to be bit-exact with reference files. But since we do a lot of floating point math to create audio files, some inaudible differences should be expected. I'll try to fix that later.

* Optimized builds crash, so I had to disable all optimisations `-O0`:(


CMakeLists.txt
```diff
@@ -408,10 +422,13 @@ SET(CPACK_PACKAGE_INSTALL_DIRECTORY "Hydrogen")
 IF(MINGW)
     #Set the other files that will be used in CPack
     SET(WIN64 "OFF" CACHE BOOL "Windows 64 Bit")
-	SET(CPACK_STRIP_FILES "${WINDOWS_DIR}/src/cli/h2cli.exe" "${WINDOWS_DIR}/src/gui/hydrogen.exe" "${WINDOWS_DIR}/src/player/h2player.exe")
+    if (CMAKE_BUILD_TYPE MATCHES Release)
+         SET(CPACK_STRIP_FILES TRUE)
+    else()
+         SET(CPACK_STRIP_FILES FALSE)
+    endif()
```
elpescado, line 429: On recent CMake (>2.6),  [`CPACK_STRIP_FILES`](https://cmake.org/cmake/help/v3.14/module/CPack.html#variable:CPACK_STRIP_FILES) is boolean.

CMakeLists.txt
```diff
@@ -408,10 +422,13 @@ SET(CPACK_PACKAGE_INSTALL_DIRECTORY "Hydrogen")
...
     #Program Files for Hydrogen
     SET(WINDOWS_DIR "windows")
-    INSTALL(FILES "${WINDOWS_DIR}/src/cli/h2cli.exe" "${WINDOWS_DIR}/src/core/libhydrogen-core-${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}.dll" "${WINDOWS_DIR}/src/gui/hydrogen.exe" "${WINDOWS_DIR}/src/player/h2player.exe" DESTINATION ./)
```
elpescado, line 414: That's not needed, `hydrogen.exe` and `libhydrogen-core-x.y.z.dll` are installed anyway by their targets, and this directive results in two copies of those files.

cmake/FindHelper.cmake
```diff
@@ -20,9 +20,9 @@ macro(FIND_HELPER prefix pkg_name header lib)
         set(${prefix}_FOUND TRUE)
     else()
         # use pkg-config if available to set find_path and find_library hints
-        if(NOT WIN32)
+        #if(NOT WIN32)
             FIND_PACKAGE(PkgConfig)
-        endif()
+        #endif()
```
elpescado, line 25: It doesn't hurt to try to use `pkg-config` on Windows, I think. It's not required, so if it's not found nothing happens. MSYS2 uses `pkg-config`.

mauser:
Wow, that's nice! I have no time currently to have a deeper look at the code, and i don't want to break the cross compile script before the 1.0 release :smile: But this is really promising. If this approach also permits us to compile Hydrogen with Jack and Ladspa Support on windows, we could also think about about using it as the primary approach to creating windows binaries (just like your MacOS build).  On the other hand, i don't think that i would want to drop the cross compile script completely, since it makes us indepedent from any particular CI service. 

As a sidenote: I noticed some weeks ago that travis is also slowly adding windows support.

elpescado:
Hi, I'm not sure if I broke the cross compile script, I wasn't able to test it, but I feel that might be the case, since I needed to modify some code used by that script. OTOH I think I rectified some bits. And there's some new stuff that might be used during cross compilation, like [`copy_thirdparty_dlls.py`](https://github.com/hydrogen-music/hydrogen/pull/861/files#diff-e723bf4c168af900b470989976f3f9ea) script that collects all DLL files that need to be bundled in installer and that is written in pure Python, so it should work on Linux host. Anyway, I also think that cross compilation script remains useful, and I'm sure that both approaches can coexist together.

Anyway, I don't think it's ready for prime time;) It needs to be tested, and needs some refinements. Esp. I'm concerned that release builds need to have optimisations disabled.

I chose AppVeyor basically by accident. I started working on this branch long time ago (I didn't have much time and hit many roadblocks in the meantime so development was extremely stretched in time), and back then, AppVeyor was, if I remember correctly, the only solution for CI on Windows platform. Recently I've noticed that some interesting alternatives have popped up:

* Travis seems to have some support for Windows
* Azure Pipelines (part of Azure DevOps), supports all three platforms and provides 10 concurrent jobs for Open Source projects
* GitHub Actions, supports all three platforms and provides 20 [!!!] concurrent jobs for Open Source projects

Anyway, I think AppVeyor is fine. What I like:

* Supports storing build artefacts (i.e. files created during build), so that there's no need to use external storage. Currently my pipeline stores Hydrogen installer, CMake logs and cache files and result of failed tests (eg. generated exported audio files that do not match reference data, so it's easier to debug).
* Can gather and display test results (which tests pass and which tests fail)
* There's RDP/SSH access to build worker, which is handy when investigating problems with build

There's one shortcoming:
* They provide only one concurrent job

E-Gwen:
Well, here I am.

I let you know that Ardour is compilated for Windows by MinGW ([http://mingw.org/])

And i just installed a software entirely compatible with Windows XP created by  NSIS ([https://sourceforge.net/projects/nsis/])(Which the latest 3.6.1 version is compatible With XP !)

Maybe it could help.

cme:
I have a couple of questions. When I've attempted to build Windows installers from master I've had to make the same changes to the CMakeLists.txt as here to get it to build. So this raises 2 questions:
  1. @mauser how are the release Windows packages built? How does that happen without running into cpack issues?
  2. Maybe we should apply the CMake portion of these changes, @elpescado, without waiting for CI to be stable and complete?

Cheers,
--
C.

mauser:
Hi! 

currently i'm building the windows binaries on my own system. I'm patching the `data/i18n/CMakeLists.txt` locally by hand, but other changes are made.  

I've started this evening by signing up to AppVeyor and testing the build process, this is why i introduced the branch `technical/windows-ci-pipeline` to this repo. After a small tweak, the build looks already good. I suppose that i can have a deeper look at the weekend, so i hope that we can merge that pull request then.

cme:
Awesome :)

mauser:
Current status: The created installers are working perfect on Windows 10 and the old MXE build process is still working. I've moved the files for the new build process to "windows/ci" to group them together and removed references to hard coded versions (like libhydrogen-1.0.0.dll).

elpescado:
Hi!

That sounds great!

Since this Pull Request was brought in context of building Hydrogen against previous versions of Qt, I thought I'd share some thoughts. The first thing to remember is that on Windows, mixing dlls and executables built with different compilers, or even different versions of the same compilers might or might not work (and in latter case, it manifests in mysterious crashes or cryptic dynamic linker messages).

Basically, to build hydrogen, one needs three things:
1. Compiler
2. Qt
3. a few dependencies (portaudio, sndfile, libarchve etc)

Fortunately, AppVeyor build virtual machines have [comprehensive list of software preinstalled](https://www.appveyor.com/docs/windows-images-software/), which is cool. There are some compilers ([MSVC](https://www.appveyor.com/docs/windows-images-software/#visual-studio-2008), [MinGW, MinGW-w64, MSYS2](https://www.appveyor.com/docs/windows-images-software/#mingw-msys-cygwin) etc.) There's several versions of [Qt](https://www.appveyor.com/docs/windows-images-software/#qt). So, my first approach was to use preinstalled MinGW, preinstalled Qt and somehow install rest of dependencies and be good to go.

There are [a few](https://www.appveyor.com/docs/windows-images-software/#tools) package managers available. Chocolatey focuses on end-user software (Adobe Reader et al), NuGet focuses on C# libraries. vcpkg has all required packages but [it seems to only support MSVC](https://stackoverflow.com/questions/63031513/vcpkg-with-mingw). Finally, MSYS2 ships with pacman ported from Arch. That's great.

Unfortunately, Qt built with MinGW `gcc` didn't play well with dlls built with MSYS2 `gcc`. I thought it should work, especially since most libraries used by Hydrogen are written in C which has less compatibility issues than C++. but it proved problematic.

So finally I install all dependencies, Qt included, from MSYS2 paceman and build Hydrogen using MSYS2 `gcc` and it works fine. The problem is that package managers like MSYS2 paceman and brew are targeted more at end users than folks using it to build their software, so they tend to provide only latest versions of software. In our [macOS pipeline I had to resort to a hack](https://github.com/hydrogen-music/hydrogen/pull/849/files) to get previous version of Qt installed by `brew`.

I'm no Windows expert, but maybe there's some way to use libraries installed by MSYS2 with `gcc` from MinGN and Qt compiled for MinGW. It's also possible to build dependencies from source, theres few of them and most of them are based on  cmake (but each of them has its own dependencies eg. sndfile required libvorbis, lib archive needs lib etc, so it might be unfeasible). Maybe binaries downloaded from official websites would work?

elpescado:
Recommended reading: [msys2 vs msys vs msysgit](https://gist.github.com/ReneNyffenegger/a8e9aa59166760c5550f993857ee437d)

elpescado:
> The problem is that package managers like MSYS2 pacman and brew are targeted more at end users than folks using it to build their software, so they tend to provide only latest versions of software.

Maybe we could work that around? See: [How to obtain older versions of packages using MSYS2?](https://stackoverflow.com/questions/33969803/how-to-obtain-older-versions-of-packages-using-msys2):

>No, MSYS2 does not directly support installing old versions of packages. Here are some things you can do though:
> 
> - Look at http://repo.msys2.org/ to see if the version you are looking for happens to still be on the server. It will be deleted eventually but you might get lucky. You can download it and then install it with pacman -U filename.tar.xz

[Here](http://repo.msys2.org/mingw/x86_64/) I can see several Qt packages, back to 5.5:

```
mingw-w64-x86_64-qt5-5.5.0-1-any.pkg.tar.xz        02-Jul-2015 14:32           648836000
mingw-w64-x86_64-qt5-5.5.0-1-any.pkg.tar.xz.sig    02-Jul-2015 14:32                  96
mingw-w64-x86_64-qt5-5.5.1-1-any.pkg.tar.xz        26-Oct-2015 11:30           687320740
mingw-w64-x86_64-qt5-5.5.1-1-any.pkg.tar.xz.sig    26-Oct-2015 11:30                  96
mingw-w64-x86_64-qt5-5.5.1-2-any.pkg.tar.xz        06-Nov-2015 16:17           683447632
mingw-w64-x86_64-qt5-5.5.1-2-any.pkg.tar.xz.sig    09-Nov-2015 05:20                  96
mingw-w64-x86_64-qt5-5.5.1-4-any.pkg.tar.xz        13-Jan-2016 10:34           683333176
mingw-w64-x86_64-qt5-5.5.1-4-any.pkg.tar.xz.sig    13-Jan-2016 10:45                  96
mingw-w64-x86_64-qt5-5.6.0-1-any.pkg.tar.xz        22-Mar-2016 17:52           723848688
mingw-w64-x86_64-qt5-5.6.0-1-any.pkg.tar.xz.sig    22-Mar-2016 17:53                  96
mingw-w64-x86_64-qt5-5.6.0-2-any.pkg.tar.xz        08-Apr-2016 16:35           741546592
mingw-w64-x86_64-qt5-5.6.0-2-any.pkg.tar.xz.sig    08-Apr-2016 19:22                  96
mingw-w64-x86_64-qt5-5.6.1-1-any.pkg.tar.xz        09-Jun-2016 16:54           755829344
mingw-w64-x86_64-qt5-5.6.1-1-any.pkg.tar.xz.sig    10-Jun-2016 05:21                  96
mingw-w64-x86_64-qt5-5.6.1-2-any.pkg.tar.xz        19-Jul-2016 16:33           767311872
mingw-w64-x86_64-qt5-5.6.1-2-any.pkg.tar.xz.sig    19-Jul-2016 16:33                  96
mingw-w64-x86_64-qt5-5.6.1-3-any.pkg.tar.xz        20-Sep-2016 17:15           772072444
mingw-w64-x86_64-qt5-5.6.1-3-any.pkg.tar.xz.sig    20-Sep-2016 17:23                  96
mingw-w64-x86_64-qt5-5.6.2-4-any.pkg.tar.xz        13-Oct-2016 00:27           775941012
mingw-w64-x86_64-qt5-5.6.2-4-any.pkg.tar.xz.sig    13-Oct-2016 03:14                  96
mingw-w64-x86_64-qt5-5.8.0-1-any.pkg.tar.xz        30-Jan-2017 17:25           851798488
mingw-w64-x86_64-qt5-5.8.0-1-any.pkg.tar.xz.sig    30-Jan-2017 18:16                  96
mingw-w64-x86_64-qt5-5.8.0-2-any.pkg.tar.xz        07-Feb-2017 09:46           824922236
mingw-w64-x86_64-qt5-5.8.0-2-any.pkg.tar.xz.sig    07-Feb-2017 09:46                  96
mingw-w64-x86_64-qt5-5.8.0-3-any.pkg.tar.xz        25-Feb-2017 16:19           824882696
mingw-w64-x86_64-qt5-5.8.0-3-any.pkg.tar.xz.sig    25-Feb-2017 16:20                  96
mingw-w64-x86_64-qt5-5.9.0-1-any.pkg.tar.xz        08-Jun-2017 03:29           938991352
mingw-w64-x86_64-qt5-5.9.0-1-any.pkg.tar.xz.sig    08-Jun-2017 03:30                  96
mingw-w64-x86_64-qt5-5.9.1-1-any.pkg.tar.xz        01-Jul-2017 16:17          1052771612
mingw-w64-x86_64-qt5-5.9.1-1-any.pkg.tar.xz.sig    01-Jul-2017 16:21                  96
mingw-w64-x86_64-qt5-5.9.2-1-any.pkg.tar.xz        09-Oct-2017 07:37          1023626436
mingw-w64-x86_64-qt5-5.9.2-1-any.pkg.tar.xz.sig    09-Oct-2017 07:39                  96
mingw-w64-x86_64-qt5-5.9.2-2-any.pkg.tar.xz        13-Oct-2017 08:33          1023640780
mingw-w64-x86_64-qt5-5.9.2-2-any.pkg.tar.xz.sig    13-Oct-2017 08:41                  96
```

E-Gwen:
Hi,

Here you can find examples for compiling 32 Bit binaries with 64 bit compiler :

https://github.com/L1L1/cardpeek/commit/1e3bef5e9ec077c09035f4234eb7fa9abcad023b

I hope it will be usefull.

elpescado:
By the way, what bothers me most, and what I think is the most prominent showstopper is this workaround:

```cmake
OPTION(FIXME_DISABLE_OPTIMIZATIONS "Fix broken builds by turning off optimizations" OFF)
(...)
IF(FIXME_DISABLE_OPTIMIZATIONS)
    string(REGEX REPLACE "-O[0123s]" "-O0" CMAKE_CXX_FLAGS                "${CMAKE_CXX_FLAGS}")
    string(REGEX REPLACE "-O[0123s]" "-O0" CMAKE_CXX_FLAGS_RELEASE        "${CMAKE_CXX_FLAGS_RELEASE}")
    string(REGEX REPLACE "-O[0123s]" "-O0" CMAKE_CXX_FLAGS_DEBUG          "${CMAKE_CXX_FLAGS_DEBUG}")
    string(REGEX REPLACE "-O[0123s]" "-O0" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
ENDIF()
```
[source](https://github.com/hydrogen-music/hydrogen/pull/861/files#diff-af3b638bc2a3e6c650974192a53c7291R130-R135)

Basically, builds with default optimisation level (`-O2`) crash right after startup for mysterious reasons (eg. backtrace point to `std::string` internals).

I didn't have time to investigate it properly so I introduced this workaround to make binaries working. Anyway, there have been some pull requests merged for valgrind bugs that might me related to this issue, so there's slight chance that this workaround is no longer necessary.

mauser:
Hi! 

The branch is now merged, it feels just "good enough" for the master branch :smile:  I've added support for portmidi and ladspa. Jack and liblo packages are currently missing from the msys2 repo, but this doesn't bother me much for those "nightly" builds. And at least liblo could be packaged quite easily, if we really want to do the effort.

@elpescado : I'm going to play around with older QT versions. Thanks for all the information about the build environments, your comments were very helpful!

elpescado:
Great!

cme:
Fantastic!

PR #862, opened 2020-05-04
Sanitizers

elpescado:
Hi!

I've expanded our Travis CI pipeline to run various sanitisers as second stage (i.e. only after regular build and unit tests succeed).

# The good

* It runs Address Sanitizer, Memory Sanitizer, Leak Sanitizer and Undefined Behaviour Sanitizer on Travis CI as part of pipeline, allowing us to catch more bugs, on every commit to master, and on every Pull Request

* It looks pretty:

Zrzut ekranu 2020-05-04 o 23 15 45

* I've also added Ubuntu 18.04 Bionic Beaver to build/unit tests job so that we test on more recent Linux distro.

* It's not intrusive, I've only added some jobs to `.travis.yml` so there are no changes in "production" code, so it can be safely merged.

# The bad

* I've left some commented out code - I've tried running those sanitisers on three platforms - Linux/GCC, Mac/LLVM and Linux/LLVM. I've found no significant differences between those platforms, so I left only Linux/GCC (except Memory Sanitizer that only ran successfully on Linux/LLVM). I left the rest commented out for reference only, and I think those jobs could be removed.

# The ugly

* Address Sanitizer, Memory Sanitizer, Leak Sanitizer fail. I've marked them with `allow_failures` so that they do not fail whole pipeline. I think there are some false positives, I'll try to sort that out. Once those jobs are stable, `allow_failures` feature can be turned off.

* Currently, pipeline is set up weird. Mac binary packages are built and uploaded on step one, and then sanitisers are run. It would make more sense to build and upload packages after sanitisers, but it wouldn't be easy to build pipeline like that.

* Sanitisers are run on test suite, and test suite tests only core, therefore sanitizers also exercise only core. GUI is not checked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,83 @@ env:
   global:
   - MAKEFLAGS="-j 2"
 jobs:
+  allow_failures:
+  - name: Address Sanitizer
+  - name: Leak Sanitizer
+  - name: Memory Sanitizer
+
   include:
-  - os: linux
+  - stage: Build
+    name: "Ubuntu 16.04 Xenial"
+    os: linux
     dist: xenial
-  - os: osx
+  - name: "Ubuntu 18.04 Bionic"
+    os: linux
+    dist: bionic
+  - name: "macOS"
+    os: osx
     env:
     - MACOSX_DEPLOYMENT_TARGET="10.12"
-  - os: osx
+  - name: "macOS (legacy)"
+    os: osx
     env:
     - MACOSX_DEPLOYMENT_TARGET="10.10"
     - HOMEBREW_NO_AUTO_UPDATE=1
     - PKG_SUFFIX="-legacy"
     osx_image: xcode9
+
+  - stage: Sanitizers
+    name: Address Sanitizer
+    os: linux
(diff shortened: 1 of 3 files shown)


mauser:
Hi! 

Great! How should we proceed? Maybe it would be a good idea to have only the working sanitizers active on the master branch, otherwise it would always show fails on pull request, which might confuse people who do not know about the "old" problems.

theGreatWhiteShark:
First of all: great work @elpescado !

> We could then enable all sanitizers on a different branch and work on the problems there.

I would also vote for fixing those sometime soon on a different branch to not interfere with the 1.0 release.

mauser:
Hi! 
 
ok, i'm going to prepare a new branch then with all sanitizers activated.

elpescado:
Hi, currently sanitisers that fail are marked as `allow_failures` in order not to fail any build, so it should not prevent pull requests from being merged etc. I'm not sure if we need separate branch - pull requests could target `master` and once individual sanitizer jobs are stabilised (green), we can remove `allow_failures` flag on them.

PR #869, opened 2020-05-11
Fix issues found by valgrind

cme:
Fixes for a few issues found by valgrind's memcheck:

  - Potential crash: using Hydrogen::setSong() to set a new song while
    an existing song is playing will delete the old song, including
    its samples, before stopping the player.
  - Mismatched new[] / delete (should be delete[]) in samples buffers
  - potentially undefined execution control flow in libsndfile under
    sf_open: the API spec says that SF_INFO.format should be
    initialised to 0 before calling sf_open

### DIFF
--- a/src/core/src/basics/sample.cpp
+++ b/src/core/src/basics/sample.cpp
@@ -169,7 +169,8 @@ bool Sample::load()
 {
 	// Will contain a bunch of metadata about the loaded sample.
 	SF_INFO sound_info;
-	
+	sound_info.format = 0;
+
 	// Opens file in read-only mode.
 	SNDFILE* file = sf_open( __filepath.toLocal8Bit(), SFM_READ, &sound_info );
 	if ( !file ) {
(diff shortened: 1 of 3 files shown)


src/core/src/basics/sample.cpp
```diff
@@ -169,7 +169,8 @@ bool Sample::load()
 {
 	// Will contain a bunch of metadata about the loaded sample.
 	SF_INFO sound_info;
-	
+	sound_info.format = 0;
```
elpescado, line 172: Would it hurt to zero whole struct, just in case? `SF_INFO sound_info = {0}`?

elpescado:
That's nice!

src/core/src/basics/sample.cpp
cme: I initially went with just 'format' because that makes it explicit that it's a specific requirement of the API (a very odd requirement, at that!) but initialising the whole thing does look cleaner and read better.

mauser:
Thanks again, merged!

PR #870, opened 2020-05-12
Test backtrace

cme:
*** Temporary ***

This is a branch which dumps a stacktrace on segv in the test driver. This is to try and find the cause of the segvs in the Travis testing.


cme:
>60 tests run with no failures.

PR #872, opened 2020-05-15
Fix missing cmake includes

cme:
Removes the need to create '/opt/local' symlink to build on MacOS systems which use Homebrew rather than MacPorts. This *should* make the sources and cmake files package-manager-agnostic.


cme:
I'd really appreciate if someone could take a look and merge at least the cmake changes, as it means I can't currently *directly* build master or any of the branches I've created for pull requests, as I can't symlink /opt/local on my machines.

cme:
Actually I've just realised I can work around this with a different top level cmake file to keep a clean repository, so I retract my request for an imminent pull, but will leave this here for a pull at leisure.  Thanks :)

mauser:
Hi! 

out of interest: why is the inclusion the of the jack/libsndfile Headers in the UI/tests/players makefile needed? Those components should not need them, in my opinion. Did you experience build problems without that include?

cme:
> out of interest: why is the inclusion the of the jack/libsndfile Headers in the UI/tests/players makefile needed? Those components should not need them, in my opinion. Did you experience build problems without that include?

Yes, the build fails because sndfile.h is included from sample.h as it defines constants for the 'format' argument of Sample::write(). Perhaps Sample should abstract that as it does some other things, but it seems fairly harmless.

elpescado:
I can't build `master` branch on my work MacBook (running 10.14 Mojave, build fails due to missing `sndfile.h`), whereas this branch builds fine.

BTW. I have recently read a bit about "Modern CMake", and it seems thot nowadays it's discouraged to use `INCLUDE_DIRECTORIES` directly, instead use `target_include_directories`. That way, include directories are passed transitively to dependent targets, i.e. dirs set for `core` would be automagically applied for `gui`, `cli` and `h2player`, at least that's how I understand it. And certainly I don't feel that I understand CMake enough to make any revolutionary changes.

[*It's Time To Do CMake Right*, Pablo Arias](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/)

PR #875, opened 2020-05-17
Fix #874 - improve scrolling to newly selected pattern

### LINKED ISSUE #874: Pattern window scrolling unexpectedly when clicking on patterns in middle of area
Hydrogen version *          : 1.0.0-beta2
Operating system + version  : Ubuntu Studio 20.04
Audio driver + version      :  Jack 1.9.12 (Not sure if this is what's being asked for here, but it's a GUI thing so shouldn't be relevant...?)

When I select a pattern, the vertical scroll immediately repositions the scrollbar such that the pattern I've selected is at the top of the area. E.g. with the default 10 patterns, if I am editing Pattern 1, and click on Pattern 4, it will immediately scroll down to put Pattern 4 at the top (if it can scroll down that far, otherwise it will just scroll down to the bottom of the range). It often does this so quickly that I have not yet let go of the mouse button, causing the pattern I've clicked on to be repositioned. The same thing happens if I have a Pattern further down selected and click on one nearer the top, but scrolling upwards instead.

This was not happening before I upgraded to 20.04, although I can't remember which version I was running before on 19.10.
Let me know if there's anything useful I've neglected to mention.

Thanks!

cme:
Fix for issue #874 -- improve scrolling the song editor's pattern list to avoid scrolling on selecting a new pattern when it's already visible.

### DIFF
--- a/src/gui/src/SongEditor/SongEditorPanel.cpp
+++ b/src/gui/src/SongEditor/SongEditorPanel.cpp
@@ -666,6 +666,7 @@ void SongEditorPanel::restoreGroupVector( QString filename )
 
 void SongEditorPanel::resyncExternalScrollBar()
 {
+	int nGridHeight = m_pPatternList->getGridHeight();
 	m_pHScrollBar->setMinimum( m_pEditorScrollView->horizontalScrollBar()->minimum() );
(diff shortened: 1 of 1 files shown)


mauser:
Thanks a lot , that behavior also annoyed me lately :)

Issue #876, opened 2020-05-18
Song Editor: crash when undoing move of patterns off to left of song


cme:
Hydrogen version *          : master
Operating system + version  : Mac, Linux
Audio driver + version      : n/a

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

To repeat:

  1. In song editor, set pattern 1 active for 4 repeats
  2. Select mode
  3. Select all 4 cells
  4. Drag from (3,0) to (0,0). The resulting song display has 1 active cell (0,0) and loses 3 cells off the beginning of the song
  5. Hi Ctrl+Z / Command+Z to undo

Hydrogen spins out and locks up / crashes.

I'll investigate ad probably fix later.

cme:
Submitted pull request #877 which should fix this.

cme:
Tested on master after #877 merged. Closing.

PR #877, opened 2020-05-18
Fix bug #876: crash undoing move patterns left

cme:
Fix for bug #876 . movingCells can contain column indices less than 0, so guard against this when selecting patterns to delete in response to an undo.


mauser:
Thanks a lot, merged!

Issue #857, opened 2020-05-03
Song exported from hydrogen has glitches / does not maintain a consistent tempo


lukesanantonio:
Hydrogen version *          : `Hydrogen 1.0.0-'dc269b98'` (just built from `master` earlier today)
Operating system + version  : Ubuntu 20.04
Audio driver + version      : Jack 1.9.12, but also tried pulseaudio, alsa drivers

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Hello! Just wanted to say thank you for an amazing piece of software!!

I'm having an issue where a song exported from hydrogen does not maintain a consistent tempo, I've been having this issue since I upgraded to 1.0.0-beta2, but the issue still seems present in master (which I compiled earlier today).

Here's how I reproduce it:

Open a new hydrogen project, use bpm 120, add beats on every count, 1,2,3,4, etc. Set this pattern to go for 32 bars in the song. Export song as 44800 Hz wav, linear interpolation. (I've also tried ogg, flac, etc, but the issue persists)

Open audacity, import the hydrogen WAV, hit generate -> rhythm track. I'm using tempo 120 bpm, 4 beats per bar, 32 bars, which sould match the recording exactly.

Okay, so then I play the two files back. They start perfectly in sync, but every 10 bars or so the hydrogen-exported track has a strange, audible glitch which puts a slight delay in the track, resulting in out of sync beats!

I've been reading through some other issues, but can't find any that only apply to exporting the song. And I've used a temporary workaround where I click play in hydrogen at just the right time to sync with LMMS. When I'm playing the pattern directly inside hydrogen it never gets out of sync.

I also noticed these messages in the console when I run hydrogen with `-Vdebug`, they happen for every filetype (ogg, flac, wav) that I've tried exporting!
```
(I) ::void* H2Core::diskWriterDriver_thread(void*) DiskWriterDriver thread start
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 1024, new size = 768
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 768, new size = 1024
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 1024, new size = 768
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 768, new size = 1024
[...]

PR #878, opened 2020-05-23
Add timed waits for audioengine lock ( part fix for #857 )

cme:
See also: https://github.com/cme/hydrogen/blob/timed_locks_profiling/profiling/report.md

This avoids cases where playback might be disrupted due to the GUI
grabbing the lock to update song position or edit patterns / song
structure while playing. Such locks are very short-lived, much
shorter than is typically needed to process a buffer of song data,
so waiting until the lock can be acquired is a very short delay.

To avoid excessive waits which would cause a buffer to be dropped
anyway or otherwise delivered to the sound driver too late, the
available slack time is calculated and lock waits are limited to
that slack time.

This change also changes the audioengine lock from a pthreads mutex
to a C++11 std::timed_mutex: this is because pthreads mutexes support
timed locking on Linux, FreeBSD and Windows, but bizarrely *not*
on macOS, presumably because it would make the mutex structure
larger. std::timed_mutex already implements this on all platforms,
so use it in preference.

### DIFF
--- a/src/core/include/hydrogen/audio_engine.h
+++ b/src/core/include/hydrogen/audio_engine.h
@@ -28,10 +28,10 @@
 #include <hydrogen/sampler/Sampler.h>
 #include <hydrogen/synth/Synth.h>
 
-#include <pthread.h>
 #include <string>
 #include <cassert>
-
+#include <mutex>
+#include <chrono>
 
 /** \def RIGHT_HERE
  * Macro intended to be used for the logging of the locking of the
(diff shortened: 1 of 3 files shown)


mauser:
So, finally this pull request made its way into the master branch. Thanks for the fix and the profiling!

cme:
This reminds me that I meant to take a look at locking in general since the monolithic lock is a fairly coarse level, and also I'm convinced there are parts of the GUI that should be locking but don't.

PR #880, opened 2020-05-24
Fix song export glitches 

mauser:
See issue #857 .


mauser:
> There's an issue with the Jack driver here -- the Jack driver uses audioEngine_process() directly as its processing callback, and Jack will interpret a return value of 2 as meaning to kill the thread. So if using Jack, and a lock from the GUI clashes with audioEngine_process(), it will kill the playback thread.
> 
> The Jack driver needs a wrapper callback to map this return value to 0 to continue playback.

Thanks for the hint! For the moment, i've chose to check to return the special value only if the process method is called from the diskwriter driver. This has already been done some lines later for the return value "1"..

Issue #881, opened 2020-05-24
ALSA sound driver glitches and underruns


cme:
Hydrogen version *          : master
Operating system + version  : Linux (Raspbian ARM and Ubuntu 20.04 x86)
Audio driver + version      :  ALSA

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

I wasn't entirely sure I was hearing this at first since I have no decent speakers connected to my Raspberry Pi and my ears aren't what they used to be (because drums :D ) but fixing the issue improves matters a lot.

The ALSA driver miscalculates the size of buffers, probably due to different drivers using different terminology for the same things. ALSA describes a unified buffer containing multiple periods, where other sound systems might treat each of these as potentially separate buffers.

The ALSA driver uses the user-set preference for nBufferSize to set the size of the buffer, dividing it by 2 to get the number of frames in the *period*. 

In the callback thread, nFrames is set from nBufferSize, and attempts to write that many frames to the sound system. This writes over *both* periods in the buffer, which will include data being read by the hardware, leading to audible artefacts that are more prominent in busy patterns.

PR #882, opened 2020-05-24
Fix invalid buffer size crash #879 and glitches #881

### LINKED ISSUE #879: Hydrogen crashes on startup / restart output with invalid buffer size and ALSA driver
Hydrogen version *          : master
Operating system + version  : Linux (raspbian)
Audio driver + version      : ALSA

Using the ALSA driver, an invalid buffer size in saved preferences will crash Hydrogen. This will happen when the user restarts the sound system manually, or when starting Hydrogen. Since the crash will happen on startup from then on, the user doesn't have the opportunity to fix the problem.

To reproduce:
  - Start Hydrogen from a fresh install
  - Options -> Preferences -> Audio System
  - Set drop-down driver selector to "ALSA"
  - Buffer size -> set to 128 (or some other random value to displease your system)
  - Hit OK. 
    * User interaction problem here: the user should expect that settings should be applied. The other alternative, "Restart output" won't see these changes yet.
  - Options -> Preferences -> Audio System -> Restart output
    * SEGV
  - Restart hydrogen
    * SEGV

cme:
Fixes #879, and also #881 (in one pull request since they're closely related and would otherwise conflict).

### DIFF
--- a/src/core/src/IO/alsa_audio_driver.cpp
+++ b/src/core/src/IO/alsa_audio_driver.cpp
@@ -224,6 +224,7 @@ int AlsaAudioDriver::connect()
 
 	if ( ( err = snd_pcm_hw_params_set_period_size( m_pPlayback_handle, hw_params, period_size, 0 ) ) < 0 ) {
 		ERRORLOG( QString( "error in snd_pcm_hw_params_set_period_size: %1" ).arg( QString::fromLocal8Bit(snd_strerror(err)) ) );
+		return 1;
 	}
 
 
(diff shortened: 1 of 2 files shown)


theGreatWhiteShark:
+1 Looks good.

I didn't hear any glitches in master (but that may be hardware/pattern dependent). But most importantly everything else still works as expected.

cme:
> +1 Looks good.
> 
> I didn't hear any glitches in master (but that may be hardware/pattern dependent). But most importantly everything else still works as expected.

Seems to be dependent on other system load. It was only noticable on my i7 machine under load; but noticable on the Raspberry Pi without anything else happening on the machine.

mauser:
Thanks for the fix! 

I've noticed such glitches when working with an older laptop, but never payed attention to this (instead, i switched to jackd in this situation).

Issue #883, opened 2020-05-25
Unintuitive behavior of audio driver restart via Preferences > Audio System 


theGreatWhiteShark:
Right now all changes applied in the Options > Preferences windows will be saved in the `Preferences` singleton the moment the user hits the 'OK' button.

As a consequence all changes applied to the audio driver settings will _not_ be considered when pressing the 'Restart output' button until the whole dialog gets closed and opened again.

I don't know much about Qt but there is most probably a way to make the 'Restart output' button evaluating the provided forms.

Drawback (also in the current implementation): it's easily possible to have a set of parameter not reflecting the one used to start the driver. Maybe there could a 'Restart output and close' button instead, which won't close the dialog if the restart of the audio driver fails.

cme:
The layout is also confusing, with the drop-down on the left and the 'Restart' button to its left. Is there a *current* use-case for the "Restart..." independently of applying new settings?

The "OK" button currently asks the user to restart if necessary. The behaviour you suggest for "Restart output and close" is pretty close to what the "OK" button already does if there has been a setting change that needs a restart, and the not-closing-the-dialog behaviour is close to a regular "Apply" button.

Would it perhaps be fairly close to user expectation to:
  - Get rid of the 'Restart' button
  - Add an 'Apply' button which behaves the same as the OK button but doesn't close the dialog. 

So the user's tweak-test-retweak cycle would look like:
  - set audio settings
  - hit 'apply'
  - hit "OK" to restart
  - Go back to fix settings if that fails

There are perhaps a few more steps in that than necessary, but they're at least fairly intuitive and well-guided and well-established in other GUIs. It also makes it clear when settings are saved.

If allowing the user to restart the sound system without making any settings changes, perhaps a button at the bottom with behaviour closer to "cancel and restart" might be appropriate? Or a "Force restart" checkbox.

The GUI code looks straightforward enough to refactor.

cme:
(Also, this has annoyed me very recently. And I guess I'm volunteering to take it on... :D )

theGreatWhiteShark:
> Is there a current use-case for the "Restart..." independently of applying new settings?

Jupp. With the JACK server Hydrogen does exclusively use the sample rate and buffer size provided by the server. If you want to tweak these settings, you have to do it in the server and restart the driver in H2 later on.

> if there has been a setting change that needs a restart

[...]

PR #893, opened 2020-05-29
Apply driver pref changes before restart output

cme:
Implemented proposed changes to Preferences dialog box for audio driver options to address #883, applying changes before restarting driver.


theGreatWhiteShark:
I like it.

But two things confuse me (which probably already have been present beforehand).
1. The actual sample rate and buffer size of the Jack server is not reflected by the displayed numbers. (The latter we could also display grey to indicate that they represent something static)
2. The choice of the Auto driver is not reflected after restarting.

But we could of course merge this PR and handle/discuss the things above in a separate one.

theGreatWhiteShark:
@cme can we merge this?

cme:
> @cme can we merge this?

Yep.

Today's lesson: resolving conflicts in `.ui` files is not my favourite thing ever.

theGreatWhiteShark (approved):

theGreatWhiteShark:
> Today's lesson: resolving conflicts in .ui files is not my favourite thing ever.

Yeah. When I first saw QtDesigner I thought tweaking the GUI will be so much easier. But the devil is in the detail and it is not nearly as simple as I thought it to be.

PR #898, opened 2020-06-16
patch stats.py with Python 2to3

sten0:
Hi,

This is just a simple fix I had to make for the Debian package, because py2 is EOL.

Regards,
Nicholas

### DIFF
--- a/data/i18n/stats.py
+++ b/data/i18n/stats.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 ## Python script checking the coverage of the translation files.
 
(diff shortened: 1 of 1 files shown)


data/i18n/stats.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
cme, line 1: I guess the change to interpreter is just to test, but should probably be kept as just ```python```, since with the below change the script is valid for 2 or 3, and keeping the interpreter as ```python``` will allow it to continue working on any other host that has only 2?

elpescado:
Hi,

This tool was actually ported to Python 3 in f57b76040cd5ada2fa494a73ef1f9d4fcec1afcd. Then, shebang line was changed back to `#!/usr/bin/env python` in 3b449146ba6c306f0f54dd04aedfb6a2ca021b55.

cme:
> For reference, this tool was actually ported to Python 3 in [f57b760](https://github.com/hydrogen-music/hydrogen/commit/f57b76040cd5ada2fa494a73ef1f9d4fcec1afcd). Then, shebang line was changed back to `#!/usr/bin/env python` in [3b44914](https://github.com/hydrogen-music/hydrogen/commit/3b449146ba6c306f0f54dd04aedfb6a2ca021b55).

Now that I look at it again, the only other change in this one is unneeded as it's already valid python3, so I don't understand why any of this change would be found to be necessary.

@sten0 - what's the rationale? The only thing I can think of is that the build environment somehow had python3 accessible only as 'python3' and not as 'python'?

mauser:
Closed - no reaction from pull request initiator. In addition, i can't see why we should change this.

Issue #899, opened 2020-06-18
Hydrogen will not start (windows 7 runtime error)


limelight022:
Ive tried every available version of Hydrogen on the site, and every single time i try to launch it i get "the application has requested the runtime to terminate" error. 

Ive tried all compatibility modes, running as administrator, restarting my pc, updating Windows and all C++ libraries and still the same. 

Here is the log:

Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	hydrogen.exe
  Application Version:	0.0.0.0
  Application Timestamp:	5ede8547
  Fault Module Name:	libportaudio-2.dll
  Fault Module Version:	0.0.0.0
  Fault Module Timestamp:	5c8ec46f
  Exception Code:	40000015
  Exception Offset:	000000000001f9f6
  OS Version:	6.1.7601.2.1.0.256.48
  Locale ID:	1033
  Additional Information 1:	da0a
  Additional Information 2:	da0a2778e5a7a401d00cd3c644a55f56
  Additional Information 3:	8157
  Additional Information 4:	8157d7416025a1d4e0acb01793d63a89

cme:
This looks like the same issue as #879 , which looks like it's an issue in PortAudio probably not coping with some devices.

What sort of sound hardware and drivers do you have in your system?

limelight022:
Standard motherboard audio, RealTek High Definition Audio, Driver version 6.0.8924.1

cme:
> Standard motherboard audio, RealTek High Definition Audio, Driver version 6.0.8924.1

Wow, that's surprising. It's almost impossible to debug something like this without reproducing it. Would you mind helping out by running an instrumented version of Hydrogen (which I might be able to whip up over the weekend) that will output some more debug info to try and work out what's going on and how to work around it?

Thanks!

limelight022:
Sure, looking forward to it. Thanks!

cme:
Instrumented binary installs are at:

https://drive.google.com/drive/folders/1AEEldmPZD7-aM0u4iVOG3e1GbbUc_cy-?usp=sharing

You can capture the debug output with:

```
"C:\Program Files\Hydrogen\hydrogen.exe" -V Debug > C:\debug.txt"
```

from a command prompt.

Thanks!

limelight022:
I got the same runtime error. Im a little confused about capturing the debug output commands though. I copy/paste each one without quotations and heres what it says:

[...]

Issue #873, opened 2020-05-16
Hydrogen crashing on startup.


Zday89:
Hydrogen version *     : 0.9.7 - 1.0.0-Beta2 
Operating system + version  : Windows 10 64 bit
Audio driver + version      : NA

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Hydrogen previously worked great, but seemingly at random it stopped. I get the splash screen and then it crashes. This happens on versions 0.9.7 through 1.0.0-beta2. I've checked the Windows event viewer and it says libportaudio-2.dll is the faulting module. I temporarily fixed this issue and got it started by downloading PortAudio.dll from the portaudio website and replaced libportaudio-2.dll with that file. That allowed me to open and use Hydrogen, but only on version 0.9.7. This did not work with 1.0.0 at all. It did, however, make some features buggy. For example, I tried to save an audio library and the program crashed and now it will not open anymore, this time with the faulting module "libhydrogen-core-0.9.7.dll." 

Now I can't use the program at all. Even after uninstalling completely using Revo Uninstaller to clear any registry values and leftover files. Hydrogen 0.9.6 alpha1 appears to work with no modification after installing. 

Here's the windows event log for a crash using a clean install of version 1.0.0-beta2:
```
Faulting application name: hydrogen.exe, version: 0.0.0.0, time stamp: 0x5e7a7416
Faulting module name: libportaudio-2.dll, version: 0.0.0.0, time stamp: 0x5c8ec46f
Exception code: 0x40000015
Fault offset: 0x000000000001f9f6
Faulting process id: 0x31c4
Faulting application start time: 0x01d62bac89b815c3
Faulting application path: C:\Program Files\Hydrogen\hydrogen.exe
Faulting module path: C:\Program Files\Hydrogen\libportaudio-2.dll
Report Id: 9fe0553e-b5a7-4b06-a51d-da0992677e65
Faulting package full name: 
Faulting package-relative application ID: 
```

cme:
Are you able to try removing or renaming the file `.hydrogen\hydrogen.conf` in your user home directory? If the config file ends up containing invalid audio configuration, it's possible that would cause Hydrogen to crash on startup (and not give you the opportunity to try different settings).

Zday89:
> 
> 
[...]

PR #901, opened 2020-06-23
Disable WDM-KS in portaudio build

cme:
Disable WDM-KS support in PortAudio build, as it's broken. This resolves (sorta) #899 and #873 .


mauser:
Perfect! I'm reporting back when my local mxe installation is ready again.

Issue #902, opened 2020-06-26
No user feedback when loading song fails to load samples


cme:
Hydrogen version *          : 1.0.0-rc
Operating system + version  : all
Audio driver + version      : all

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

When loading a song fails to load some samples (as will happen when loading an old song using the old default drumkit on a newer Hydrogen), there's no indication to the user (other than to stderr, which is generally discarded or ignored, certainly on Mac and Windows) that anything is wrong, other than silence. Additionally there's suggestion on how to fix it.

The topic has come up coincidentally in other issues here, but doesn't have an Issue of its own. I'm creating one now because I just had a conversation with a long-time Hydrogen user who has lots of old files and just installed 1.0.0-rc and thought it was broken because their old songs just didn't play anything with no indication of trouble.

I think either 
  1. a simple QMessage dialog when this happens, suggesting that the fix might be to load the default drum kit, possibly with a button to just load the default kit.
  2. using the "infobar" which usually warns about potentially non-GM kits to display the message, and again possibly provide a button to load the default kit.

cme:
Closing, #903 has been merged to master.

PR #903, opened 2020-06-28
Warn user of missing samples on song load

cme:
Make the GUI warn users of missing samples when loading songs
  * Refactor HydrogenApp's use of InfoBar to allow multiple InfoBars
  * Present an InfoBar on missing samples, with a button offering to open drumkits

This addresses #902.


mauser:
Hi Colin! 

what about storing the "missing samples" information for an instrument (iso. storing it per song)? It would feel more "natural" (?) to me to have an indicator which tells me "Hey, there is something wrong with that Kick drum instrument here.." . The user could then  decide himself if this is a problem or not. Maybe there could also be a small yellow warning sign (those triangular things with an exclamation mark..) at the beginning of the instrument name in the pattern editor. Maybe a missing sample is ok if you know that the instrument is not important (maybe the third bottom snare sample your friend added..), but irt a t could be fatal if this is your primary kick drum or whatever..

Just a thought to start a discussion, i'm not wanting to say that this has to implemented exactly in this way to get accepted :)

cme:
I like this idea. I think the infobar notification ought to be kept, though, as it's useful in what might be the most common case.

Also, a tooltip with list of missing samples on hovering over the indicator might be useful.

cme:
Updated to flag individual instruments as degraded, with little warning triangle, with tooltip and an error box. 

image

trebmuh:
Just throwing an idea in the dark here (I don't know if it's already in the code or not).
As a user, I'd like to have a bit of an explanation while mouse-hovering the "yellow attention triangle" to understand why it is here.

HTH

cme:
> Just throwing an idea in the dark here (I don't know if it's already in the code or not).
> As a user, I'd like to have a bit of an explanation while mouse-hovering the "yellow attention triangle" to understand why it is here.

I agree. It's already there:

image

Open to suggestions on wording etc.

trebmuh:
Nice!

mauser:
Thanks a lot, merged!

Issue #907, opened 2020-07-12
Set auto velocity checkbox on loading multiple sample layers ignored.


cme:
Hydrogen version *          :  master
Operating system + version  : 
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

The AudioFileBrowser has a "Set automatic velocity" checkbox. Selecting this and loading a sample will cause the velocities of all sample layers to be automatically laid out in linear regions.

When loading multiple samples at the same time, the resultant behaviour is to *always* behave as if that checkbox was selected.

This is probably against most user expectations, and is a pain if the user:
  * is building a set of simple round-robin layers (in which case they would want the exact behaviour of having the checkbox closed), or
  * has already built up a set of layers and set appropriate velocity ranges, and now wants to add more samples (eg. as a second level for round-robin or random samples), the user will find all their previously-edited layers unexpectedly modified, with no Undo facility.

(that last one is particularly nasty)

The action **should** match that which the user has selected in the load dialog.

If the case for apply auto-velocity to multiple selected samples is particularly strong, it would be better to automatically select the "Set automatic velocity" checkbox when multiple files are selected so that the user will have some indication of what will happen and the opportunity to disable it; but it doesn't feel to me that the case for that is very strong.

It also feels wrong (and against reasonable user expectation) that selecting an option in the "Load" dialog should affect samples/layers that have already been loaded, but fixing that would involve more changes to the Instrument Editor UI which seems... painful.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

cme:
Fix applied to master.

Issue #830, opened 2020-04-02
GUI refresh issue in Instruments Layer window


thijz:
Hydrogen version *          : 1368d91
Operating system + version  : ubuntu studio 18.04
Audio driver + version      : jack

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

1) create a new song (default kit is now loaded)
2) add a new instrument (Instruments - Add instrument)
3) select one of the existing instruments and you see the waveform in the instruments layers window :
![image](https://user-images.githubusercontent.com/1384210/78298347-703bc000-7532-11ea-9131-02fe5b3d5e31.png)
4) now select the new instrument (that has no sample loaded yet) and the previous waveform is not cleared :
![image](https://user-images.githubusercontent.com/1384210/78298423-906b7f00-7532-11ea-8ad1-8cd5b0354290.png)
This is quite confusing :-)

cme:
Yea, I noticed this a while ago and forgot about it. The fix (#908) is hilariously trivial :)

cme:
Fixed in master.

PR #908, opened 2020-07-12
Fix instrument editor bugs #907 and #830

cme:
Small fixes for Instrument Editor issues:

  #907 (unexpected velocity setting): removes the code that forces
    auto-velocity on multiple sample load.

  #830 (missing refresh of wave display): the null handling case of
    updateDisplay() sets the name and layer pointer and returned
    early, missing the update() call necessary to start a repaint.


cme:
Hmm, spurious-looking test failure on Travis on Mac OS:
```
Can't open reference file

- File contains data in an unimplemented format.
```

Linux build looks fine.

PR #912, opened 2020-08-03
Fix i18n data install path

dvzrv:
data/i18n/CMakeLists.txt:
Prepend H2_SYS_PATH to the DESTINATION parameter of the install() call
for the i18n files.
Given a CMAKE_INSTALL_PREFIX of '/usr' this installs the i18n files to
/usr/share/hydrogen/data/i18n on a Linux distribution (instead of to the
non-FHS compliant /usr/share/data directory).


mauser:
Thanks for the pull request, merged !  That change slipped in there unintentionally..

PR #919, opened 2020-08-09
Pattern editor (drum tab & piano roll) selection, move/copy/delete

cme:
Still somewhat work-in-progress, but functional enough that it should be representative of the user experience. Any feedback would be appreciated!

Quick summary:

  - Drag to lasso (or shift+arrow keys)
  - Drag selected note to move (or hit "Return" from selection, or on a selected note)
  - Ctrl + click to add or remove from selection
  - Alt+move for fine-grained movement
  - Ctrl + drop or Ctrl + Return to copy (like song editor)


cme:
This is fairly stable now, feature-wise. Selectiion, drag and drop, cut and paste, and menus in  both Pattern Editors.

@Porrumentzio , @trebmuh , @muziker, @digitalraven (and anyone else!) -- if you want to take a look, tear it apart etc. Cheers :)

trebmuh:
Just tried and it's very good so far!

One feature I found myself to try naturally is the capability to select all notes from a line.

Mockup:

h2-select-all

cme:
Good idea. I didn't really consider interaction between the selection and the Instrument List widget, but having a whole-instrument-line selection on the Instrument List seems natural.

Additionally, I now realise that the selection formats won't allow cut and paste between the pattern editor and the instrument line copy/paste functions. I'll add that too.

Great, thanks for the feedback!

trebmuh:
Rebuild from 0ece8d2 and it is very nice.
I can see this being very useful when "_trying what sounds better for my pattern between "hat close", "ride", "hat open"..._".
:+1:

trebmuh:
Another request from me is the following: if I click on a note, then I press Ctrl (and keep it pressed), then I click on another note, then it adds the second note to the selection. Good.

What I'd like to have too is: I rectangle-draw-select a few notes, then I press Ctrl (and keep it pressed), then I rectangle-draw-select another note (or another few notes). Right now, doing so just cancels my first selection which is a bit counter-intuitive.

cme:
> What I'd like to have too is: I rectangle-draw-select a few notes, then I press Ctrl (and keep it pressed), then I rectangle-draw-select another note (or another few notes). Right now, doing so just cancels my first selection which is a bit counter-intuitive.

Excellent idea! Added. Thanks :)

cme:
Although the implementation doesn't feel quite right. Dragging with Control held down will only add to the selection, so if you then make the drag rectangle smaller, it leaves newly selected notes selected even though they're now outside that selection rectangle.

I think perhaps a more sensible behaviour might be for a control+drag to accept all previously selected notes, and notes *currently* under the drag rectangle.

trebmuh:
Marvelous. Thank you very much.
It looks very good like that from my point of view.

cme:
@trebmuh thanks. Hopefully the last change to ctrl-drag meets expectation too

cme:
I have some refactoring of the pattern editors which commonises some of this (and other code), which i think I want to call a separate pull request. So for the moment I'm considering this feature complete and will work on updating documentation. 

@mauser, anything else that should happen before merging?

mauser:
Wonderful feature! I really like it! 

There is just one minor thing which i noticed when testing: 

- After selecting notes in the pattern editor, i could find no easy way to "disband" the selection again. Most other programs would do that after the clicks in another area (outside of the selection), but this just placed new notes in the pattern editor. Is it possible to do this without creating a new note?

src/gui/src/PatternEditor/PianoRollEditor.cpp
```diff
@@ -491,50 +585,79 @@ void PianoRollEditor::addOrRemoveNote( int nColumn, int nRealColumn, int nLine,
...
-	pDraggedNote = m_pPattern->find_note( nColumn, nRealColumn, pSelectedInstrument, notekey, octave );
-
-	int oldLength = -1;
-	float oldVelocity = 0.8f;
-	float oldPan_L = 0.5f;
-	float oldPan_R = 0.5f;
-	float oldLeadLag = 0.0f;
-	int oldNoteKeyVal = 0;
-	int oldOctaveKeyVal = 0;
+	Note* pFoundNote = pFoundNote = m_pPattern->find_note( nColumn, nRealColumn, pSelectedInstrument,
```
mauser, line 593: Is that intentional?

src/gui/src/PatternEditor/PianoRollEditor.cpp
```diff
@@ -729,32 +856,132 @@ void PianoRollEditor::addOrDeleteNoteAction( int nColumn,
...
+									  Note::Octave newOctave,
+									  Note::Key newKey,
+									  Note *pNote)
+{
+	Hydrogen *pHydrogen = Hydrogen::get_instance();
+	Song *pSong = pHydrogen->getSong();
+
+	AudioEngine::get_instance()->lock( RIGHT_HERE );
+	PatternList *pPatternList = pSong->get_pattern_list();
+	InstrumentList *pInstrumentList = pSong->get_instrument_list();
```
mauser, line 931: clang says this value is never read, which seems to the case :)

src/gui/src/PatternEditor/PianoRollEditor.cpp
```diff
@@ -962,54 +1185,285 @@ void PianoRollEditor::mouseReleaseEvent(QMouseEvent *ev)
...
+		}
+	}
+	update();
+}
+
+void PianoRollEditor::deleteSelection()
+{
+	if ( m_selection.begin() != m_selection.end() ) {
+		// Delete a selection.
+		InstrumentList *pInstrumentList = Hydrogen::get_instance()->getSong()->get_instrument_list();
```
mauser, line 1232: Value is never read

src/gui/src/PatternEditor/PianoRollEditor.cpp
```diff
@@ -962,54 +1185,285 @@ void PianoRollEditor::mouseReleaseEvent(QMouseEvent *ev)
...
+		m_selection.clearSelection();
+	}
+}
+
+///
+/// Copy selection to clipboard in XML
+///
+void PianoRollEditor::copy()
+{
+	Song *pSong = Hydrogen::get_instance()->getSong();
```
mauser, line 1264: Value is never read

src/gui/src/PatternEditor/PianoRollEditor.cpp
```diff
@@ -1085,16 +1532,10 @@ void PianoRollEditor::editNotePropertiesAction( int nColumn,
 {
 
 	Hydrogen *pEngine = Hydrogen::get_instance();
+	PatternList *pPatternList = pEngine->getSong()->get_pattern_list();
```
mauser, line 1535: According to clang, the value is never read

mauser:
>     * After selecting notes in the pattern editor, i could find no easy way to "disband" the selection again. Most other programs would do that after the clicks in another area (outside of the selection), but this just placed new notes in the pattern editor. Is it possible to do this without creating a new note?

After looking at the code, i found the "clear selection" functionality in the context menu. This was what i was looking for! I suppose that most people are used to do that by clicking somewhere out of the selection area, but i see that this might be a bit harder to implement. 

One related,smaller thing: The text-input marker appears out of nowhere after "clear selection" is executed.

mauser (approved):

src/gui/src/PatternEditor/PianoRollEditor.cpp
cme: No, editing error.  I guess in coalescing a previously separate declaration and initialisation

src/gui/src/PatternEditor/PianoRollEditor.cpp
cme: >     * After selecting notes in the pattern editor, i could find no easy way to "disband" the selection again. Most other programs would do that after the clicks in another area (outside of the selection), but this just placed new notes in the pattern editor. Is it possible to do this without creating a new note?

Hmm, yes, this sounds like it's the right thing to do. Excellent! Will do, thanks.

Thanks for the review comments!

cme:
> > * After selecting notes in the pattern editor, i could find no easy way to "disband" the selection again. Most other programs would do that after the clicks in another area (outside of the selection), but this just placed new notes in the pattern editor. Is it possible to do this without creating a new note?
> 
> After looking at the code, i found the "clear selection" functionality in the context menu. This was what i was looking for! I suppose that most people are used to do that by clicking somewhere out of the selection area, but i see that this might be a bit harder to implement.

'Escape' and 'Shift+Ctrl-A' (or whichever platform dependent definition Qt uses for "Select None") also do this as those seemed like sensible things to assume the user might do. Left (or middle) click has now been added to that list of actions.

> One related,smaller thing: The text-input marker appears out of nowhere after "clear selection" is executed. This happens also when using other actions from the context menu.

Excellent find. Fixed, thanks :)

mauser:
Perfect, it is merged now! This is really a great feature which has been missing for so long. Nice to see this finally in Hydrogen!

trebmuh:
Congrats !

Issue #914, opened 2020-08-04
ui is too big


Reaper10:
Hydrogen version *          :1.0
Operating system + version  : windows 10
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

the ui is too big

![Annotation 2020-08-04 102926](https://user-images.githubusercontent.com/8137561/89312940-7d320280-d63d-11ea-8c09-ff7c433e6b70.jpg)

cme:
Hi, can you tell us about your system, please? 

Is your monitor High-DPI and very small? Do you have Windows UI scaling turned on?

Your screenshot has a resolution of 144DPI which would be consistent with having hidpi emulation but user interface scaling set to 50%? 

On a straight win10 without hidpi or any scaling set, the user interface is as expected:

![Screenshot 2020-08-04 at 17 58 21](https://user-images.githubusercontent.com/768466/89322486-2c8dca00-d67c-11ea-8187-b42ccc032862.png)

Reaper10:
dose this help?
![Annotation 2020-08-04 145017](https://user-images.githubusercontent.com/8137561/89338283-fa6f6e80-d661-11ea-831e-eab1c7512dae.jpg)

cme:
Thank you. Yes, with that and some research I see what the problem is.

I guess 150% is the default that Windows picks for scaling given the size of your monitor. The issue is that by default Qt rounds the UI scaling factor to the nearest integer. At 150% scaling, that means it'll round it up to scale by 200%. Yuck.

If you try setting your desktop scale to 125% rather than 150%, Qt will select 100% scale for Hydrogen to use, which will be half the size.

That may be too small, and indeed it seems wrong to have to set the scaling of your entire desktop just for one application!

It's possible to get it to scale to match your UI size by going to:

  Windows Settings -> Edit environment variables for your account

Hit "New..." and enter the variable name

```QT_SCALE_FACTOR_ROUNDING_POLICY```

and give it the value:

```PassThrough```

or the value

```Floor```

to scale to 100% regardless of your Windows UI scaling factor.

cme:
Fix is on master, pending release.

PR #921, opened 2020-08-12
Allow user to select high dpi scaling rounding policy

cme:
This addresses #914 by allowing the user to choose the rounding mode.

Not yet tested on Windows as my MXE is in the process of rebuilding.


mauser:
Thanks a lot, merged!

cme:
I should note that I haven't actually tested this on Windows -- the Appveyor build uses Qt 5.12 (which at least let me test the fallback behaviour!) but I haven't had time to get a working mxe + windows build going locally to test.

cme:
Finally managed to pull together a working mxe build from mac os -> windows again to test this, and found that it doesn't actually work.

The issue is the call to set the scaling factor has to happen before initialisation of the QApplication object. However, there's circular ordering dependencies that mean there's no valid ordering that allows us to use a preference so early:

 * set rounding policy must precede QApplication initialisation
 * QApplication initialisation must precede Filesystem::bootstrap()
 * Filesystem::bootstrap must precede preferences loading
 * Preferences loading must precede setting rounding policy

So it looks like making this actually *work* might need some substantial refactoring to break apart initialisation dependencies. I'll continue to take a deeper look.

mauser:
We could also add the rounding policy as a command line param. This is a bit tricky for the windows users, but it would be a workaround for the dependency problem. What i don't like about this solution is that the command line params are more or less "invisible" for a standard windows user. The UI settings in the preferences dialog are easier to find.

cme:
Having gone through the dependencies, it looks like the cycle can be broken by changing Filesystem::bootstrap() to act in two stages (which seems reasonable given the bootstrap nature of the method). The only dependency it has on QApplication is for sys_data_path.

I have a working version which I'm going to test on Windows and Linux shortly, which basically:

  * delays pQApp initialisation until later in main() in the GUI app
  * Filesystem::bootstrap() checks QCoreApplication::isStartingUp(). If so, it skips setting sys_data_path, but sets other paths (so that Preferences can be loaded). A null string value is set in __sys_data_path.
  * adds a second call to Filesystem::bootstrap() in main(), after pQApp initialisation, to allow the bootstrap to complete.
  * Filesystem::sys_data_path() asserts that it isn't returning the dummy value, to make sure that no calls to sys_data_path() happen before the bootstrap is completed.

The CLI programs should still successfully bootstrap on the first call since they have should QCoreApplication initialised at this point.

cme:
Filesystem::bootstrap couldn't be cleanly split after all, as some of the initialisation does need access to the application path via QApplication. So #936 is alternate solution which uses a dummy core app to bootstrap configuration before the GUI app starts.

PR #930, opened 2020-08-20
Fix #929: translation selection prefers languages in wrong order.

### LINKED ISSUE #929: macOS locale detection broken
Hydrogen version *          : 1.0.0
Operating system + version  : MacOS 10.15.6
Audio driver + version      : n/a

Hydrogen does not follow the macOS preferred system locale when more than one locale is defined.

On startup, Hydrogen goes through the list of available languages defined in the Language and Region control panel, and uses the first one it finds an i18n file for in the Hydrogen package. Since macOS reports UK English (correctly) as en_GB, Hydrogen ignores it — the i18n file looks like it's expecting a locale approximating en_UK, but that does not exist, [per the language codes defined in ISO 693-1](http://www.andiamo.co.uk/resources/iso-language-codes). 

As an additional wrinkle on this bug, If Portuguese (Brazilian) is anywhere in the macOS list of available languages, Hydrogen uses it. It does not matter where Portuguese is in the list or what else is in the language list.

With UK English set as the preferred language in macOS, the only way to get Hydrogen to display an English UI is to remove _all_ other preferred languages from the macOS Language and Region control panel.

Hydrogen package contents: 
Hydrogen package

Language settings: French preferred, Brazilian Portuguese included:
Imgur

Hydrogen UI with those language settings:
Imgur

Language settings (UK English preferred, with non-Portuguese alternates):
Imgur

Hydrogen UI with those settings
Imgur

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

cme:
Explicitly search through user-prefererred language options and
look for a matching translation (either specific regional language,
or generic language). This avoids Qt's default problem of preferring
an exact-match language+region for any available UI language over an
imprecise / generic match.

Also add generic English translation, and British English (just a couple of different spellings)

### DIFF
--- a/data/i18n/updateTranslations.sh
+++ b/data/i18n/updateTranslations.sh
@@ -40,6 +40,7 @@ $CMD ../../data/i18n/hydrogen_el.ts
 $CMD ../../data/i18n/hydrogen_gl.ts
 $CMD ../../data/i18n/hydrogen_sr.ts
 $CMD ../../data/i18n/hydrogen_ca.ts
+$CMD ../../data/i18n/hydrogen_uk.ts
 $CMD ../../data/i18n/hydrogen_en.ts
 $CMD ../../data/i18n/hydrogen_en_GB.ts
 
--- a/data/i18n/updateTranslations.sh
+++ b/data/i18n/updateTranslations.sh
@@ -40,7 +40,8 @@ $CMD ../../data/i18n/hydrogen_el.ts
 $CMD ../../data/i18n/hydrogen_gl.ts
 $CMD ../../data/i18n/hydrogen_sr.ts
 $CMD ../../data/i18n/hydrogen_ca.ts
-$CMD ../../data/i18n/hydrogen_uk.ts
+$CMD ../../data/i18n/hydrogen_en.ts
+$CMD ../../data/i18n/hydrogen_en_GB.ts
 
 echo "Creating *.qm files"
 cd ../../data/i18n
--- a/data/i18n/hydrogen_ca.ts
+++ b/data/i18n/hydrogen_ca.ts
@@ -380,7 +380,7 @@
         <translation type="unfinished">Exportar arxiu MIDI</translation>
     </message>
     <message>
-        <source>Directory %1 does not exists</source>
+        <source>Directory %1 does not exist</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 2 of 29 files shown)


elpescado:
That's cool! I'd really want to see some unit tests for this, though. I think there are some cases missing: if user has preference for a "generic" dialect (say, `pt`), but there's only regional dialect available (eg. `pt_BR`).

Alternatively, we might declare that that case is not supported, and ensure that translation file for "generic" dialect (`hydrogen_pt.qm` in that case) is always present.

cme:
> That's cool! I'd really want to see some unit tests for this, though. I think there are some cases missing: if user has preference for a "generic" dialect (say, `pt`), but there's only regional dialect available (eg. `pt_BR`).
> 
> Alternatively, we might declare that that case is not supported, and ensure that translation file for "generic" dialect (`hydrogen_pt.qm` in that case) is always present.

Agreed, I think it's not at all clear what the right thing to do in such cases is; particularly since there's such a lot of variation between languages in the similarity of dialects!

I can add a test that at least can illustrate that for Hydrogen's use. Might need a little refactoring to take the uiLanguages call out or that.

Thanks!

cme:
Updated to add unit test for default selection, and add a Language menu to the preferences dialog box (in a very obvious position on the first tab, in case a user needs to navigate to change from an unfamiliar language).

trebmuh:
> ... add a Language menu to the preferences dialog box ...

Is this taken into account directly or is it necessary to restart hydrogen?

cme:
> > ... add a Language menu to the preferences dialog box ...
> Is this taken into account directly or is it necessary to restart hydrogen?

It seems to be a hard requirement that translations are loaded before creating any widgets, so it needs a restart. The change includes a message box to inform the user of that.

(Yet another string to be translated!)

cme:
image


image

image

trebmuh:
> Yet another string to be translated!

Typo: languague -> language.

data/i18n/updateTranslations.sh
```diff
@@ -40,7 +40,8 @@ $CMD ../../data/i18n/hydrogen_el.ts
 $CMD ../../data/i18n/hydrogen_gl.ts
 $CMD ../../data/i18n/hydrogen_sr.ts
 $CMD ../../data/i18n/hydrogen_ca.ts
-$CMD ../../data/i18n/hydrogen_uk.ts
+$CMD ../../data/i18n/hydrogen_en.ts
+$CMD ../../data/i18n/hydrogen_en_GB.ts
```
trebmuh, line 44: It seems you removed the ukrainian translation update by mistake.

cme: Well spotted, thanks!

PR #943, opened 2020-09-20
Osc scripting of Timeline and Jack options

theGreatWhiteShark:
I did add some OSC commands to automate some transport related buttons of the GUI and to add/delete tempo markers from the timeline.

e.g. try
```bash
oscsend localhost 9000 /Hydrogen/TIMELINE_ADD_MARKER if 14 199
```

But in addition I also restructured large parts of the Timeline code. I had some difficulties understanding the dynamics of this class since almost all code acting on its member variables was located outside of the class (especially within the GUI). I, instead, put it all into the Timeline class itself and only provide a read-only version of the member variables to the GUI. This way - at least for me - it is way more obvious what is happening at which point. 

(Sorry for entangling two different things in here)

Edit:
Adding/removing the tempo marker is not covered by the undo/redo action yet. In our current Event type only a single integer can be provided, which is not enough to let the GUI know both position and speed of the marker. I will make an issue and take care of this later.

### DIFF
--- a/src/core/include/hydrogen/core_action_controller.h
+++ b/src/core/include/hydrogen/core_action_controller.h
@@ -118,10 +118,61 @@ class CoreActionController : public H2Core::Object {
 		 */
 		bool quit();
 
+		/**
+		 * (De)activates the usage of the Timeline.
+		 *
+		 * Note that this function will fail in the presence of the
+		 * Jack audio driver and an external timebase master (see Hydrogen::haveJackTimebaseClient()).
+		 *
+		 * @param bActivate If true - activate or if false -
+		 * deactivate.
+		 *
+		 * @return bool true on success
+		 */
 		bool activateTimeline( bool bActivate );
+		/**
+		 * Adds a tempo marker to the Timeline.
+		 *
+		 * @param nPosition Location of the tempo marker in bars.
+		 * @param fBpm Speed associated with the tempo marker.
+		 *
+		 * @return bool true on success
+		 */
 		bool addTempoMarker( int nPosition, float fBpm );
+		/**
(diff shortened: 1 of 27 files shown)


theGreatWhiteShark:
I'm a little bit puzzled about the error log in the Ubuntu build.

> ‘shared_ptr’ is not a member of ‘std’

Aren't we using C++14? Or is there some language/compiler detail I am missing?

cme:
> I'm a little bit puzzled about the error log in the Ubuntu build.
> 
> > ‘shared_ptr’ is not a member of ‘std’
> 
> Aren't we using C++14? Or is there some language/compiler detail I am missing?

Is `<memory>` included anywhere from `timeline.h`? It doesn't look like it on first glance.

theGreatWhiteShark:
Indeed. Thanks! Quite strange it did work on my Ubuntu machine and the MacOS build without it.

mauser:
Thanks for re-writing that Timeline code, that was really necessary :smiley:  The code looks code, i hope that i can test the OSC commands at the weekend!

mauser:
For the record: This pull requests adds the following OSC commands:
| Name  	| Parameter   	|   	
|---	|---	
| TIMELINE_ACTIVATION  	|  i 	|   	
| TIMELINE_ADD_MARKER 	|  if  	|   	
| TIMELINE_DELETE_MARKER|  i   |   	
| JACK_TRANSPORT_ACTIVATION|  i   |   	
| JACK_TIMEBASE_MASTER_ACTIVATION|  i   |

mauser:
Hey Phil,

the pull request has been merged to the master branch. Thanks a lot!

PR #944, opened 2020-09-22
3 strings missing i18n

trebmuh:
I've spotted a place were strings were missing translation (when exporting a drumkit):

hydrogen-traduction-manquante

Then, I've looked into the code and here is my try to fix those.

I hope (and think) I did it correctly but since I don't have the time to test a rebuild with this patch right now, I'd appreciate if a programmer could validate this PR after checking it'll not make all the birds exploding. Thanks.


mauser:
Thanks a lot, looks good to me!

PR #950, opened 2020-10-02
Fix #922: Rewriting Jack synchronization in the presences of an external timebase master

### LINKED ISSUE #922: Transport out of sync in certain condition
Hydrogen version *          : 1.0.0
Operating system + version  : ubuntu studio (Focal Fossa 20.04 LTS)
Audio driver + version      : Alsa 1.0.25

I'm using Hydrogen in combination with Ardour (as time master) and Jack transport synchronization under Ubuntu Studio. Because I'm using live recorded audio files I need bpm speeds with decimal fractions and some speed changes in a song.
There is an issue concerning moving the play-head. When a song has a speed change then there is a small time offset between Ardour and Hydrogen after the play-head is set to a point behind the speed change.

theGreatWhiteShark:
I reintroduced the original code for updating the transport position on relocation based on the BBT information in the presence of an external Jack timebase master.

This one is on me since I removed it prior to 1.0.0. Conceptional this code is required because otherwise there is no way to sync H2 and Jack if there are tempo markers and the transport got relocated by the user. I didn't thought of this scenario. But the original code was buggy and only worked for patterns of length 8. I fixed this too and it should now work for all lengths.

In addition I introduced two debugging function to the Jack audio driver. Some print statements I have introduced at least 10 times by now and I something tells me that this might was not the last time.

---

Update:

	 * Since Hydrogen uses both fixed pattern lengths and recalculates
	 * the tick size each time it encounters an alternative tempo, its
	 * transport is incompatible to the BBT information provided by 
	 * the Jack server. Only if the length of each pattern corresponds
	 * to the measure of the respective bar in the timebase master
	 * application, the bar information provided by Jack can be used
	 * directly to determine chosen pattern. If this, however, is not
	 * the case - which can quite easily happen - a complete history 
	 * of all measure and tempo changes would be required to correctly
	 * identify the pattern. Since this is not provided by Jack, one
	 * has to either assume the measure or tempo to be constant or 
	 * that the user took care of adjusting the lengths properly.

The user has now the option to either only provide different speeds via the Jack timebase master (TBM) or both speeds and measures but, in addition, has to make sure the length of each pattern matches the measure of the corresponding bar in the TBM. This behavior can be chosen via the Preferences menu

### DIFF
--- a/data/i18n/hydrogen_de.ts
+++ b/data/i18n/hydrogen_de.ts
@@ -2777,10 +2777,6 @@ The path to the script and the scriptname must be without whitespaces.</source>
         <translation>konstanter Takt</translation>
     </message>
     <message>
-        <source>constant tempo</source>
-        <translation>konstantes Tempo</translation>
-    </message>
-    <message>
         <source>matching bars</source>
         <translation>Übereinstimmende Takte</translation>
     </message>
--- a/data/i18n/hydrogen_de.ts
+++ b/data/i18n/hydrogen_de.ts
@@ -452,7 +452,7 @@ Overwrite the existing file?</source>
     </message>
     <message>
         <source>Directory %1 does not exist</source>
-        <translation type="unfinished">Verzeichnis %1 existiert nicht</translation>
+        <translation>Verzeichnis %1 existiert nicht</translation>
     </message>
 </context>
 <context>
@@ -572,10 +572,6 @@ Overwrite the existing file?</source>
     <message>
         <source>Interpolation: </source>
         <translation>Interpolation: </translation>
-    </message>
-    <message>
-        <source>Choose type of interpolation methode</source>
-        <translation>Wähle eine Interpolationsmethode</translation>
     </message>
     <message>
         <source>Linear</source>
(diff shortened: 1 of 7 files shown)


theGreatWhiteShark:
There are still a couple of things to do

- [ ] ensure synchronization works in the presences of a couple of different measures
- [ ] introduce three modes of synchronization (assuming constant tempo, constant measure, same pattern length as measures in TBM)
- [ ] Establish option for chosing one of these modes in the GUI/preferences
- [ ] Introduce section in the manual how the three different options work
- [ ] Open issue to fix this inconsistency at some point in the future

theGreatWhiteShark:
Since I introduced a new preference option which is only related to the Jack audio driver I implemented it in such a way the option is not shown if another driver is selected. @mauser @cme @trebmuh would think this behavior to be more intuitive than merely disabling the options meant for another driver (like the per instrument output). I think this redundant information only confuses the user. But anyway, I would like to keep/make things consistent.

theGreatWhiteShark:
I decided to hide the "per track output" and the "connect defaults" options for all audio drivers other than Jack too.

PR #955, opened 2020-10-09
Further OSC commands

theGreatWhiteShark:
I introduced some more OSC commands which will be handy for some debugging I do plan during the next week.

- (De)activating loop mode
- Switching between Song and Pattern mode
- Relocating to a particular pattern group

I made sure to move all the engine-related code into the CoreActionController. E.g. there were numerous occasions of the GUI relocating the transport position without taking care of adjusting the tempo according to the timeline. Since the actions of the GUI do happen not at a fixed position in the process cycle of the audio engine, this could very well produce some minor (probably inaudible) glitches with the tempo getting only adjusted in the next process cycle.


mauser (approved):

mauser:
Thanks a lot, merged!

Issue #853, opened 2020-04-27
Keyboard note input doesn't work with a Dvorak keyboard


cme:
Hydrogen version *          : master
Operating system + version  : all
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Entering notes using the keyboard is set up for Qwerty, Qwertz and Azerty, determined by guessing based on the locale. This doesn't work very well for Dvorak layouts where the notes will be distributed haphazardly around the keyboard, and will also present the wrong layout to users who are using a keyboard for a language other than the default for their locale. 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

cme:
After a little digging, Qt's support for keyboard layout interrogation is basically non-existent. Its possible to easily fix the issue of using a non-local keyboard by asking for the input locale instead of the locale, but for Dvorak / Colemak etc there's no way to find the layout.

It's possible to get the right physical keys using virtual keycodes, but the values are platform-dependent so would require specific support for each platform and a fallback for any newer platform.

An alternative would be to allow a user-editable mapping, which is an even bigger project.

The audience for such a change is almost certainly very small indeed. Possibly just me... ;)

Issue #793, opened 2019-12-28
Inconsistent naming of drumkits


theGreatWhiteShark:
For some of the drumkits in H2 its overall name (the one in the second-level `<name>` node in the corresponding _drumkit.xml_ file) is not the same as the name of the folder it is stored under _~/.hydrogen/data/drumkits/_.

Right now H2 is only using the folder name for loading a drumkit (when loading a Song). But due to this discrepancy you can basically not tell which drumkit you are dealing with when just inspecting the content of its dedicated folder (which gives me quite some headache in #792 )

The problem occurs (at least) in

- ForzeeStereo
- Roland_MC-307_CR78&Cheaps
- rumpf_kit_z01_h2
- Boss_DR-110

theGreatWhiteShark:
Closed with #1620

PR #964, opened 2020-10-22
Pattern editor refactor (and NotePropertiesRuler selection support)

cme:
Refactoring and cleanup of pattern editors to share commonalities between drum and piano editors


theGreatWhiteShark:
Hey @cme. 

Since you already delved deeply into the GUI I would like to ask you a question about the current implementation. AFAICS all the Qt stuff is created using `new` - on the heap - in the constructors of our GUI classes. But nothing is ever freed in the destructors. Do we have a massive memory leakage in the GUI or am I missing some Qt or C++ magic doing this in the background? After all e.g. the Mixerlines are destroyed and recreated when switching drumkits.

cme:
> Since you already delved deeply into the GUI I would like to ask you a question about the current implementation. AFAICS all the Qt stuff is created using `new` - on the heap - in the constructors of our GUI classes. But nothing is ever freed in the destructors. Do we have a massive memory leakage in the GUI or am I missing some Qt or C++ magic doing this in the background? After all e.g. the Mixerlines are destroyed and recreated when switching drumkits.

Qt magic mostly. All QWidgets in a QApplication are linked together in a widget tree (established by every QWidget constructor taking a 'parent' pointer parameter), which is the definitive owner of all widgets. When a widget is deleted (including the QApplication at application termination) the tree is recursively destroyed.

cme:
Though for any non-Widget objects that are allocated in constructors, they ought to be freed in destructors. Have you spotted any that aren't? Those would be leaky!

theGreatWhiteShark:
Nice. I had a feeling that something like this couldn't be happening without anyone noticing otherwise.

> Though for any non-Widget objects that are allocated in constructors, they ought to be freed in destructors. Have you spotted any that aren't? Those would be leaky!

I haven't searched the code base. Yet alone in the constructor of `MixerLine` - which is displayed in my editor right now - there are MAX_FX + 2 `Action` I don't find the corresponding `delete` command to. There's probably more. Well, the Qt code kinda encourage missing the destruction for people unfamiliar with the particularities of C++ (like me :D )

cme:
Okay, so not all of them will be leaky... The Actions allocated in `MixerLine()` are owned by the Rotary they're added to. That inherits from MidiLearnable and so be deleted in `~MidiLearnable()`. Since Rotary is a QWidget under Mixer, it'll be destroyed when the Mixer is destroyed, so not a leak in this case.

theGreatWhiteShark:
Ah, I see. Thanks for clarifying!

cme:
I'm almost done with this for the moment -- the next thing I wanted to do was add dragging selected notes to adjust multiple notes at a time in the Note Properties Editor, similar to the main part of #664 , (unless @theGreatWhiteShark would rather take that as part of #793 !) which is definitely more "new feature" than "refactor" so may be a separate PR.

@theGreatWhiteShark , @mauser , if either of you have time and inclination to have a look over some of this I'd appreciate it! Thanks :)

theGreatWhiteShark:
Sure. I try to do so soon.

> (unless theGreatWhiteShark would rather take that as part of 793 !)

Not at all. For me it's actually more fun to tweak/fix the core part and there are still a lot of things to do

cme:
> Sure. I try to do so soon.

Awesome, thank you! :)

mauser:
Hi! 

I have a question about the NotePropertiesRuler selection support. Is this supposed to work at this point in time or am i doing sth. wrong? I can select the notes, but i'm not sure how to edit them (for example change the velocity for a selection of notes). Or does only selection work with this PR, but not changing?

mauser:
Btw. the precompiled headers give me a performance gain of 10-20% during compilation, not bad!!

cme:
> I have a question about the NotePropertiesRuler selection support. Is this supposed to work at this point in time or am i doing sth. wrong? I can select the notes, but i'm not sure how to edit them (for example change the velocity for a selection of notes). Or does only selection work with this PR, but not changing?

Yes, selection is implemented but doesn't do anything in NotePropertiesRuler yet. This was entering the grey area where it seemed silly not to implement the selection mechanism to complete the support for the subclass, but was creeping part 'refactor' into 'new feature'. I'm continuing to work on adding actual selection editing but not sure yet whether that will be a i separate PR or if I'll fold into this. Probably the latter.

theGreatWhiteShark:
I have a couple of comments (and suggestions we can tackle for the next release)

- Cycling the focus with `tab` works. But with `shift + tab` the focus will remain in the `SongEditor` and does not switch back to the `NotePropertyRuler`. 
- (And now I finally understood what this strange blackish border is supposed to mean. It's the focus! - Or is it? When drawing a new pattern in the `SongEditor` it disappears and reappears at the clicked position after pressing `Ctrl + A`. I would suggest we use a different method to indicate focus (maybe using foreground and background color). But not as part of this PR.)
- I'm unable to undo deletions of group selections of patterns in the `SE`. But undoing deleting multiple notes in the `PE` does work fine.
- I'm a little bit puzzled about the use of the `NoteKey` in `NPR`. Isn't this exactly what the `PianoRollEditor` provides? Should we drop the former?
- (Since `SE` and `PE` are now children of the same base class, how a *draw mode* for the `PE`? This would make their UX even more consistent and with keyboard shortcuts for adjusting the resolution this could seriously speed up pattern generation. But, again, not as part of this PR.)
- (I was a little bit surprised that - while focus was on the `PE` - `+` does increase the tempo of the song. Maybe we can come up with focus specific keyboard shortcuts to ease editing. I would volunteer to create a cheat sheet like [this](https://github.com/theGreatWhiteShark/lightning-keymap-mode/blob/master/res/cheatsheet.png) one to put them all in one place afterwards.)

cme:
Thanks for the excellent feedback!

> * Cycling the focus with `tab` works. But with `shift + tab` the focus will remain in the `SongEditor` and does not switch back to the `NotePropertyRuler`.

Interesting, I can't replicate that. It's *possible* it might show up under Linux but not Mac OS, so I'll try it out on Linux shortly. 

> * (And now I finally understood what this strange blackish border is supposed to mean. It's the focus! - Or is it? When drawing a new pattern in the `SongEditor` it disappears and reappears at the clicked position after pressing `Ctrl + A`. I would suggest we use a different method to indicate focus (maybe using foreground and background color). But not as part of this PR.)

The appearing / disappearing behaviour was intended to make sure that the cursor didn't show up and confuse mouse-only users (the behaviour can be turned off to always show it in Preferences; I should document that too).

Unhiding it on `Ctrl+A` etc was consistent with the idea that "any keyboard activity should unhide the cursor" but it's also likely that copy, delete, paste etc might be used in conjunction with just the mouse without needing to know about a keyboard cursor position, so perhaps leaving it hidden on those shortcuts might be better.

> * I'm unable to undo deletions of group selections of patterns in the `SE`. But undoing deleting multiple notes in the `PE` does work fine.

Good catch, it should. That's old code that I left unchanged and just never noticed it didn't have undo functionality!

> * I'm a little bit puzzled about the use of the `NoteKey` in `NPR`. Isn't this exactly what the `PianoRollEditor` provides? Should we drop the former?

Dropping NoteKey in the NPE would make some things a bit tidier, but I don't have a feel for its use. I don't see anything that it does that the PianoRollEditor doesn't do, apart from making it quick and easy to transpose notes up or down whole octaves. I don't see much harm in keeping it though.

> * (Since `SE` and `PE` are now children of the same base class, how a _draw mode_ for the `PE`? This would make their UX even more consistent and with keyboard shortcuts for adjusting the resolution this could seriously speed up pattern generation. But, again, not as part of this PR.)

I guess the use for this would be to draw continuous series of notes at time intervals set by the grid resolution? The "fill notes..." function covers this but only for the whole pattern duration. 

> * (I was a little bit surprised that - while focus was on the `PE` - `+` does increase the tempo of the song. Maybe we can come up with focus specific keyboard shortcuts to ease editing. I would volunteer to create a cheat sheet like [this](https://github.com/theGreatWhiteShark/lightning-keymap-mode/blob/master/res/cheatsheet.png) one to put them all in one place afterwards.)

I mostly think so too, but am a bit hesitant since tempo change in particular is something that might be used in a live context, so changing behaviour of those keys feels hazardous?

Thanks!

theGreatWhiteShark:
>  The appearing / disappearing behaviour was intended to make sure that the cursor didn't show up and confuse mouse-only users (the behaviour can be turned off to always show it in Preferences; I should document that too).
>
> Unhiding it on Ctrl+A etc was consistent with the idea that "any keyboard activity should unhide the cursor" but it's also likely that copy, delete, paste etc might be used in conjunction with just the mouse without needing to know about a keyboard cursor position, so perhaps leaving it hidden on those shortcuts might be better.

Ah, I see. It's the cursor, not the focus. Just tried this new feature. Awesome! Is it documented?

> I don't see anything that it does that the PianoRollEditor doesn't do, apart from making it quick and easy to transpose notes up or down whole octaves.

If I'm not mistaking transposing is only possible for the first note when entering multiple notes at a single position via the piano roll . But then again, when entering them via the piano roll there is not much need to alter their octave in the NPR. But to be honest I never use these features anyway. :D

>  I guess the use for this would be to draw continuous series of notes at time intervals set by the grid resolution? The "fill notes..." function covers this but only for the whole pattern duration.

Your right. This would probably add only minimal benefit for the user while bloating the code.

> so changing behaviour of those keys feels hazardous?

Definitely. IMHO the more keyboard shortcuts/functions we provide, the more useful would be another tab in the Preferences allowing the user to alter the keyboard shortcuts. So, we could we could set those functions to available keys and e.g. provide two different shortcut profiles "compatibility" and "recent". This way the user could also work around #853 without altering the code.

cme:
> Ah, I see. It's the cursor, not the focus. Just tried this new feature. Awesome! Is it documented?

Yes, but fairly minimally at the moment. This PR will add a couple of things that need to be documented.

> ...
> > so changing behaviour of those keys feels hazardous?
> 
> Definitely. IMHO the more keyboard shortcuts/functions we provide, the more useful would be another tab in the Preferences allowing the user to alter the keyboard shortcuts. So, we could we could set those functions to available keys and e.g. provide two different shortcut profiles "compatibility" and "recent". 

I like this idea. I should make a Feature Request for that.

One thing I'm not quite sure about is the interaction of keymaps with the 'input mode' (piano/pitch vs drum/instrument), where keys can have more than one effect. Either the keymaps define a key's behaviour in both, or perhaps the functionality of 'input mode' just becomes picking either a piano-style keymap or a drum-style one?

> This way the user could also work around #853 without altering the code.

+1 !

theGreatWhiteShark:
> One thing I'm not quite sure about is the interaction of keymaps with the 'input mode' (piano/pitch vs drum/instrument), where keys can have more than one effect. Either the keymaps define a key's behaviour in both, or perhaps the functionality of 'input mode' just becomes picking either a piano-style keymap or a drum-style one?

That's right. There is the input mode. Another option would be to have input mode-instrument, input mode-drumkit, and edit mode with a keyboard shortcut to cycle between them and a graphical feedback for which mode was chosen.

I never really used the input mode. But I have a feeling that introducing additional keys to this mode would make recording using them more error prone.

theGreatWhiteShark:
Ah, and another thing: rectangle selection using the cursor + Shift works fine in the PatternEditor but unfortunately not in the SongEditor

cme:
> Ah, and another thing: rectangle selection using the cursor + Shift works fine in the PatternEditor but unfortunately not in the SongEditor

Yep, the SongEditor hasn't been refactored to use the same Selection model yet. Still working on that, it'll be a different PR.

theGreatWhiteShark:
Some more notes (well, they are somewhat further things that could be done on top of your PR. Not something that needs implement right now (especially not before the release. Just putting them here because of a lack of another place. Maybe a Github project?)

I just found some time to *use* Hydrogen again and I'm amazed how much the interaction improved by selection and the keyboard input. However, there are still quite a number of low hanging fruits like
- Supporting [page-up], [page-down],  [home], and [end]
- Allow for scrolling during the drag selection
- Allow for a drag selection when pressing e.g. [ctrl]+mouse even during the drawning mode (I'm not quite sure if we still have to have two separate modes at all)

and there are some parts that are still a huge pain to interact with and could benefit from the refactoring too, like the automation path.

I would suggest we could dwell on the UX including consistent cursor changes, highlighting the current focus, providing a configurable keyboard layout etc. for the 1.2 release.

cme:
> * Supporting [page-up], [page-down],  [home], and [end]

Home and End should be covered by the standard Qt mappings of `QKeySequence::MoveToStartOfLine` and `QKeySequence::MoveToEndOfLine`, ie. move to the time beginning and ends of the current pattern / instrument line.

But yes, page-up/page-down should be added too. Thanks!

> * Allow for scrolling during the drag selection

Excellent idea! Yes.

> * Allow for a drag selection when pressing e.g. [ctrl]+mouse even during the drawning mode (I'm not quite sure if we still have to have two separate modes at all)

Having already folded most of the functionality of the drawing mode into the select mode (indeed *all* the functionality of it that existed before 1.0), I think it makes sense to change the default mode to be the 'select' mode, but I think there's enough use for the continuous-draw mode that it's worth keeping Draw mode. If it's not the default mode, I don't think interacting with selections in it seems necessary?

> I would suggest we could dwell on the UX including consistent cursor changes, highlighting the current focus, providing a configurable keyboard layout etc. for the 1.2 release.

Sounds good. I'll probably merge this in the next few days (as the Song Editor refactor work I'm doing is on a branch off of this, so would make a very messy PR until this is merged).

Thanks for the feedback and encouragement! :)

theGreatWhiteShark:
> Home and End should be covered by the standard Qt mappings of QKeySequence::MoveToStartOfLine and QKeySequence::MoveToEndOfLine, ie. move to the time beginning and ends of the current pattern / instrument line.

Ah, I see. Haven't got accustomed to the cursor yet. I was more referring to the transport position. So, how about a keyboard shortcut to start to relocate to the position the cursor is located and start playback there. Also while at it: how about a keyboard shortcut to switch to the pattern in the PatternEditor the cursor is located in the SongEditor at. More or less using it's vertical and horizontal position for selection/navigation.

When pressing the former shortcut while the cursor is focused on the PatternEditor or NotePR the playback mode could change to Pattern mode and start playback at that particular Note.

> Having already folded most of the functionality of the drawing mode into the select mode (indeed all the functionality of it that existed before 1.0)

I might be missing something here. But when making the continuous drawing the default interaction of the select mode and binding the drag selection to Ctrl+left mouse key both modes would have been merged (given that the behavior of mouse press and Ctrl+mouse press differs whether a selected (group of) note(s) as clicked or not).

cme:
> Also while at it: how about a keyboard shortcut to switch to the pattern in the PatternEditor the cursor is located in the SongEditor at. More or less using it's vertical and horizontal position for selection/navigation.
> 
> When pressing the former shortcut while the cursor is focused on the PatternEditor or NotePR the playback mode could change to Pattern mode and start playback at that particular Note.

Interesting, that would be handy! Any ideas on what would seem like an appropriate key combination to use?

> > Having already folded most of the functionality of the drawing mode into the select mode (indeed all the functionality of it that existed before 1.0)
> 
> I might be missing something here. But when making the continuous drawing the default interaction of the select mode and binding the drag selection to Ctrl+left mouse key both modes would have been merged (given that the behavior of mouse press and Ctrl+mouse press differs whether a selected (group of) note(s) as clicked or not).

Sorry, I don't mean that the exact behaviour is the same, just that pre-1.0 the 'draw' mode would only set/clear active cells, and that this can now be done from the select mode as well.

theGreatWhiteShark:
> Any ideas on what would seem like an appropriate key combination to use?

How about  `ctrl`+`space`?

theGreatWhiteShark:
Another note (should I put them here or in a separate issue?)

In the PatternEditor I can delete a selected group via `del` but not an individual note using the cursor + `del`.

cme:
> In the PatternEditor I can delete a selected group via `del` but not an individual note using the cursor + `del`.

Oh, nice! Yes, delete should definitely delete a note under the keyboard cursor. Will add!

> Another note (should I put them here or in a separate issue?)

Here seems as good a place as any, I still have it on notifications and it feels wrong to pollute the issue tracker with these. 

Thanks a ton for trialing this and reporting! :)

theGreatWhiteShark:
Some more:

- I found the copy-cut-paste behavior using keyboard unexpected. What would expect is that when pasting the current cursor position or the top left corner of the mouse cursor would become the top left corner of the selected rectangle. But instead it's the bottom right corner. Is this the standard way of doing it (come to think of it, I might already live in my tiling window manager world for too long)
- paste/cut/copy does not work in the SongEditor yet (but I guess you are fully aware of this)
- When editing in the `NotePropertyRuler`/`PatternEditor` using keyboard and selecting a different Instrument/Pattern in the `PatternEditor`/`SongEditor` the focus stays in the latter. While this is totally plausible from the perspective of the code I was a bit confused at first when using Hydrogen. After all, clicking on a different pattern has no impact on the SongEditor at all. How about not changing focus when clicking on the side pannels?

> Thanks a ton for trialing this and reporting! :)

No problem. Thanks for writing all that code :wink:

cme:
> * I found the copy-cut-paste behavior using keyboard unexpected. What would expect is that when pasting the current cursor position or the top left corner of the mouse cursor would become the top left corner of the selected rectangle. But instead it's the bottom right corner. Is this the standard way of doing it (come to think of it, I might already live in my tiling window manager world for too long)

Hmm, this also sounds like a reasonable expectation to me, and I think it's the way spreadsheets typically behave. (I'm assuming by 'top left' and 'bottom right' you mean the locations of where the selection was started and where the selection was finished?).

I think the difference is made by the way the selection is defined and expanded. In the common selection manager (at the moment) the cursor is moved when expanding the selection, which is defined between the start point, and the current cursor location. In a quick survey of a few spreadsheets, the cursor stays where it is, and the opposite corner of the selection box is moved.

Does that sound like the sort of behaviour you'd expect? 

Would need a little bit of restructuring in the selection manager and the keypress handlers since all the movement is currently based on the cursor position, but not too drastic.

> * paste/cut/copy does not work in the SongEditor yet (but I guess you are fully aware of this)

Yep, see #1079 :)

> * When editing in the `NotePropertyRuler`/`PatternEditor` using keyboard and selecting a different Instrument/Pattern in the `PatternEditor`/`SongEditor` the focus stays in the latter. While this is totally plausible from the perspective of the code I was a bit confused at first when using Hydrogen. After all, clicking on a different pattern has no impact on the SongEditor at all. How about not changing focus when clicking on the side pannels?

Easy enough to do, I'll try it!

Hmm, increasingly thinking of creating a project board now. :D

Thanks!

theGreatWhiteShark:
> I'm assuming by 'top left' and 'bottom right' you mean the locations of where the selection was started and where the selection was finished?

Exactly. Which are of course not the correct terms since they assume right-handedness as default.

But for the point of insertion I do refer to the top left pixel - or the tip - of the cursor.

>  In the common selection manager (at the moment) the cursor is moved when expanding the selection, which is defined between the start point, and the current cursor location. In a quick survey of a few spreadsheets, the cursor stays where it is, and the opposite corner of the selection box is moved.

I do like the current way of selection via keyboard. Moving the cursor while expanding/shrinking the selection does feel like text editing and very familiar. It's just the pasting where I expected the first grid cell the selection was started at being insert at the cursor when pressing Crtl+v.

> Hmm, increasingly thinking of creating a project board now. :D

Yeah. Maybe even an associated Discussion thread to have more user feedback. Come to think of it it's not particular low-threshold to comment a PR for people who haven't contributed yet :D

theGreatWhiteShark:
Also I wonder whether hiding the cursor should be the default behavior. People already familiar with Hydrogen might miss out on that feature because they don't feel the need for keyboard interaction. On the other hand, turning it off if it's annoying is a rather simple thing to do.

cme:
> > In the common selection manager (at the moment) the cursor is moved when expanding the selection, which is defined between the start point, and the current cursor location. In a quick survey of a few spreadsheets, the cursor stays where it is, and the opposite corner of the selection box is moved.
> 
> I do like the current way of selection via keyboard. Moving the cursor while expanding/shrinking the selection does feel like text editing and very familiar. It's just the pasting where I expected the first grid cell the selection was started at being insert at the cursor when pressing Crtl+v.

I think I've got a little confused reading your description (because now we have "cursor" that can be *keyboard cursor* or *mouse cursor*) :D Currently, copying and pasting with *keyboard* movement:

  * Start selection at, say, top left.
  * Final keyboard cursor position is bottom right.
  * Copy
  * Move keyboard cursor elsewhere
  * Paste: pasted selection appears above and to the left of current keyboard cursor
      * should it? Or should the selection be pasted to the right and below the keyboard cursor?

With mouse movements:

  * Start selection drag from top left
  * Release mouse at bottom right
  * Ctrl+C (or Right-click + copy)
  * Move elsewhere
  * Right click + paste: pasted selection appears above and to the left of click point.
     * Should it? Or should the selection be pasted to the right and below the mouse cursor click point?

Having dug a bit more into spreadsheet behaviour, they seem to *always* paste such that the top left (not even the 'start point' of the selection as I'd assumed, just **always** 'top left').

This doesn't make it clear what should be done in cases where a selection has been made in multiple stages by using Drag and then Ctrl+Drag. When I tried this with LibreOffice and Excel, both refused to copy cells that were selected this way :D I think a sensible behaviour in this circumstance, consistent with the spreadsheet model, would be to use the top left point of the most *recent* selection action as the 'origin' point of the selection.

cme:
> Also I wonder whether hiding the cursor should be the default behavior. People already familiar with Hydrogen might miss out on that feature because they don't feel the need for keyboard interaction. On the other hand, turning it off if it's annoying is a rather simple thing to do.

@mauser and @trebmuh were concerned the appearance of the keyboard cursor might throw off existing users, and hiding it unless keyboard navigation keys (which new users will probably instinctively press at some point!) are pressed seemed a good compromise between discoverability and avoiding confusing people. But I think you're right that that leaves old users who already believe the keyboard doesn't do anything without a way to discover it!

I think this is probably mitigated enough by mentioning it prominently in the release notes.

I'm not invested either way, but perhaps this is another potential point I should mention in a Discussion topic :)

theGreatWhiteShark:
> Or should the selection be pasted to the right and below the keyboard cursor?
> Or should the selection be pasted to the right and below the mouse cursor click point?

Those would be the things I would expect.

> This doesn't make it clear what should be done in cases where a selection has been made in multiple stages by using Drag and then Ctrl+Drag. When I tried this with LibreOffice and Excel, both refused to copy cells that were selected this way :D I think a sensible behaviour in this circumstance, consistent with the spreadsheet model, would be to use the top left point of the most recent selection action as the 'origin' point of the selection.

Hmm. But this does makes the pasting depending on the order of selection. I think I would go for the most top left point of the whole selection. 

For example imagine one has such a sophisticated pattern as the one below and is used to copy the whole thing over and over again. Occasionally, the note at bottom right is missed during selection and the user decides to include it with Ctrl+Drag

first_selection

Using the top left point of the most recent selection action as the 'origin' point of the selection would alter the horizontal and vertical offset of the pasted result significantly.

theGreatWhiteShark:
> I think this is probably mitigated enough by mentioning it prominently in the release notes.

+1

cme:
> > Or should the selection be pasted to the right and below the keyboard cursor?
> > Or should the selection be pasted to the right and below the mouse cursor click point?
> 
> Those would be the things I would expect.

Great. Implemented in #1079 , which I'll merge soon.

> > This doesn't make it clear what should be done in cases where a selection has been made in multiple stages by using Drag and then Ctrl+Drag. When I tried this with LibreOffice and Excel, both refused to copy cells that were selected this way :D I think a sensible behaviour in this circumstance, consistent with the spreadsheet model, would be to use the top left point of the most recent selection action as the 'origin' point of the selection.
> [...]
> Using the top left point of the most recent selection action as the 'origin' point of the selection would alter the horizontal and vertical offset of the pasted result significantly.

Thanks, this is a great example and really helps make it clear. Definitely feels "top right" is the right move. It also removes any dependency on where the drag rectangle was actually created, which helps if you drag a larger area that doesn't have any notes.

theGreatWhiteShark:
> >    In the common selection manager (at the moment) the cursor is moved when expanding the selection, which is defined between the start point, and the current cursor location. In a quick survey of a few spreadsheets, the cursor stays where it is, and the opposite corner of the selection box is moved.
>
> I do like the current way of selection via keyboard. Moving the cursor while expanding/shrinking the selection does feel like text editing and very familiar. It's just the pasting where I expected the first grid cell the selection was started at being insert at the cursor when pressing Crtl+v.

After using the new keyboard selection for a couple of days I think I was wrong on this one. One get used to it very quickly but it still feels a little bit weird. Turns out in text editing the cursor does not move either. But pressing [left arrow] or [right arrow] moves the cursor to the left or to the right of the selection. The latter gave me the impression of a moving cursor while selecting.

theGreatWhiteShark:
> So, how about a keyboard shortcut to start to relocate to the position the cursor is located and start playback there. Also while at it: how about a keyboard shortcut to switch to the pattern in the PatternEditor the cursor is located in the SongEditor at. More or less using it's vertical and horizontal position for selection/navigation.

BTW, I'm currently working on this one.

Issue #958, opened 2020-10-19
sample leakage


theGreatWhiteShark:
Hydrogen version *          : b51508b3d6267df9c36e0e8fe0ae7de58f5fd972
Operating system + version  :  Linux
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

If I open Hydrogen and change the drumkit from e.g. GMRock to BigMono I get

```
Objects map :                        class	constr   destr   alive
	[            AudioEngineInfoForm ]	     1	     1	     0
	[                       Director ]	     1	     1	     0
	[             FilesystemInfoForm ]	     1	     1	     0
	[              SimpleHTMLBrowser ]	     1	     1	     0
	[                    HydrogenApp ]	     1	     1	     0
	[               InstrumentEditor ]	     1	     1	     0
	[          InstrumentEditorPanel ]	     1	     1	     0
	[                   LayerPreview ]	     1	     1	     0
	[                    WaveDisplay ]	     2	     2	     0
	[                 InstrumentRack ]	     1	     1	     0
	[             LadspaFXProperties ]	     4	     4	     0
	[                       MainForm ]	     1	     1	     0
	[                          Mixer ]	     1	     1	     0
	[              DrumPatternEditor ]	     1	     1	     0
	[            NotePropertiesRuler ]	     5	     5	     0
	[    PatternEditorInstrumentList ]	     1	     1	     0
	[             PatternEditorPanel ]	     1	     1	     0
	[             PatternEditorRuler ]	     1	     1	     0
	[                PianoRollEditor ]	     1	     1	     0
	[                  PlayerControl ]	     1	     1	     0
	[                MetronomeWidget ]	     1	     1	     0
	[                 PlaylistDialog ]	     1	     1	     0
	[                     SongEditor ]	     1	     1	     0
	[          SongEditorPatternList ]	     1	     1	     0
	[        SongEditorPositionRuler ]	     1	     1	     0
	[                SongEditorPanel ]	     1	     1	     0
	[           SoundLibraryDatabase ]	     1	     1	     0
	[               SoundLibraryInfo ]	     1	     1	     0
	[              SoundLibraryPanel ]	     1	     1	     0
	[               SoundLibraryTree ]	     1	     1	     0
	[                   SplashScreen ]	     1	     1	     0
	[             AutomationPathView ]	     1	     1	     0
[...]

PR #968, opened 2020-11-01
using shared_ptr for Sample and minor fixes

theGreatWhiteShark:
In the wake of #958 I rewrote some code to only use `Sample`s as shared pointers in order to prevent any leakage.

### DIFF
--- a/src/gui/src/SampleEditor/MainSampleWaveDisplay.cpp
+++ b/src/gui/src/SampleEditor/MainSampleWaveDisplay.cpp
@@ -154,7 +154,7 @@ void MainSampleWaveDisplay::updateDisplayPointer()
 void MainSampleWaveDisplay::updateDisplay( const QString& filename )
 {
 
-	Sample *pNewSample = Sample::load( filename );
+	auto pNewSample = Sample::load( filename );
 	
 	if ( pNewSample ) {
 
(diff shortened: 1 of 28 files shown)


src/gui/src/SampleEditor/MainSampleWaveDisplay.cpp
```diff
@@ -167,8 +167,8 @@ void MainSampleWaveDisplay::updateDisplay( const QString& filename )
 
 		float fGain = height() / 4.0 * 1.0;
 
-		float *pSampleDatal = pNewSample->get_data_l();
-		float *pSampleDatar = pNewSample->get_data_r();
+		auto pSampleDatal = pNewSample->get_data_l();
```
mauser, line 170: Is this intentional? There are also some other places where float * is replaced by auto. I was unsure if this might be an effect from auto-replacing text.

mauser (approved):

src/gui/src/SampleEditor/MainSampleWaveDisplay.cpp
theGreatWhiteShark: Yeah. Somewhat. I toyed with having the sample data being represented as `std::vector` instead of a raw float arrays. I thought it would might be handy to be able to use the algorithms of the standard library on it and also to not have any raw audio leaked at some point in the future.

But it turned out copying takes about 4x as long for mono and 1.5x for stereo files. That was more than I expected. So I dropped these changes but left all the `auto` in place cause they don't hurt and it would make revisiting those changes faster.

PR #969, opened 2020-11-02
fix #960 responsible OSC settings

### LINKED ISSUE #960: Toggling OSC support in Preferences requires restart
Hydrogen version *          : b51508b3d6267df9c36e0e8fe0ae7de58f5fd972
Operating system + version  : 
Audio driver + version      : 

Ideally, this should work without restart. Or at least with a pop up window informing the user about this requirement.

theGreatWhiteShark:
Both the enabling and the port of the OSC server can now be altered (and take effect) without restarting Hydrogen.

To do so I made the following changes:

- introduce the `Hydrogen::toggleOscServer()` and `Hydrogen::recreateOscServer()` functions to allow for the two behaviors mentioned above
- split `OscServer::start()` into `::init()` and `::start()` as well as introducing a `::stop()` function
- make both the `OscServer::handleAction()` and `OscServer::m_pClientRegistry` members of the class. They handling would become less straight forward ones it was possible to recreate the `OscServer`
- make the corresponding `PreferencesDialog` section call the functions listed in the first bullet point

### DIFF
--- a/src/gui/src/PreferencesDialog.cpp
+++ b/src/gui/src/PreferencesDialog.cpp
@@ -505,9 +505,15 @@ void PreferencesDialog::on_okBtn_clicked()
 	pPref->m_nMidiChannelFilter = midiPortChannelComboBox->currentIndex() - 1;
 
 	//OSC tab
-	pPref->setOscServerEnabled( enableOscCheckbox->isChecked() );
+	if ( enableOscCheckbox->isChecked() != pPref->getOscServerEnabled() ) {
+		pPref->setOscServerEnabled( enableOscCheckbox->isChecked() );
+		H2Core::Hydrogen::get_instance()->toggleOscServer( enableOscCheckbox->isChecked() );
+	}
 	pPref->setOscFeedbackEnabled( enableOscFeedbackCheckbox->isChecked() );
-	pPref->setOscServerPort( incomingOscPortSpinBox->value() );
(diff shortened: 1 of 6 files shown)


src/gui/src/PreferencesDialog.cpp
```diff
@@ -505,9 +505,15 @@ void PreferencesDialog::on_okBtn_clicked()
 	pPref->m_nMidiChannelFilter = midiPortChannelComboBox->currentIndex() - 1;
 
 	//OSC tab
-	pPref->setOscServerEnabled( enableOscCheckbox->isChecked() );
+	if ( enableOscCheckbox->isChecked() != pPref->getOscServerEnabled() ) {
+		pPref->setOscServerEnabled( enableOscCheckbox->isChecked() );
+		H2Core::Hydrogen::get_instance()->toggleOscServer( enableOscCheckbox->isChecked() );
```
cme, line 510: I just updated and found that these need to be guarded with `#ifdef`s. Trivial fix so I pushed it, I hope you don't mind! :)

mauser: @cme: It seems that both of us had the same idea there b78a54059f595cbb15dfca26dad4cda55709cb3e :smile:

cme: At almost *exactly* the same time! Uncanny1 :D

PR #979, opened 2020-11-14
Set midiOutChannel to standard 10 by default

BrunoVernay:
Setting to 10 by default allow MIDI TX to work by default.
(It would be nice if there was a global parameter instead of having to fix each instrument in each DrumKit.)


theGreatWhiteShark:
Hey @BrunoVernay ,

Thank's for your patch! The `-1` is currently the only way for the user to disable MIDI output (apart from miss-configuring the MIDI setup). We still need another means to provide this feature (see #2113) and then I will merge your PR.

Issue #971, opened 2020-11-04
drum pattern plays short when in song playback


snakedad:
Hydrogen version *          :
Operating system + version  : 
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

< describe your issue or feature request here ** >

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

trebmuh:
@snakedad we need more information. Please update you initial message.

snakedad:
one of my tracks will begin playing a pattern (that plays perfectly in pattern mode) and when it reaches the second beat restarts the pattern (when playing in song mode).  So in essence, a 4 measure pattern plays 3.25 measures.  I thought it was a BPM change issue, because it occurred later in the song just before a BPM change, but it also happens earlier in the song when no change is present.

trebmuh:
Please, feel the asked info in the your first message (hydrogen version, operating system + version, audio driver + version).

cme:
@snakedad We'll need to reproduce this in order to understand it. Is there any way that you could share the 2hsong file with us by attaching it here?

If you don't want to share your actual beats, you could maybe try taking a copy of it, and stripping out notes until the bare minimum that reproduces the issue is left?

Thanks

snakedad:
I think I may have just figured it out.  I had some fill and cymbal tag patterns occurring simultaneously that were a different measure count (ie: only 1 measure since it was just a cymbal tag or a short fill), and it seems like the program was trying to satisfy both durations of patterns.  instead of playing the short pattern aligned to the longer pattern, it shortened the longer pattern.

i corrected it by making ALL patterns the same number of beats and measures.

trebmuh:
It is a known limitation of Hydrogen.

cme:
The Song Editor GUI probably ought to provide some feedback on that, at least.

trebmuh:
#691 would probably be something great to be worked on.
@cme: do you know someone with coding skills, hydrogen interest, and ability to work with Hydrogen current codebase?

_(/me whistle :eyes: )_

cme:
> #691 would probably be something great to be worked on.

[...]

Issue #691, opened 2019-01-24
Patterns of different length in the same song


gutte:
Hydrogen version *          : 0.9.7
Operating system + version  : Debian 9.6
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Hi,

I have a scenario where I have a 1 bar (size 8) groove and a 2 bar (size 16)ride pattern (the ride should play out over two repetitions of the groove).

Groove:

![image](https://user-images.githubusercontent.com/767995/51697051-96510080-200f-11e9-9ad5-974d52cfbeb1.png)

Ride:

![image](https://user-images.githubusercontent.com/767995/51697073-a7017680-200f-11e9-9ba5-2cd071cd892c.png)

Now in **song mode**:
When i play the ride only it plays the whole pattern over the 1 "bar" seen in the song editor (size 16 -> 1 song bar):

![image](https://user-images.githubusercontent.com/767995/51697203-f0ea5c80-200f-11e9-9a14-4c7882a1bced.png)

However, when I add the groove, size 8 maps to 1 "bar" of the song:

![image](https://user-images.githubusercontent.com/767995/51697429-73731c00-2010-11e9-93a0-809dcda08766.png)

This is ok i suppose, BUT the ride then starts repeating after the first bar, and there is no way to get the whole ride pattern into the song.

I would expect to be able to do this to hear the whole ride:

![image](https://user-images.githubusercontent.com/767995/51697504-a1586080-2010-11e9-81ec-7cc7bb21ceb1.png)

The only solution I can find is to double the length of the groove, which may be simple enough in my case, but would be very tedious if I had many patterns that were shorter than my longest one.

Is there something I have missed?

In my view the underlying problem is that discussed in [issue 257](https://github.com/hydrogen-music/hydrogen/issues/257): **The size of the pattern should be defined unambiguously in bars, and those bars should be a 1:1 with the bars in the song editor.**

However, this issue could also be fixed independently by allowing for a longer pattern to continue for multiple bars in the song.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

thijz:
[...]

PR #984, opened 2020-11-16
Show pattern length mismatches visually in Song Editor

cme:
Addresses #971 and #691 by
  1. defining the length of a song / sequencer slot as the length of the *longest* pattern
  2. showing any pattern shorter than the longest pattern using a narrower rectangle, whose right edge corresponds to the position of the playhead at the end of that pattern

image


mauser:
This is a great solution for that problem! At least it seems to be very natural and understandable for me.

mauser (approved):

cme:
I'll merge in a while unless anyone has objections, and close #971. I'll leave #691 open as mixing patterns of different length is still painful and I think there's more that can be done longer-term.

trebmuh:
Looks nice!

PR #988, opened 2020-11-22
Add AudioEngine lock checks to PatternList

cme:


src/core/AudioEngine.h
```diff
@@ -284,11 +296,65 @@ class AudioEngine : public H2Core::Object
...
+};
+
+
 inline float AudioEngine::getElapsedTime() const {
 	return m_fElapsedTime;
 }
-	
-};
 
+inline void AudioEngine::assertLocked( ) {
```
mauser, line 352: What about renaming this to `assertAudioEngineLocked()` ? It sounds overflous in this class definition, but when used in a derived class, it might make the intention about what is beeing locked more clear

mauser (approved):

src/core/AudioEngine.h
cme: Hmm. Yes, agreed. It's probably also good to have the method name different from that in AudioEngine.

PR #998, opened 2020-11-29
allow read-only access of songs

### LINKED ISSUE #996: Open demo broken
Hydrogen version *          : GIT master b17f662
Operating system + version  : LibraZiK-3 (Debian Buster based)
Audio driver + version      : jack (but not relevant for this issue)

Fresh build to check the NSM support (thanks @theGreatWhiteShark & @mauser !).
Trying to load a demo, I've got this popup window:

errorloadingsong

and starting fromp the CLI, I've got this error log:

```
(E) CoreActionController::isSongPathValid Error: Unable to handle path [/usr/share/hydrogen/data/demo_songs/GM_kit_Diddley.h2song]. You must have permissions to write the file!
```

Please note that if I' copy/paste the demo in a location where I have written permission (e.g. $HOME), H2 loads it fine.

From a user point of view, it shouldn't be needed to have the write permission to open and play a demo.

_(I'll go testing the NSM support now! :sunglasses: )_

theGreatWhiteShark:
Fixes #996 

Allows for accessing read-only songs (was broken with the introduction of a uniform loading function) and displays a warning informing the user to run "Save as" in order to enable the autosave again.

### DIFF
--- a/src/core/CoreActionController.cpp
+++ b/src/core/CoreActionController.cpp
@@ -388,19 +388,13 @@ bool CoreActionController::openSong( const QString& sSongPath ) {
 		return false;
 	}
 	
-	QFileInfo songFileInfo = QFileInfo( sSongPath );
-	if ( !songFileInfo.exists() ) {
-		ERRORLOG( QString( "Selected song [%1] does not exist." )
-				 .arg( sSongPath ) );
-		return false;
-	}
-	
 	// Create an empty Song.
 	auto pSong = Song::load( sSongPath );
 
 	if ( pSong == nullptr ) {
 		ERRORLOG( QString( "Unable to open song [%1]." )
 				  .arg( sSongPath ) );
+		return false;
 	}
 	
 	return setSong( pSong );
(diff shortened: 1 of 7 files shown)


theGreatWhiteShark:
I also made `SongWriter::writeSong` check whether the Song is actually writable to the supplied location. This reduces the error message from (3 lines)

> (W) SongWriter::writeSong File save reported an error.
(I) SongWriter::writeSong Saving song /usr/local/share/hydrogen/data/demo_songs/GM_kit_demo2.autosave.h2song
QIODevice::write (QFile, "/usr/local/share/hydrogen/data/demo_songs/GM_kit_demo2.autosave.h2song"): device not open

to (1 line)

> (E) SongWriter::writeSong Unable to save song to [/usr/local/share/hydrogen/data/demo_songs/GM_kit_Diddley.autosave.h2song]. Path is not writable!

theGreatWhiteShark:
In addition, I introduced some tests to check whether the permission checking in `Filesystem` is working. This is motivated since the behavior of the Qt implementation might change in the future and is platform specific (see https://doc.qt.io/qt-5/qfiledevice.html#Permission-enum). 

@mauser @cme @trebmuh  Unfortunately I only have access to Linux systems. Could you give me a path with 1) only read access and 2) nor read or write access which is present on all Windows and Mac systems?

trebmuh:
Same for me, I only do Linux System.

cme:
For Mac OS, `/etc/master.passwd` in inaccessible for non-root, and `/etc/profile` should be the same as on Linux.

Windows... who knows? I'll spin up a VM and have a look for candidates :D

theGreatWhiteShark:
@cme thanks a lot!

cme:
So, it looks like there isn't anything on Windows that's both consistently named and unreadable. And I don't trust Windows enough to say with any confidence that things that look unwritable really are. 

:/

theGreatWhiteShark:
Okay. Thanks for researching! I think we can just leave at that. Permission checks are turned of for NTFS in Qt5 due to performance issues anyway and I can't recall an issue in which a user reported such problems running Windows.

Issue #918, opened 2020-08-06
jack-session has been marked as deprecated


trebmuh:
See: https://jackaudio.org/news/2020/07/16/jack-session-has-been-marked-as-deprecated.html

Not being a coder, I don't know what that implies for Hydrogen. NSM support is already there. Maybe Hydrogen should drop jack-session support?

mauser:
Ah, that "New Session Management" discussion :) Yes, i suppose we can drop support for jack-session then.

theGreatWhiteShark:
I would suggest to just remove the support during compilation by disabling the `H2CORE_HAVE_JACKSESSION` flag. If no one complains, we can drop the code in a couple of releases.

> NSM support is already there

Well, it's still too rough around the edges to really use Hydrogen with NSM. The new fork (so glad this did happen!) does probably uses `JACKPatch` as well. This will prevent all wiring of the per-instrument JACK ports of H2 to persist between sessions. In addition H2 is not fully compliant with the API yet. But most of those things are already done in  #792 and I will soon start with the finishing touches ;)

trebmuh:
Sounds clever.

PR #1004, opened 2020-12-05
deprecate Jack-session

theGreatWhiteShark:
all source code is still in place. By disabling the corresponding flag during configuration it just will not be built anymore.

If users request for supporting `Jack-session` even after its deprecation, we can just enable its build again.

Addresses #918

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,9 +51,7 @@ OPTION(WANT_LADSPA          "Enable use of LADSPA plugins" ON)
 
 IF(APPLE)
 	OPTION(WANT_OSC  "Enable OSC support" OFF)
-	OPTION(WANT_JACKSESSION "Enable use of Jack-Session-Handler" OFF)
 ELSE()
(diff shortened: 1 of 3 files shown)


trebmuh:
Good to see that happening now there is a better NSM support in H2. :+1:

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ IF(APPLE)
 	OPTION(WANT_OSC  "Enable OSC support" OFF)
 	OPTION(WANT_JACKSESSION "Enable use of Jack-Session-Handler" OFF)
 ELSE()
-	OPTION(WANT_JACKSESSION "Enable use of Jack-Session-Handler" ON)
+	OPTION(WANT_JACKSESSION "Enable use of Jack-Session-Handler" OFF)
```
mauser, line 56: Should we remove the option completely? This will still be shown in cmake frontends like "ccmake"

theGreatWhiteShark: Sure. Haven't thought about that one.

trebmuh:
:+1:

Issue #1002, opened 2020-12-02
memory leakage


theGreatWhiteShark:
Hydrogen version *          : bc94f858c22a0a70ddf65a3d2ea37598973abd5a
Operating system + version  :  Linux
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

> Objects map :                        class	constr   destr   alive
	[            AudioEngineInfoForm ]	     1	     1	     0
	[                       Director ]	     1	     1	     0
	[             FilesystemInfoForm ]	     1	     1	     0
	[              SimpleHTMLBrowser ]	     1	     1	     0
	[                    HydrogenApp ]	     1	     1	     0
	[               InstrumentEditor ]	     1	     1	     0
	[          InstrumentEditorPanel ]	     1	     1	     0
	[                   LayerPreview ]	     1	     1	     0
	[                    WaveDisplay ]	     2	     2	     0
	[                 InstrumentRack ]	     1	     1	     0
	[             LadspaFXProperties ]	     4	     4	     0
	[                       MainForm ]	     1	     1	     0
	[                          Mixer ]	     1	     1	     0
	[              DrumPatternEditor ]	     1	     1	     0
	[            NotePropertiesRuler ]	     5	     5	     0
	[    PatternEditorInstrumentList ]	     1	     1	     0
	[             PatternEditorPanel ]	     1	     1	     0
	[             PatternEditorRuler ]	     1	     1	     0
	[                PianoRollEditor ]	     1	     1	     0
	[                  PlayerControl ]	     1	     1	     0
	[                MetronomeWidget ]	     1	     1	     0
	[                 PlaylistDialog ]	     1	     1	     0
	[                     SongEditor ]	     1	     1	     0
	[          SongEditorPatternList ]	     1	     1	     0
	[        SongEditorPositionRuler ]	     1	     1	     0
	[                SongEditorPanel ]	     1	     1	     0
	[           SoundLibraryDatabase ]	     1	     1	     0
	[               SoundLibraryInfo ]	     1	     1	     0
	[              SoundLibraryPanel ]	     1	     1	     0
	[               SoundLibraryTree ]	     1	     1	     0
	[                   SplashScreen ]	     1	     1	     0
	[             AutomationPathView ]	     1	     1	     0
	[                         Button ]	   213	   213	     0
	[                  CpuLoadWidget ]	     1	     1	     0
[...]

PR #1006, opened 2020-12-06
fix memory leakage in when upgrading drumkit

theGreatWhiteShark:
The code for creating a new `Drumkit` and loading the samples is run twice in `Drumkit::load_file` if the there are problems in reading the drumkit and it is a more recent one (not legacy).

addresses #1002

### DIFF
--- a/src/core/Basics/Drumkit.cpp
+++ b/src/core/Basics/Drumkit.cpp
@@ -109,6 +109,8 @@ Drumkit* Drumkit::load( const QString& dk_dir, const bool load_samples )
 
 Drumkit* Drumkit::load_file( const QString& dk_path, const bool load_samples )
 {
+	bool bReadingSuccessful = true;
+	
 	XMLDoc doc;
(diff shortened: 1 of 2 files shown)


mauser:
Hi! 

good finding! I'm surprised that we haven't found this one earlier :-/ I can also reproduce it with 1.0, and the code seems to be in there since April. Maybe it is because it just brings up problems when the drumkit is upgraded?

theGreatWhiteShark:
Probably.

With b24ca5c2c6302cb87835436c1a6e4cf74b530a4b also the most recent drumkits (including the GMRockKit we ship) are not valid anymore and require upgrading. Beforehand we might not have had this problem.

mauser:
Oops. I'm going to fix those drumkit files :)

theGreatWhiteShark:
https://github.com/hydrogen-music/hydrogen/pull/1006/commits/d3de430807286a0c54d2ca9008f8cf8da305daec does fix memory leakage introduced by the command line drumkit.

This one really called for using shared pointers for the `Drumkit`.

PR #1008, opened 2020-12-06
Audioengine refactoring

mauser:
Currently, there are a lot of static audioengine_* methods in the Hydrogen class. The Hydrogen class and those static methods are communicating via global variables. In addition, there is a separate AudioEngine class. 

This "architecture" is a big mess and needs a cleanup:

- [x] Move all audioengine_* method to the `AudioEngine` class. 
- [x] rename all `AudioEngine` members according to naming scheme
- [x]  make member variables private & introduce getters/setters
- [X]  introduce a member `m_pAudioEngine` in Hydrogen class-
- [x] rename instances of `Hydrogen` from "pEngine" to "pHydrogen"


theGreatWhiteShark:
I would also suggest to make a search-and-replace from `pEngine` to `pHydrogen` in the GUI. Having the `Hydrogen` class referred to by the former will be even more confusing after the refactoring

mauser:
@theGreatWhiteShark : I've re-targeted the pull request to "development" and merged in the current state of the master branch. Is there anything on master which should not go into "development"? Haven't thought about that :) 

There have been also some thing which we had planned for this pull request before (like constness and naming of Hydrogen class members), which i would like to leave untouched for now to minimize the merge problems master <-> development. Those can be adressed after the release..

theGreatWhiteShark:
> I've re-targeted the pull request to "development" and merged in the current state of the master branch. Is there anything on master which should not go into "development"? Haven't thought about that :)

No. AFAIU our branching model `development` can contain commits of `master` but not the other way around.

> There have been also some thing which we had planned for this pull request before (like constness and naming of Hydrogen class members), which i would like to leave untouched for now to minimize the merge problems master <-> development. Those can be adressed after the release..

+1 . Sounds good. I will have a look during the weekend.

src/core/AudioEngine.cpp
```diff
@@ -35,37 +69,96 @@ namespace H2Core
...
 	Effects::create_instance();
 #endif
 
 }
 
-
-
 AudioEngine::~AudioEngine()
 {
 	INFOLOG( "DESTROY" );
```
theGreatWhiteShark, line 164: How about checking for `! m_songNoteQueue.empty()`, `m_pPlayingPatterns != nullptr`, `m_pNextPatterns != nullptr`, and `m_pMetronomeInstrument != nullptr` in here and calling `destroy()` in case one of these conditions is false. This way we can ensure everything is clean after `AudioEngine` is destroyed regardless of the changes that might be applied to `Hydrogen::~Hydrogen` in the future.

theGreatWhiteShark (approved):
Looks good and everything seems to work :+1:

Issue #948, opened 2020-09-30
Mark L and R on panning editor


Porrumentzio:
Currently it is confusing because there is no way further than trying out to know which side is left and which right.

Indicating it with L and R would solve this problem

trebmuh:
Not sure what window/space you're referring to. The only thing I can think of already displays a L/R:

![hydro-pan](https://user-images.githubusercontent.com/8705846/94698098-4294cf80-0339-11eb-83ad-39e07560511a.png)

Porrumentzio:
On the pattern editor, below the tracks/instruments.
There is a editing space where every note can be edited, changing velocity, probabili

Porrumentzio:
Sorry for closing, it was a mistake.

The space I say is in the bottom side of the pattern editor, where the velocity, probability... of each note can be edited. And also the panoramization

oddtime:
![screenshot-b95f1a3f](https://user-images.githubusercontent.com/54219881/101774470-6ff6ba00-3aee-11eb-8ea3-89b535d63528.jpg)
 A message similar to this (appearing while changing note Pan) could be useful and easy to program

oddtime:
Opened pull request #1015 for the message

oddtime:
added R L in Pull Req. But they scroll

theGreatWhiteShark:
The note property display can use some scale in general. For *velocity* and *probability* having e.g. a 0, 0.5, and 1 wouldn't hurt either. For the *lead and lag* displaying the amount of milliseconds would be nice too.

oddtime:
> For the lead and lag displaying the amount of milliseconds would be nice too.

Is it a fixed time value or depends on bpm?

theGreatWhiteShark:
> Is it a fixed time value or depends on bpm?
From the current InfoMessage on the LCD it seems it has steps of 0.1 ticks so showing the scale in ms could be tricking

It's given in ticks (probably multiplied by some constant) and depends on the bpm. So, you are right. Displaying milliseconds would require to redraw the widget when passing a tempo marker. This does not seem to be a good idea. Maybe using something like "fraction of a quarter" would be more appropriate instead.

theGreatWhiteShark:
Implemented in #1544. It does not added a "L" and "R" indicator as this did prove difficult/not working in this scrollable widget. But on all changes of the panning you get now a status message telling whether the note is panned to the left or right

PR #1015, opened 2020-12-10
Adding "Set note pan" message (Note Properties Ruler)

oddtime:
pan
This message was missing
Allows to know if pan is Right or Left (since there's no indication on the note editor ruler) and the value.
Minimal solution to issue #948 

made some messages translatable
reduced size of buffer strings


trebmuh:
LGTM.
Is there any chance for the "R"/"L" to be translatable?

oddtime:
Not now but I can do it

trebmuh:
It'd be awesome.

oddtime:
screenshot-0230ec97
See #948
Unfortunately the text is fixed on background (it scrolls with it) but for now we can live with it.

oddtime:
I have changed the code to input pan from GUI calling the pan law methods (the ones in Sampler class). This solves #1167.
Maybe a factorization here would be useful (similar lines are coded in both "mouse move" and "mouse wheel" functions...), I would leave this as it concerns all the note properties if I am not wrong.

oddtime:
Actually I am doubting the necessity of still having `pan_l` and `pan_r` in `Note` and `Instrument` classes, I opened an issue: #1242
I am tempted to deprecate them and get a one to one correspondence between the widgets values and the new unique `pan` member

PR #1043, opened 2020-12-26
Rebuild libsndfile for compatability

cme:
Rebuild libsndfile for compatability

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -50,6 +50,7 @@ for:
       brew install qt5; export CMAKE_PREFIX_PATH="$(brew --prefix qt5)";
       brew install libarchive; export PKG_CONFIG_PATH="$(brew --prefix libarchive)/lib/pkgconfig";
       brew install libsndfile jack pulseaudio cppunit
+      brew reinstall --build-from-source libsndfile
 
     build_script: |-
       git submodule init && git submodule update


mauser:
Looks good, certainly quick enough! 

PS: The build will fail since it has not the correct credentials for scp upload, this is disabled for pull request for security reasons..

mauser:
@cme: Could you try [this ](https://sourceforge.net/projects/hydrogen/files/Hydrogen/CI/Hydrogen.dmg/download) binary?

cme:
> @cme: Could you try [this ](https://sourceforge.net/projects/hydrogen/files/Hydrogen/CI/Hydrogen.dmg/download) binary?

Ugh, this seems to suffer the same problem. From the fact that it's *only* libsndfile that seems to suffer from this, I guess that means the issue might actually just be something in the libsndfile build system that's doing something host-OS-specific.

I've tried copying over the Homebrew libsndfile.1.dylib from my Mojave machine to the older machine, and that seems to work. This implies that there's a good chance that changing the `appveyor_build_worker_image` to `macos-mojave` would fix it. Can you try that?

I'll take a look at the libsndfile build system to see if I can find the actual source of the problem.

mauser:
>I've tried copying over the Homebrew libsndfile.1.dylib from my Mojave machine to the older machine, and that seems to work. This implies that there's a good chance that changing the appveyor_build_worker_image to macos-mojave would fix it. Can you try that?

Done, build is rolling: https://ci.appveyor.com/project/mauser/hydrogen/builds/37005297

cme:
> > I've tried copying over the Homebrew libsndfile.1.dylib from my Mojave machine to the older machine, and that seems to work. This implies that there's a good chance that changing the appveyor_build_worker_image to macos-mojave would fix it. Can you try that?
> 
> Done, build is rolling: https://ci.appveyor.com/project/mauser/hydrogen/builds/37005297

...and that seems to fail *somehow* on building by failing to find `libsndfile.h` when building hydrogen. Perhaps the 'reinstall --from-source' somehow failed to complete (though it claims to have been successful). The rebuild shouldn't be needed on Mojave and can be removed.

mauser:
It is even more confusing. The first build i've started (which is linked above), has *not* your commit from this pull request, because i've mistakenly created that branch before doing a `git pull`. It confuses me why cmake does not complain about a missing header (and the sndfile lib is found without problems)

cme:
> It is even more confusing. The first build i've started (which is linked above), has _not_ your commit from this pull request, because i've mistakenly created that branch before doing a `git pull`. It confuses me why cmake does not complain about a missing header (and the sndfile lib is found without problems)

I ran into that (and a similar one) a while ago, fixes on master are: 0e331b328  and 6436d6419. It's surprising that this only seems to be an issue on Mojave!

mauser:
I've cherry-picked those commits, they seem to fix the problem. An updated installer is now available via Sourceforge, i hope that this finally fixes the initial problem :smiley:

cme:
> I've cherry-picked those commits, they seem to fix the problem. An updated installer is now available via [Sourceforge](https://sourceforge.net/projects/hydrogen/files/Hydrogen/CI/Hydrogen.dmg/download), i hope that this finally fixes the initial problem 😃

SUCCESS! Runs just fine on 10.13.

If I have time today, I intend to try and build a 10.12 VM to check if it really does run on 10.12. I already built a 10.10 VM intending to try and make a legacy build there, but Homebrew won't install on 10.10 at the moment.

PR #1060, opened 2021-01-02
fix OSC command signatures

theGreatWhiteShark:
Quite a lot of OSC commands are registered both to work without an argument and with a float. But none of them does actually use the float argument. I removed the second version since it bloats up the list of available commands while providing no additional benefits.

Also, most commands are registered to expect a float argument, rounding it, and providing an integer to internal functions. I changed their signature in expecting an int because it makes the intention of the command more obvious and the code less complex.

@mauser we have the `OscServer::generic_handler` function which listens for a number of URLs, e.g. */Hydrogen/STRIP_VOLUME_ABSOLUTE/(\\d+)*, all carrying the index of a particular strip in the URL and expecting a single float as input argument. Is there a specific reason they were designed this way? How about converting them into individual handler functions requiring two arguments: an integer for the index and a float for the corresponding value.? IMHO this reduce quite a lot of complexity in the `OscServer` class.


mauser:
Hey Phil,
 
the float arguments as well as the behavior of the `OscServer::generic_handler` are workarounds for limitations of the TouchOSC Android/IOS app. When i've started to implement OSC for Hydrogen, i mainly used it to remote control Hydrogen from my phone or tablet during drum rehearsal. At that time, this app has been widely used for this task with other DAWs. 

IIRC, it was not possible to configure the data type of each message with TouchOSC, so i decided to let those variants stay in the code. 

I'm not sure how many people really use this app. I always planned to write an article on our website about this possibility, but never found the time :frowning_face:

theGreatWhiteShark:
I see.

I found it quite strange for the commands handling e.g. the relative volume expecting `20` for a 20% volume instead of 0.2 while explicitly requiring floats.

But maybe this is just me. At least [Renoise](https://tutorials.renoise.com/wiki/Open_Sound_Control#TouchOSC_via_Duplex) and [Ardour](https://manual.ardour.org/using-control-surfaces/controlling-ardour-with-osc/osc-control/) do the same (according to their manual. Never tried it.)

I reverted the changes as well as making the newer commands expecting floats as well.

But double registration of some commands are redundant, right?

mauser:
Hi! 

IIRC, i've used Ardour as a "reference" for implementing those commands. 

Do you mean with "double registration" that we have commands with int as well as float parameters? I remember that i had the problem that TouchOSC used only floats, but other applications used int. So if you registered only float handlers, Hydrogen was not able to work with applications that send integers. But there was no feedback from users on that question, i've just tried out different OSC programs and tried to be compatible..

theGreatWhiteShark:
> Do you mean with "double registration" that we have commands with int as well as float parameters?

Nope. I'm referring to such lines

> m_pServerThread->add_method("/Hydrogen/PLAY", "", PLAY_Handler);
	m_pServerThread->add_method("/Hydrogen/PLAY", "f", PLAY_Handler);

mauser:
Touch OSC also adds an parameter to the OSC message for Toggle buttons. So if you press a "PLAY" button, it will send "/Hydrogen/PLAY 1" (the documentation for the controls can be found [here](https://hexler.net/docs/touchosc-controls-reference)). However, other OSC clients did not append a parameter for toggle buttons.  I do not know if there is a way to do this more elegant in the code. I remember that i looked for a "catchall" method then, but haven't found sth. usable.

theGreatWhiteShark:
Ah. I see. I will add some comments clarifying this and leave the code as is

Issue #665, opened 2018-10-14
Implementation of the panning


theGreatWhiteShark:
I have a quite general question about the implementation of the panning.

In Hydrogen the panning is realized by introducing one volume for the right channel **pan_R** and one for the left **pan_L**, with both of them ranging from 0 to 0.5. If the sound is positioned in the center, both of them are set to 0.5 and if the source moves in either direction, the volume of the opposite channel is reduced.

But in this way the loudness of a sample depends on its location. It is the loudest in the middle and the quietest at either far right or far left. Wouldn't it make more sense to have it at the same level all the time? After all we are adjusting the panning and not the distance. This could be done by keeping the total level of a sample constant for all pannings and expressing it as a combination of the volumes of the two channels. Depending on how they are added, like a 1 = pan_R + pan_L or 1 = sqrt( pan_R^2 + pan_R^2), the scaling between perceived location and panning value can be altered. I'm not a hundred percent sure about which one to use, but from intuition I would guess the second one will lead to a linear scaling of the perceived location since ears and loudspeakers do form a plane. But textbooks will tell us.

quasart:
I think there is no optimal curve for this "pan law". Some DAW just provide a parameter to let user choose : https://www.soundonsound.com/sound-advice/q-what-pan-law-setting-should-use

trebmuh:
_FWIW, I've got no particular point of view on this._

oddtime:
Yes, I like the idea and could work on it.

For what I've read the optimal parameter depends on the room in which you are hearing, that's why some manufacturers use -4.5dB compensation at center and others -3dB.
https://en.wikipedia.org/wiki/Pan_law
We could add a parameter in preferences for the desired pan law

theGreatWhiteShark:
> We could add a parameter in preferences for the desired pan law

Although I like the idea of giving the user control over internal algorithms, adding a parameter for the pan law is probably a little bit over the top. For proper audio production people will most likely use the JACK per track outputs and some dedicated software mixing and mastering tools and ambisonics incorporating the characteristics of the room. (At least I think so #1027).

But nevertheless some attenuation would seem reasonable. @oddtime you can of course implement it yourself if you want to

oddtime:
I am doing personal studies on the matter and I really would like to do some experiments!
So it would not be a problem for me to spend work on it.

Good point #1027

What do you think if we deprecate  pan_R and pan_L in `Note` and `Instrument` classes and introduce a single `pan` member?
[...]

PR #525, opened 2017-08-31
Work In Progress: Probability 2.0

elpescado:
I've been playing with probability attribute for notes and this got me some new ideas how to further improve this feature. I'd like to:

1. Rename "probability" note attribute to "weight" or "importance". The semantics would be same, but it would place less emphasis on randomness.

2. Add two global controls, like "fill" and "randomize" (and invent some more descriptive names). "Fill" would define threshold for note "importance" in order to be played - higher values of "fill" would cause more notes to be played, while lower settings would make patterns more "sparse". "Randomize" would define how that threshold would be randomized. So, instead of note "probability" compared with random value, now note "importance" would be compared with threshold computed from "fill" and "randomize". With "randomize" set to 0, there would be no randomness at all, and every note above threshold will be played. Setting "randomness" to 1 would preserve current behavior - threshold is totally random. Those controls are added to mixer, below humanize controls.

Both "fill" and "randomize" are controllable by MIDI. Users could program a pattern with a few notes with lesser "importance" and while playing song live adjust number of notes ("sparse/fullness") of pattern by twisting a knob on a MIDI controller.

As an extra, this branch allows MIDI control of humanize time, humanize velocity and swing controls, which was not previously possible.

As with previous pull request, this branch should have no impact on legacy behavior - default "importance" value for notes is 1.0 so unless explicitly changed, all notes are played regardless of "fill" and "randomize" values.

Status:
[x] fill and randomize controls
[x] MIDI actions
[x] sampler
[x] MIDI export
[ ] rename parameters
[ ] graphics
[ ] documentation

PR #1061, opened 2021-01-03
Working on Implementation of Panning

oddtime:
See issue #665

I don't want this pull request to be merged now, but just share it.

In Hydrogen the two pan parameters control two separate balances, like a chain, then if they are hard panned to opposite sides, the signal is killed. If there is a reason to mantain this, or somebody likes it, we can add the new feature as option in preferences

**What is changed** now is,  from a user point of view:
**`note` and `instrument` pan interation**: `note` pan moves the pan in a smaller pan range centered at `instrument `pan.
In this pull req. a resultant pan is calculated then the balance law is called. The signal is never killed.
Of course if the `instrument` pan is set to the left, you cannot move the signal to Hard right with the `note` pan (since  `note` pan range is smaller and centered at `instrument` pan); to achieve this, other strategies would be required as discussed in #665 (for example `note` pan bypassing `instrument` pan, this could be another option in preferences if somebody likes it).

In this pull req, the way hydrogen saves pan is unchanged: pan_L and pan_R for both `note` and ` instrument` objects, even if it is now strange and h2 needs to convert them to single parameters and then applies the pan law again.

I used some pointers to member functions, in the future hypothesis of different pan laws implementation. Just to test them, but they should become song members.

// Edit: Added many pan law options, set in mixer: new button on top-right


oddtime:
It is in progress if we decide to add other pan features...
But you can test it and give an opinion on what is already done, described in the first comment

src/core/Sampler/Sampler.cpp
```diff
@@ -233,6 +233,29 @@ void Sampler::noteOff(Note* pNote )
 }
 
 
+/// functions for pan parameters and laws-----------------
+float Sampler::getRatioPan( float fPan_L, float fPan_R ) {
+	// returns the single pan parameter in [-1,1] from the L,R gains
+	// It doesn't return ERROR if (L,R) = (0,0) nor if they are negative!!!!!
```
theGreatWhiteShark, line 239: Regarding the errors in those two functions: If there is no straight forward default value one can return/set, the C++ way of implementing errors would be an exception. But in Hydrogen it's almost never used (and introducing it would make the code most probably harder to maintain).

I would suggest to introduce these limits to `Instrument::set_pan` and `Note::set_pan` instead. If > 0, `0` will be assigned and a WARNINGLOG triggered and the same for 1 from above.

src/core/Sampler/Sampler.cpp
```diff
@@ -257,6 +280,30 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, Song* pSong )
...
 	}
 
+	// new instrument and note pan interaction--------------------------
+	// note pan moves the pan in a smaller pan range centered at instrument pan
+
+	// get the note and instrument pan parameters in [-1,1]
+	// TODO a new song member, set in preferences, must point the desired pan parameter Sampler member function
+	float (*getPan)( float, float );
+	// use "ratio" pan parameter (due to the current pan law)
+    getPan = &this->getRatioPan;
```
theGreatWhiteShark, line 290: Why don't you use `getRatioPan` straight away?

src/core/Sampler/Sampler.h
```diff
@@ -108,7 +108,8 @@ class Sampler : public H2Core::Object
 	 * layer will be loaded with a nullptr instead.
 	 */
 	void reinitializePlaybackTrack();
-
+	static float getRatioPan( float fPan_L, float fPan_R );
```
theGreatWhiteShark, line 111: Could you use Doxygen-style comments to describe the functions? You can e.g. look at the `OscServer` class methods for reference and build the resulting developer documentation using `build.sh d`.

src/core/Sampler/Sampler.cpp
```diff
@@ -257,6 +280,30 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, Song* pSong )
...
+	float fNotePan = getPan( pNote->get_pan_l(), pNote->get_pan_r() );
+	float fInstrPan = getPan( pInstr->get_pan_l(), pInstr->get_pan_r() );
+	
+	// resultant pan: note pan moves the pan in a smaller pan range centered at instrument pan
+	float fPan = fInstrPan + fNotePan * ( 1 - fabs( fInstrPan ) );
```
theGreatWhiteShark, line 295: Why do you use a weight for the note pan to make it more pronounced at the center?

src/core/Sampler/Sampler.cpp
```diff
@@ -257,6 +280,30 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, Song* pSong )
...
+	
+	// Pass fResPan to the Pan Law
+	// use Straight pol pan law.
+	// TODO a new song member, set in preferences, must point the desired pan law Sampler member function		
+	float (*panLaw)( float );
+    panLaw = &this->ratioStraightPolPanLaw;
```
theGreatWhiteShark, line 301: How about making a single `Sampler::panLaw( float, enum)` and an enum class for the different pan laws? Inside this function you could use switch-case statement to distinct between the laws.

src/core/Sampler/Sampler.cpp
```diff
@@ -257,6 +280,30 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, Song* pSong )
...
+    float a = 0.5; // max gain (it has been 0.5 until version 1.0)
```
theGreatWhiteShark, line 302: How about dropping this factor along with [1]

src/core/Sampler/Sampler.cpp
```diff
@@ -619,9 +669,10 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, Song* pSong )
 			cost_L = cost_L * pSong->get_volume();	// song volume
 			cost_L = cost_L * 2; // max pan is 0.5
```
theGreatWhiteShark, line 670: [1] this factor over here? (and the same for the right channel)

src/core/Sampler/Sampler.cpp
oddtime: Not all the pan laws use that definition of the parameter.
In the standard adopted linear panning (a particular case of constant sum pan law i.e. pan_L + pan_R = const) for example the pan parameter has a different dependence on pan_L and pan_R.
So ideally another `getLinearPan( float, float)` function could be pointed

src/core/Sampler/Sampler.h
oddtime: Ok

src/core/Sampler/Sampler.cpp
oddtime: because the resultant note pan range extension is maximum (really from hard left to hard right) when the instrument pan is at center.
For example when instrument pan is hard panned to the right, there is 0 pan extension for the motion due to note_pan
that was the very meaning of "smaller pan range centered at..."
how small depends on the instrument_pan value.

src/core/Sampler/Sampler.cpp
oddtime: I preferred to use pointers to bypass the conditions checks and to go straight to the function

src/core/Sampler/Sampler.cpp
oddtime: Good point! I didn't noticed it

src/core/Sampler/Sampler.cpp
oddtime: Like a "matrioska" pan
image

src/core/Sampler/Sampler.cpp
oddtime: `I would suggest to introduce these limits to Instrument::set_pan and Note::set_pan instead. If > 0, 0 will be assigned and a WARNINGLOG triggered and the same for 1 from above.`

Ok, I keep it for the final version

src/core/Sampler/Sampler.cpp
oddtime: You made me think that the implementation could **always** use the method `ratioStraightPolPanLaw(pan)` to save `pan_L` and `pan_R` and **always** `getRatioPan(pan_L, pan_R)` to reconvert them to the initial single parameter. So all this is just a fixed bijection, that we use only to keep unchanged the way h2 stores the pan parameter information. Even if this will be less intuitive for developers.
Then, when rendering the note, the single pan parameter can be interpreted differently depending on the pan law.
So there is no need of `getLinearPan( float, float)` or `getLinearPan( float, float)`.

src/gui/src/PreferencesDialog_UI.ui
```diff
@@ -564,6 +564,32 @@
...
+         <item>
+          <widget class="QComboBox" name="panLawComboBox">
+           <item>
+            <property name="text">
+             <string>Balance Law (0 compensation)</string>
+            </property>
+           </item>
+           <item>
+            <property name="text">
+             <string>Costant Power</string>
```
trebmuh, line 583: seems to be a typo here (Costant->Co**n**stant)

src/gui/src/PreferencesDialog_UI.ui
```diff
@@ -384,45 +384,45 @@
...
+           <item row="4" column="1">
+            <widget class="QComboBox" name="jackBBTSyncComboBox">
+             <property name="minimumSize">
+              <size>
+               <width>0</width>
+               <height>22</height>
+              </size>
+             </property>
+             <property name="toolTip">
+              <string>Specifies the variable, which has to remain constant in order to guarantee a working synchronization and relocation in the presence of another Jack timebase master.</string>
```
trebmuh, line 412: Since it's an acronym, Jack should be written capitalised: JACK

src/gui/src/PreferencesDialog_UI.ui
```diff
@@ -384,45 +384,45 @@
...
+           <item row="4" column="0">
+            <widget class="QLabel" name="jackBBTSyncLbl">
+             <property name="minimumSize">
+              <size>
+               <width>0</width>
+               <height>22</height>
+              </size>
+             </property>
+             <property name="toolTip">
+              <string>Specifies the variable, which has to remain constant in order to guarantee a working synchronization and relocation in the presence of another Jack timebase master.</string>
```
trebmuh, line 396: Since it's an acronym, Jack should be written capitalised: JACK

trebmuh (changes requested):
Not a programmer, but still helping where I can :wink:

src/gui/src/PreferencesDialog_UI.ui
oddtime: thanks

src/gui/src/PreferencesDialog_UI.ui
oddtime: out of the subject of this pull req, I don't know if I can do it right here

oddtime: is it ok?

oddtime: Alsa too is an acronym... I would prefer do not changing these names in the current pull request

oddtime:
I am attaching a songfile to test the pan laws!
It is the same mono sound going **linearly** from left pan to right pan
[panning.zip](https://github.com/hydrogen-music/hydrogen/files/5793211/panning.zip)
Note: **linearly** refers to the pan parameter, not to the perceived angle, which depends on the pan laws.

src/core/Sampler/Sampler.cpp
oddtime: As the addresses of methods depend on the `Sampler` object, maybe this pointer should be saved in some form in the song (for example written as an `int` or `enum`, which type is new to me now), but should live in `Sample` class.
What do you think?

src/gui/src/PreferencesDialog_UI.ui
trebmuh: I thought it was a new addition since it's a "+" line.
No worries, it can be done later.
_(but still it can be done here while you're at it, there is not potential complication IMHO)_

oddtime:
A summary of what is included until the last commit: different pan laws to set in preferences->audio.
Basically at this point you can hear how different laws sound.
Results sound good in my opinion, there's also a huge variety of options (for the pleasure of research), which is not given in any DAW I know.
With the "linear" pan parameter it seems that you have better control of the stereo position with the pan pot, in my opinion.
Ask if you have any questions. I described something in the comments.

The type of pan law is already saved in the songfile, but I am not convinced about the way it is saved.
I am not totally convinced about the way to store the pan law in the core as well (in which class, using indices, pointers etc).

src/core/Sampler/Sampler.cpp
theGreatWhiteShark: > I preferred to use pointers to bypass the conditions checks and to go straight to the function

I see. But I would like to ask you to use switch-case or if statements nevertheless. It's not because I have a personal reference but the code base of Hydrogen does use the former over function pointers. It's more easy to maintain it if there is a consistent coding style. (Admittedly, the one we deal with is not the best but, well, it works. :wink: )

theGreatWhiteShark:
>  different pan laws to set in preferences->audio.

> The type of pan law is already saved in the songfile, but I am not convinced about the way it is saved.
I am not totally convinced about the way to store the pan law in the core as well (in which class, using indices, pointers etc).

Hydrogen has two main ways for storing information: within each `.h2song` file and a global `hydrogen.conf` file. The latter should contain all the customization the user is likely to use for a lot of sessions and the former those stuff that might change with every song. The panning law feels like a global setting (especially since it can be set via the preference dialog). Therefore, I would move it from the `Song` class into `Preferences`.

Regarding storing: I think the options the user is able to alter should be in the `Preferences` and everything else can be a member of `Sampler`. Or do you think other parts of the program expect the `Sampler` will use these functions too?

theGreatWhiteShark:
>  different pan laws to set in preferences->audio.

> The type of pan law is already saved in the songfile, but I am not convinced about the way it is saved.
I am not totally convinced about the way to store the pan law in the core as well (in which class, using indices, pointers etc).

Hydrogen has two main ways for storing information: within each `.h2song` file and a global `hydrogen.conf` file. The latter should contain all the customization the user is likely to use for a lot of sessions and the former those stuff that might change with every song. The panning law feels like a global setting (especially since it can be set via the preference dialog). Therefore, I would move it from the `Song` class into `Preferences`.

Regarding storing: I think the options the user is able to alter should be in the `Preferences` and everything else can be a member of `Sampler`. Or do you think other parts of the program expect the `Sampler` will use these functions too?

oddtime:
> Hydrogen has two main ways for storing information: within each .h2song file and a global hydrogen.conf file. The latter should contain all the customization the user is likely to use for a lot of sessions and the former those stuff that might change with every song. The panning law feels like a global setting (especially since it can be set via the preference dialog). Therefore, I would move it from the Song class into Preferences.

Since the pan law affects how the project sounds and how it will be mixed, it should be stored in the .h2song in my opinion.
We can save it in `hydrogen.conf ` just to keep it saved for the next songs

> Or do you think other parts of the program expect the Sampler will use these functions too?

I don't think so

oddtime:
And what about the way of storing in the file? with an integer number? Maybe I would go with a meaningful string

src/core/Sampler/Sampler.cpp
oddtime: I understand the reason.
I would just ask you if you would consider consistent this alternative:
`float (*m_pPanLaw)( float )` is a member in `Sampler`
While import the song, or setting the preferences, the corresponding member function is assigned to `m_pPanLaw` with `if `statements instead of that weird array of addresses

theGreatWhiteShark:
> We can save it in hydrogen.conf just to keep it saved for the next songs

I would store it in just one file. Else it will only cause confusion.

> Since the pan law affects how the project sounds and how it will be mixed, it should be stored in the .h2song in my opinion.

I do see your point. But if we store it in the Song file the option must not be set via the `PreferenceDialog`. Do you think user would set different values for each song or just play with it and decide for one that fits them best?

theGreatWhiteShark:
> And what about the way of storing in the file? with an integer number? Maybe I would go with a meaningful string

An `enum class`. This way the variable name is a meaningful name but it's value is an integer which is way more convenient to handle. See https://github.com/hydrogen-music/hydrogen/blob/master/src/core/Preferences.h#L375 for an example.

oddtime:
> I do see your point. But if we store it in the Song file the option must not be set via the PreferenceDialog. Do you think user would set different values for each song or just play with it and decide for one that fits them best?

I will try to explain why I would like to save it in the song
Actually my concern is: once you change the type of pan law the song sounds in a different way.
The main reason is not only the change of constraint (constant sum, constant power...) indeed, but the way the pan parameter maps the "virtual stereo positions". I mean that the same pan parameter value may "put" the signal in any point of the half stereo panorama depending on the pan law, except for the "central" and "hard-side" positions. We may ask why do not adopt just one bijection between pan and the "virtual stereo placement"? The reason is, in my thought, that there is no a solution to this problem, because the perceived angle could depend on difficult quantities like frequency (also for psycho-acoustics and non-linearity). Also, even if the perceived angle would depend on gain_L and gain_R only, it is not clear what would be the "linear" dependency between gain_L, gain_R and the perceived angle (I said "linear" but of course we don't know any relation). That is to say, the pan laws that I found on manuals, which are all different and inconsistent in my opinion (i.e. switching between them the angle does change, as you can hear experimentally), are all valid theorically. So I included all that pan laws but introduced the "ratio","linear", "quadratic", "polar" distinction to create classes of pan law that map the sounds always in the same points (at least it seems so by ear, and I have another reasonable argument for that...), with compensated gain depending on which pan law. Plus I added a custom pan law which generalises the constraint L^k+R^k=1, making consistent compromises.
All this is to say that if you share a song or for some reason you have changed the pan law setting preference, when you re-open the song you may listen to a very different mix. Tests by ear confirm that (try to switch between "linear" and "quadratic")
On daws also the pan laws are stored in the project (I searched briefly on the web).

> Do you think user would set different values for each song?

If you use pan note property for a surrealistic effect of instruments moving in the space, this could happen. Otherwise I think that every pan law is equivalent, as you will compensate the gain by ear.

theGreatWhiteShark:
Yes, you are right. Also when sharing a song the panning law shouldn't be altered by the preferences of the other user.

This still leaves the question of where to put the control/widget to alter this setting. Since it just concerns the internal sound generation it might go to the `Mixer`. But there isn't a place for it yet.

But wait. Does it only concerns the internal sound generation? Doesn't it affect the MIDI export of the notes too? (disclaimer: I have almost no idea about MIDI)

oddtime:
> This still leaves the question of where to put the control/widget to alter this setting.

Ok. I approve putting a new widget in the mixer. I think a button on the right, accordingly with the pull req #525 
Another place could be in menu->project

>  Doesn't it affect the MIDI export of the notes too?

In MIDI, pan has its dedicated message from 0-127 so it seems that midi does not transmit the information of separate gain_L and gain_R. I don't know how the pan value is output in midi currently and I should update the code to export the resultant pan coming from instrument and note pan.

oddtime:
Here's a space in the mixer:
Untitled
(maybe too small and bad font)
the button opens a dialog window:
screenshot-8691da12

theGreatWhiteShark:
> Ok. I approve putting a new widget in the mixer. I think a button on the right, accordingly with the pull req #525
Another place could be in menu->project

How about we use a general settings/cog icon in the Mixer instead? This will open a new window from which the pan law dialog can be accessed. This will scale much better with respect to further features/buttons etc. in the mixer.

> is there a strong reason to put it in the mixer?

Since it is only used in the sampler, it only takes effect when using the post fader audio output of Hydrogen. If you use it solely as sequencer to power another drum synth - like drumgizmo - or use the per-track pre fader audio outputs via JACK, the pan law will not alter the result. By placing it in the menu it kinda suggests to affect the panning done in the instrument rack and note property ruler too.

oddtime:
Right. I have implemented the cog icon button in the mixer and removed pan law settings from menu.
With that the pan work is almost finished, I just need to review and clean the code.

src/core/Sampler/Sampler.cpp
oddtime: > I would suggest to introduce these limits to Instrument::set_pan and Note::set_pan instead.

There is already a check for negative pan:
```
void Note::set_pan_l( float pan )
{
	__pan_l = check_boundary( pan, PAN_MIN, PAN_MAX );
}

```

In this way H2 one can set both pan_L = pan_R = 0. In this case there is a problem to reconvert back to the pan parameter. It could happen importing a bad song.

src/gui/src/PreferencesDialog_UI.ui
oddtime: It appeared as if I had added that lines but it was rewritten by the graphical software. However that file is untouched in the last commits.

theGreatWhiteShark:
How about introducing 4 headings/categories in the "Select Pan Law" options: linear, polar, ratio, quadratic. Right now it's a little bit hard to get an overview with so much different options densely packed.

oddtime:
Well I had the same idea but was lazy to set it

oddtime:
Unfortunately I am not an expert of Qt and this seems complex.
Is there a way of putting category headings in a QComboBox, simpler than the following?
https://stackoverflow.com/questions/33012292/grouped-qcombobox

theGreatWhiteShark:
> Unfortunately I am not an expert of Qt and this seems complex.
Is there a way of putting category headings in a QComboBox, simpler than the following?
https://stackoverflow.com/questions/33012292/grouped-qcombobox

Well, unfortunately neither am I.

First of all I would recommend using the `QtCreator` (I was manually editing the UI files before realizing this was not the intended way). Doubling clicking the combobox gives some control over its content. Maybe some lines can be disabled and be display as bold or italic. Also, the resolution combobox in the pattern editor features separators. They might help too

oddtime:
The current pan law combobox already uses separators but there is not so much contrast to see them clearly.

I used Qt creator to create the dialog, but added the combobox items in `MixerSettingsDialog.cpp` because I can associate the `enum class` to each (which is in `Sampler class`). This seemed good to avoid referring to indices.

oddtime:
I found a way

oddtime:
Are you ok with the current namings?
Basically _linear_ and _quadratic_ are abbreviations of _"average, linear weights"_ and _"average, quadratic weights"_, _polar_ means _"polar coordinates"_, _ratio_ well means the ratio :)
I hope this makes sense if you have read the comments in the code, otherwise I can explain it better.

theGreatWhiteShark:
> I found a way

Nice!

I'll do my best to review the changes during the next days.

> Are you ok with the current namings?
Basically linear and quadratic are abbreviations of "average, linear weights" and "average, quadratic weights", polar means "polar coordinates", ratio well means the ratio :)

Sounds good.

src/core/Basics/Song.cpp
```diff
@@ -725,6 +726,53 @@ Song* SongReader::readSong( const QString& sFileName )
...
+	} else if ( sPanLawType == "QUADRATIC_CONST_K_NORM" ) {
+		pSampler->setPanLawType( pSampler->QUADRATIC_CONST_K_NORM );
+	} else {
+		pSampler->setPanLawType( pSampler->RATIO_STRAIGHT_POLYGONAL );
+		WARNINGLOG( "Unknown pan law type in import song. Set default." );
+	}
+
+	float fPanLawKNorm = LocalFileMng::readXmlFloat( songNode, "pan_law_k_norm", K_NORM_DEFAULT );
+	if ( fPanLawKNorm <= 0.0 ) {
+		fPanLawKNorm = K_NORM_DEFAULT;
```
theGreatWhiteShark, line 772: Could you make `K_NORM_DEFAULT` a static member variable of `Sampler`? This way it will be perfectly clear where to find its definition and documentation.

src/core/LocalFileMgr.cpp
```diff
@@ -386,6 +386,49 @@ int SongWriter::writeSong( Song * pSong, const QString& filename )
...
+	} else if ( nPanLawType == pSampler->RATIO_CONST_K_NORM ) {
+		sPanLawType = "RATIO_CONST_K_NORM";
+	} else if ( nPanLawType == pSampler->QUADRATIC_CONST_K_NORM ) {
+		sPanLawType = "QUADRATIC_CONST_K_NORM";
+	} else {
+		WARNINGLOG( "Unknown pan law in saving song. Saved default type." );
+		sPanLawType = "RATIO_STRAIGHT_POLYGONAL";
+	}
+	LocalFileMng::writeXmlString( songNode, "pan_law", sPanLawType );
+	LocalFileMng::writeXmlString( songNode, "pan_law_k_norm", QString("%1").arg( pSampler->getPanLawKNorm() ) );
```
theGreatWhiteShark, line 430: Could you move this one to `Song` too?

src/core/Sampler/Sampler.cpp
```diff
@@ -233,6 +236,229 @@ void Sampler::noteOff(Note* pNote )
 }
 
 
+// functions for pan parameters and laws-----------------
+
+float Sampler::getRatioPan( float fPan_L, float fPan_R ) {
+   /** returns the single pan parameter in [-1,1] from the L,R gains
```
theGreatWhiteShark, line 242: Could you move the doc string above the definition in order for Doxygen to find it? You can check using `./build.sh d`. And also in header file to not bloat the source code?

src/core/Sampler/Sampler.cpp
```diff
@@ -233,6 +236,229 @@ void Sampler::noteOff(Note* pNote )
...
+	} else {
+		if ( fPan_L >= fPan_R ) {
+			return fPan_R / fPan_L - 1.;
+		} else {
+			return 1. - fPan_L / fPan_R;
+		}
+	}
+}
+
+   /** PAN LAWS
```
theGreatWhiteShark, line 257: Nice. Thanks for the thorough documentation! Could you move it above the `enum` definition in the header? Right now, it will be shown as the description of the ` ratioStraightPolygonalPanLaw()` function.

src/gui/src/Mixer/MixerSettingsDialog_UI.ui
```diff
@@ -0,0 +1,200 @@
...
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>490</width>
+    <height>157</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form1</string>
```
theGreatWhiteShark, line 14: Could you remove this property? It will be overwritten in the .cpp file but get's listed for the translators who will have no idea what to do with it

theGreatWhiteShark (changes requested):
Nice. I like it!

Some minor things but almost done.

src/core/Basics/Song.cpp
```diff
@@ -725,6 +726,53 @@ Song* SongReader::readSong( const QString& sFileName )
 	pSong->setPlaybackTrackEnabled( bPlaybackTrackEnabled );
 	pSong->setPlaybackTrackVolume( fPlaybackTrackVolume );
 	pSong->setActionMode( actionMode );
+	
+	// pan law
+	Sampler* pSampler = AudioEngine::get_instance()->get_sampler();
+	QString sPanLawType( LocalFileMng::readXmlString( songNode, "pan_law", "RATIO_STRAIGHT_POLYGONAL" ) );
```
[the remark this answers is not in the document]
oddtime: Ok, so how to match?
You mean defining something similar to an array of strings?
`panlawstrings[]`
and referring as `panlawstrings[ RATIO_STRAIGHT_POLYGONAL ]` for example?

src/core/Basics/Song.cpp
```diff
@@ -725,6 +726,53 @@ Song* SongReader::readSong( const QString& sFileName )
...
+	if ( sPanLawType == "RATIO_STRAIGHT_POLYGONAL" ) {
+		pSampler->setPanLawType( pSampler->RATIO_STRAIGHT_POLYGONAL );
```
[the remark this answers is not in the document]
oddtime: Ok

src/core/Basics/Song.cpp
oddtime: ok

src/core/Basics/Song.cpp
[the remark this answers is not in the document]
theGreatWhiteShark: Since the `enum` can be used as an `int` you can do something like

```c++
int nPanLawType = LocalFileMng::readXmlInt( songNode, "pan_law_type", 0 );
pSong->setPanLawType( nPanLawType );
```

If you use an `enum class` this, however, is not possible and you have to map it's members to `int` or `QString`.

Anyway. It's not that important how to store the variable (as a string or a int) since a) the amount of memory it requires on the system of the user is negligible and b) it's just an internal serialization and not intended to be edited directly. You can also leave it as it is. I don't mind.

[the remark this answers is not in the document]
oddtime: I see. The reason why I would not save it as an int is because one could add another pan law type in the future (who knows? or for personal pan law tests) and the string keeps the file interpretation save.
I had in mind that writing/reading the string from/to the xml file could have be done in a safer way (I just have to go deeper in the `enum class` )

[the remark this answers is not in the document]
theGreatWhiteShark: Okay. Then just leave it as is. You don't have to rewrite it as `enum class`. (Why I brought that up the begin with: a number of `enum`s in the code where/are handled like `if ( m_nPanLawType == 5 )` instead of `if ( m_nPanLawType == LINEAR_CONST_SUM )` which makes the code harder to grasp in a totally unnecessary way. But your code hygiene is better, so don't mind :) )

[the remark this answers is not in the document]
oddtime: Ok!

src/core/Basics/Song.cpp
```diff
@@ -725,6 +726,53 @@ Song* SongReader::readSong( const QString& sFileName )
 	pSong->setPlaybackTrackEnabled( bPlaybackTrackEnabled );
 	pSong->setPlaybackTrackVolume( fPlaybackTrackVolume );
 	pSong->setActionMode( actionMode );
+	
+	// pan law
+	Sampler* pSampler = AudioEngine::get_instance()->get_sampler();
```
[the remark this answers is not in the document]
oddtime: Ok

oddtime:
Thanks for the check! I did the changes. If you want merge I would add the feature in Changelog

theGreatWhiteShark (approved):

theGreatWhiteShark:
> If you want merge I would add the feature in Changelog

Yes, please.

@oddtime one last thing. One of the unit tests does fail. Could you look into it (using `./build.sh t`)?

@trebmuh are all of your requested changes covered? I'm fine with everything as soon as the unit test is resolved.

oddtime:
```
1) test: FunctionalTest::testExportAudio (F) line: 130 /home/scarbo/programming/H2Repo/hydrogen/src/tests/FunctionalTests.cpp
Files differ at sample 89898
- Expected: /home/scarbo/programming/H2Repo/hydrogen/src/tests/data//functional/test.ref.flac
- Actual  : /tmp/hydrogen/test-wiTawq.wav
```

Any ideas? Files are 2 s long at 44100Hz...

trebmuh:
> @trebmuh are all of your requested changes covered?

They were only typo/cosmetic requests. It looks all good from my PoV.

oddtime:
the first tracks are the exported and reference files: test-wiTawq.wav test.ref.flac. The third trck is the difference
screenshot-6503ea24

Normalization shows that the difference is not 0 at this sample.
Untitled
but it something like -50dB...
There is another sample like that, may it be caused by a small CPU approximation error (getting back the pan from panL and panR and recalculating the pan law)?

oddtime:
I modified the c++ file doing the test to print the different samples in the shell:

`.........................sample1 = 7484, sample2 = 7485`

theGreatWhiteShark:
Yeah, two little differences. One in the middle of the song and one at 3/4. I noticed that when diffing the hexdump of these files too.

I did place some prints to `std::cout` in the `DiskWriterDriver` and the `Sampler` to see what's going on. There were quite a number of (very tiny) differences in the float values written in the former. But in `Sampler::renderNote()` the `cost_*` variables have not changed at all. And, AFAIR, this is the very place you did some changes in.

I will keep digging until I find something

oddtime:
Thanks @theGreatWhiteShark !

theGreatWhiteShark:
Okay. Everything's fine.

There is some truncation happening causing very small differences in the resulting output. In principle we could make the pan calculate use `double` instead of `float` but IMHO this would be over the top.

Could you take one of these temporary results and replace the `test.ref.flac` file?

oddtime:
Ok, done.
(but there's still that `test.wav` which seems the old reference)

PR #1064, opened 2021-01-03
Improve handling of pattern naming/renaming

Ratfink:
I noticed a few annoying quirks with how Hydrogen avoided duplicate pattern names, so I thought I'd try my hand at fixing them.  When editing a pattern name inline in the song editor, if the user pressed Enter without changing the pattern name, it would get a `" #1"` suffix added to the end.  Repeating this would result in a second `" #1"` suffix being appended  Also, the pattern properties dialog used a different procedure for avoiding name collisions, appending `"#2"` until the name was unique.

To address these issues, this merge request:

* Enhances `PatternList::find_unused_pattern_name` to detect if a `" #[number]"` suffix has already been added to a pattern's name, and if so, increment the number instead of adding a new suffix.
* Makes the pattern properties dialog use functions from the core for detecting and avoiding pattern name collisions, instead of reimplementing that functionality in a different way in the GUI code.
* Adds an optional pattern to ignore in `PatternList::find_unused_pattern_name` and `PatternList::check_name` so that when renaming a pattern, its own previous name can be ignored.

### DIFF
--- a/src/core/Basics/PatternList.cpp
+++ b/src/core/Basics/PatternList.cpp
@@ -195,21 +195,21 @@ void PatternList::virtual_pattern_del( Pattern* pattern )
 	for( int i=0; i<__patterns.size(); i++ ) __patterns[i]->virtual_patterns_del( pattern );
 }
 
-bool PatternList::check_name( QString patternName )
+bool PatternList::check_name( QString patternName, Pattern* ignore )
 {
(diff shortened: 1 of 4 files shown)


theGreatWhiteShark (approved):
Thanks a lot!

The renaming was indeed quite annoying.

PR #1079, opened 2021-01-07
Refactor of song editor to use common Selection manager.

cme:
Also adds cut and paste, page up/down keys, scrolling during drag,
etc.

Todo list:
  - [x] Context menu 
  - [x] Doc comments
  - [x] Change default mode to 'select' rather than 'draw'
  ~~- [ ] Maintain GridCell map on edits, rather than rebuilding from scratch on draw~~ no need, rebuild on draw is efficient enough
  - [x] Lazy selection updates: only update selection when mouse crosses boundaries between cells
  - [x] Change to 'select' mode if shift+arrows used in 'draw' mode
  ~~- [ ] Shortcut for pattern select~~ TBD


theGreatWhiteShark:
I just realize the middle-click of a mouse does select an instrument and move the cursor in the `PatternEditor` without introducing a new note or opening the context menu. In the `SongEditor`, however, the middle-click does toggle a pattern and move the cursor.

It would be nice that have a consistent behaviour and I would vote for the first one - selecting a row without introducing something new.

theGreatWhiteShark:
In the `PatternEditor` the instrument is the currently selected one which is associated with the keyboard cursor. I would suggest to have the same for the `SongEditor`.

I do like this behaviour a lot! This is much more straight forward than a shortcut for cycling through the patters/instrument.

theGreatWhiteShark:
I also have another suggestion for the `NotePropertyRuler`.

When altering the value of a property via keyboard the corresponding value does not show up in the main LCD display (the one below the "Mixer" and "Instrument rack" buttons). I kinda miss this because when fine tuning the instruments of a many copies of a single pattern I would like to align the velocities of the instruments.

Also, the [up] and [down] buttons do nicely in/decrement the value by 0.1. But it would be convenient to have e.g. [Shift+up|down] to make a further fine tuning by 0.01.

cme:
> When altering the value of a property via keyboard the corresponding value does not show up in the main LCD display (the one below the "Mixer" and "Instrument rack" buttons). I kinda miss this because when fine tuning the instruments of a many copies of a single pattern I would like to align the velocities of the instruments.

Excellent idea! Will add.
 
> Also, the [up] and [down] buttons do nicely in/decrement the value by 0.1. But it would be convenient to have e.g. [Shift+up|down] to make a further fine tuning by 0.01.

Excellent. Although 'shift+arrows' would naturally be a selection operation though.

The "Alt" key in the pattern editor enables fine-grained movement of selected notes. Maybe that makes 'alt' appropriate to use for this too.

Also, "alt" should probably do the same thing when clicking in the pattern editor, and scroll wheel events in the properties editor should accept "alt" as meaning fine-grained motion too?

cme:
> In the `PatternEditor` the instrument is the currently selected one which is associated with the keyboard cursor. I would suggest to have the same for the `SongEditor`.
> 
> I do like this behaviour a lot! This is much more straight forward than a shortcut for cycling through the patters/instrument.

I like this idea, but am a bit wary as it feels like a larger change. Maybe enabled as an option rather than always-on behaviour.

The problem mostly is that it implies a change to what happens when you click a cell to enable/disable it. Since the keyboard cursor follows the mouse click, if the selected pattern followed the cursor, then activating / deactivating a cell would potentially change selected pattern too. Not that this is inherently a problem, just that it's a change.

Does that sound like desirable or undesirable behaviour?

Perhaps some middle ground of changing patterns on keyboard move, but not when positioning the cursor with the mouse?

theGreatWhiteShark:
> The "Alt" key in the pattern editor enables fine-grained movement of selected notes. Maybe that makes 'alt' appropriate to use for this too.

It does? On my laptop "Alt" + arrow keys scrolls.

But in general it's a very nice idea to have a separate key for fine tuning. In case we add keyboard support for all the knobs and faders in the mixer and instrument rack the arrow & "Alt" + arrow duo will become handy too.

theGreatWhiteShark:
> The problem mostly is that it implies a change to what happens when you click a cell to enable/disable it. Since the keyboard cursor follows the mouse click, if the selected pattern followed the cursor, then activating / deactivating a cell would potentially change selected pattern too. Not that this is inherently a problem, just that it's a change.
>
> Does that sound like desirable or undesirable behaviour?

TBH, I didn't thought about the mouse click but I like it! At least for the "select mode" it gives an optical feedback by showing the added/deleted pattern and for the drag selection it is even better. I instantly thought of a use case like "I want to select all patterns up to a certain one". Using this behavior one does not have to keep an eye on the row anymore but could also drag until the desired pattern shows up in the pattern editor.

But I agree, this will be a larger change and some people might not like it.

 > Maybe enabled as an option rather than always-on behaviour.

Hmm. I'm starting believe there will be a couple of other options popping up in the future as we change the UX. How about condensing them - including the "hide cursor" option - in an "legacy" option to increase maintainability?

PR #1084, opened 2021-01-09
refactor Song class

theGreatWhiteShark:
- adopt naming convention
- make member variables private and use setter and getters
- review constness

At first I just wanted to introduce a setter for the bpm member of the `Song`. But things went out of hand.

### DIFF
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -334,8 +334,8 @@ int main(int argc, char *argv[])
 			/* Still not loaded */
 			if (! pSong) {
 				___INFOLOG("Starting with empty song");
-				pSong = Song::get_empty_song();
-				pSong->set_filename( "" );
+				pSong = Song::getEmptySong();
+				pSong->setFilename( "" );
 			}
 
 			pHydrogen->setSong( pSong );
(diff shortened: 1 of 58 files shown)


cme:
Ace. How much of this was automated / repeatable? Because I just realised anything I am likely to commit soon will cause conflicts with this, so merging this sooner would be better and reduce conflicts.

(I had a quick look through, looks good!)

> At first I just wanted to introduce a setter for the bpm member of the `Song`. But things went out of hand.

I know that feeling... :D

jeremyz:
All of this because I can't stand neither camelCase nor member variable prefixes, sorry ;)

theGreatWhiteShark:
> How much of this was automated / repeatable?

Let's say there was enough automation for the whole thing to only take a whole evening. What mauser is doing in #1008 is on another scale. (Admittedly, I started that one once too. But gave halfway through :smile: )

theGreatWhiteShark:
> All of this because I can't stand neither camelCase nor member variable prefixes, sorry ;)

Well, it's not my favorite either. But the inconsistency gives an almost physical pain if looking at it for too long. :wink:

PR #1086, opened 2021-01-09
Prevent duplicate names in the pattern properties dialog

### LINKED ISSUE #1083: Multiple patterns can have the same name
Hydrogen version: 1.0.1 or 7b2371758d9f7534194e7379701cd8c59b91581e

After making a couple changes to pattern renaming in #1064, I noticed that it's currently possible to get two patterns with the same name using the pattern properties dialog.  Steps to reproduce:

1. Create a new project (Project > New).
2. Right-click on Pattern 1 in the pattern list, choose Properties.
3. Change "New Pattern Name" to "Pattern 2".  OK button will become inactive.
4. Change "Pattern Category" to anything other than the default of "not_categorized".  OK button will become active.
5. Press the OK button.  There will now be two patterns called "Pattern 2".

This appears to be due to a combination of not checking for duplicate names when OK is pressed, combined with flawed logic that makes the OK button active whenever either the pattern name or category was just changed.

My proposed solution would be to check for and fix duplicate names using the same logic as inline pattern renaming.  Then we also wouldn't need to disable the OK button (though it would probably be worth it to check if all the fields are unedited, and not push anything onto the undo stack if so).  I'll plan to submit a PR to do this, probably sometime this weekend.

Ratfink:
This PR fixes #1083 by letting the pattern properties dialog change the entered name just like the inline pattern name editor.  With this change in place, the OK button is also kept enabled at all times.  If it is pressed without any field having changed, nothing is pushed onto the undo stack, to avoid confusion that may result from a no-op "undo" action.

---

The changes to the OK button's behavior made the `PatternPropertiesDialog::on_categoryComboBox_editTextChanged` and `PatternPropertiesDialog::on_patternNameTxt_textChanged` methods unnecessary, so I removed them both.  I also removed `PatternPropertiesDialog::nameCheck` because it was very short and only called in one place, so simply moving it to the caller made more sense to me.  The same point could be made about `PatternPropertiesDialog::defaultNameCheck`, which could be moved to the dialog's constructor, but I've left it in place for now.


theGreatWhiteShark:
Hey @Ratfink,

could you rebase your PR on current master? Sorry for that.

Ratfink:
Hmm, looks like I *really* messed that history up.  Probably best to close this one and make another, right?

src/gui/src/PatternPropertiesDialog.cpp
```diff
@@ -125,45 +128,11 @@ void PatternPropertiesDialog::on_okBtn_clicked()
...
 
-void PatternPropertiesDialog::on_patternNameTxt_textChanged()
-{
-	if ( nameCheck( patternNameTxt->text() ) ) {
-		okBtn->setEnabled( true );
-	}
-	else {
-		okBtn->setEnabled( false );
-	}
+	patternNameTxt->setText( tr("%1").arg(pattName) );
```
theGreatWhiteShark, line 137: Could you remove the `tr()` around "%1"? This marks the string "%1" translatable and it will show up in `.ts` file intended for the translators. No idea why this had been there in the first place

theGreatWhiteShark (commented):
Don't worry about the commits showing up here. Their hash is unchanged and merging the branch locally does not cause any problems. It's probably just some rendering oddities of Github.

src/gui/src/PatternPropertiesDialog.cpp
Ratfink: Actually, is the call to `arg()` needed here at all, or could the whole line just be `patternNameTxt->setText(pattName)`?

theGreatWhiteShark: Yes, your right.

theGreatWhiteShark:
Thanks a lot!

Issue #1081, opened 2021-01-08
Grid width of PatternEditor and SongEditor arn't stored


theGreatWhiteShark:
Hydrogen version *          : 13fbc2b40b54dd7a8a9213889f85c5e99e34ec46
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

AFAICS neither `PatternEditorPanel::zoomInBtnClicked()` nor `PatternEditor::zoomIn()` to the other zoomIn functions call `Preferences::setPatternEditorGridWidth` to store the new value of the grid width (although it is read and used for initialization).

For the `SongEditor` there seem to even not be a corresponding XML node to store it. I would vote for introducing it to make storing of the properties of those two widgets consistent.

PR #1090, opened 2021-01-11
Fix state storage

### LINKED ISSUE #1082: Visibility of instrument rack is not stored
Hydrogen version *          : 13fbc2b

Hydrogen does to not remember the visibility of the instrument rack. There is a `Preferences::drumkitManagerProperties` in place but it seems to just be dead code.

theGreatWhiteShark:
IMHO it's debatable whether to store the state of a particular button/checkbox in the `*.h2song` or in the `hydrogen.conf` file. But the behavior of storing some states but not others does not seem right. (Although e.g. comment in the code of the instrument rack widget suggest it's intended for it to be shown on default. Well, but turning it of every single time (to have a larger PatternEditor) is quite frustrating.

I introduced
- Storing the grid width/height for the `PatternEditor` and `SongEditor` (@cme the height I just introduced for the sake of completeness. AFAICS we do not support alternating this value via the GUI. I think this is legit and we can keep it this way. What do you think?)
- Storing the visibility of the instrument rack
- Storing whether the select or draw mode was selected for the `SongEditor` (@cme I made the select mode the new default as you mentioned you were gonna do this anyway. But since the selection will be remembered, the default value might not be that important anyway.)

Fixes #1082 and #1081


cme:
Nice.

I'm not sure about moving the draw/select mode itself into Preferences though; using it directly from the Preferences object feels uncomfortable because it's part of the song editor state and "belongs" there. Conceptually at least, if we had the capability of having multiple song editors, the state should belong to each one individually, but initialised from Preferences and written back at some points (perhaps every change, or just destroying the SongEditorPanel object).

What's the criteria for allowing direct access to a Preferences field vs. using accessor functions? Previously the actionmode was at least read by an accessor from its primary location.

theGreatWhiteShark:
> I'm not sure about moving the draw/select mode itself into Preferences though; using it directly from the Preferences object feels uncomfortable because it's part of the song editor state and "belongs" there. Conceptually at least, if we had the capability of having multiple song editors, the state should belong to each one individually, but initialised from Preferences and written back at some points (perhaps every change, or just destroying the SongEditorPanel object).

Yes. Your right. I would move it into the `Song` class instead and store it in the `.h2song`. It's right next to e.g. the Timeline usage or the Playback track which all are song-dependent too.

> What's the criteria for allowing direct access to a Preferences field vs. using accessor functions? Previously the actionmode was at least read by an accessor from its primary location.

Hmmm. It's 60% rule 41 of the C++ coding standard by Sutter & Alexandrescu, 2005 "Make data members private, except in behaviorless aggregates" (admittedly, the `Preferences` class is not a struct but pretty much just a container for variables + loading and saving) and 40% laziness. But I can introduce a getter/setter (which I would do when moving it into the `Song` anyway).

cme (approved):
Looks good!

I'm still not sure what the rationale for moving the actionmode from the SongPanelEditor to the Song is though? It feels more like editor state than song state to me, but I only have a feeling to go on here, nothing concrete! :)

theGreatWhiteShark:
> I'm still not sure what the rationale for moving the actionmode from the SongPanelEditor to the Song is though? It feels more like editor state than song state to me, but I only have a feeling to go on here, nothing concrete! :)

Yes. But currently we only have two places to store things: in the `Preferences` for all things that affect all sessions and `Song` for all things corresponding to an individual session. Since both classes are in the core part of the code - which can only interact with the GUI via the `EventQueue` - the variable has to be defined there too in order to save it to disk. But I agree, it's a little bit strange to have it there. The same is true for the song mode and the filename of the playback track.

PR #1094, opened 2021-01-13
GMKit -> GMRockKit (MainForm.cpp)

trebmuh:
There is no "GMKit" anymore. This looks to be an leftover from the time we switched from the GMKit to the GMRockKit. This could confusing or users.

I don't have any knowledge about using this lilypond converter so I'm not sure if the exports are still corrects. Hence I'm not merging this PR myself.


theGreatWhiteShark:
> I don't have any knowledge about using this lilypond converter so I'm not sure if the exports are still corrects. Hence I'm not merging this PR myself.

Well, since your changes do only concern the text in the popup window, the code won't be affected.

Issue #1088, opened 2021-01-10
if() statement before calling random generators for note probability and random pitch


oddtime:
in `src/core/Hydrogen.cpp` line 980 circa

```
float rnd = (float)rand()/(float)RAND_MAX;
if (pNote->get_probability() < rnd) {
m_songNoteQueue.pop();
pNote->get_instrument()->dequeue();
continue;
}
```

```
// Offset + Random Pitch ;)
pNote->set_pitch( pNote->get_pitch()
			+ pNote->get_instrument()->get_pitch_offset()
			+ getGaussian( 0.4 ) * pNote->get_instrument()->get_random_pitch_factor() );
```

the random generators could be called after a check if noteprobability and randomPitch are different from 1 and 0.

oddtime:
see pull req #1104

theGreatWhiteShark:
Closed with #1104

PR #1104, opened 2021-01-17
if() statement before calling random functions for note probability and random pitch

oddtime:
refers to issue #1088

### DIFF
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -979,12 +979,17 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 		if ( isNoteStart || isOldNote ) {
 			// Humanize - Velocity parameter
 			pNote->set_velocity( pNote->get_velocity() * velocity_adjustment );
-
-			float rnd = (float)rand()/(float)RAND_MAX;
-			if (pNote->get_probability() < rnd) {
-				m_songNoteQueue.pop();
-				pNote->get_instrument()->dequeue();
-				continue;
+			
+			/* Check if the current note has probability != 1
+			 * If yes remove call random function to dequeue or not the note
+			 */
+			float fNoteProbability = pNote->get_probability();
+			if ( fNoteProbability != 1. ) {
+				if ( fNoteProbability < (float) rand() / (float) RAND_MAX ) {
+					m_songNoteQueue.pop();
+					pNote->get_instrument()->dequeue();
+					continue;
+				}
 			}
 
 			if ( pSong->getHumanizeVelocityValue() != 0 ) {
@@ -1002,9 +1007,15 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 			}
 
 			// Offset + Random Pitch ;)
-			pNote->set_pitch( pNote->get_pitch()
-						 + pNote->get_instrument()->get_pitch_offset()
-						 + getGaussian( 0.4 ) * pNote->get_instrument()->get_random_pitch_factor() );
+			float fPitch = pNote->get_pitch() + pNote->get_instrument()->get_pitch_offset();
+			/* Check if the current instrument has random picth factor != 0.
+			 * If yes add a gaussian perturbation to the pitch
+			 */
+			float fRandomPitchFactor = pNote->get_instrument()->get_random_pitch_factor();
+			if ( fRandomPitchFactor != 0. ) {
+				fPitch += getGaussian( 0.4 ) * fRandomPitchFactor;
+			}
+			pNote->set_pitch( fPitch );
 
 
 			/*


theGreatWhiteShark (approved):
I'm not so much into the C++ compiler. But it features some smart optimization stuff causing the body of `if` clauses to (sometime?) be evaluated before the condition of the clause is checked. If it's false, all results from the body will just be thrown away.

So, I'm not quite sure if this will have any effects.

Do you have the feeling Hydrogen is too slow and needs to be optimized? I can't recall that someone did a proper profiling of the core.

oddtime:
> I'm not so much into the C++ compiler... So, I'm not quite sure if this will have any effects.

Neither do I actually.
At least this checks explicitly what the compiler maybe already would do.

> Do you have the feeling Hydrogen is too slow and needs to be optimized? I can't recall that someone did a proper profiling of the core

I can't say that I have this feeling, maybe optimization could be appreciated on slower computers. I saw some "todo" comments in the code about inefficiency (overwriting objects each time...).

cme:
> I'm not so much into the C++ compiler. But it features some smart optimisation stuff causing the body of `if` clauses to (sometime?) be evaluated before the condition of the clause is checked. If it's false, all results from the body will just be thrown away.

This (or something similar) will happen *sometimes* but usually only for code where the body of the 'if' is trivial enough (including such criteria as not containing any non-inlinable function calls) so it wouldn't happen in either of these cases. 

It's hard to tell whether the first part of the change would help performance (because it'll depend on the pattern and how many notes have a non-1.0 probability; it might help some where everything has probability of 1.0, but might make it slower for some patterns with distributed probabilities), but the second part of the change is almost certainly a performance win: the getGaussian function has an unbiased unpredictable branch, so avoiding that with fairly predictable branch is a win.

But... overall the impact of this will be negligible because it's only executed once per note played, so that's really a tiny number of times in the overall time taken in audio processing. To give you a rough idea, here's some profiling data from the audio thread, playing the GM_kit_Diddley.h2song demo (a reasonably busy song) on my 2012 Macbook Air:

[sample-out.txt](https://github.com/hydrogen-music/hydrogen/files/5837834/sample-out.txt)

(built in release mode, with inlining disabled so that the call tree makes more sense)

It's random sampling so has some error, but gives a fairly accurate overall picture. The number of counts next to a location is roughly proportional to the time occupied by the code (cumulative by call tree).

The thread was sampled 854,148 times, and of that 821,130 (96.13%) were in a semaphore wait, so the thread was entirely idle.

Only 23,944 samples (2.8% of the time) was spent in audioEngine_process. Of that 18116 (2.12% of the total) was spent in renderNote, mostly in ADSR interpolation, which makes sense as it has work to do for every sample.

The call site for audioengine_process_playNotes (where this change would show up) shows a count of 258, or 0.03% of the total CPU time available to the audio thread. That's the *whole* of audioEngine_process_playNotes and everything called from it.


When something's as far from the performance-critical code as this, it makes sense to focus on maintainability rather than performance. I think your changes make the code more readable and that's probably the most important thing here :)

oddtime:
Thanks for the analysis!

theGreatWhiteShark:
> I can't say that I have this feeling, maybe optimization could be appreciated on slower computers. I saw some "todo" comments in the code about inefficiency (overwriting objects each time...).

Since some of these TODO comments are about 20 years old we don't really have to mind them. Of course, the more efficient an application the better. But when looking at the way the code of the audio engine is written/structured, I thinks it's more sustainable for the development of Hydrogen to go for readability and accessibility than for optimization.

That being said, there is one part of the application I found to be almost unusable because it's so inefficient: the LADSPA plugins. Using `Non-Mixer` as a host I run over 100 of such plugins when producing music without any xruns. But using just the four ones  in Hydrogen brought my laptop to its knees. But since no one has complained yet, this issue can wait for later.

oddtime:
Interesting to know about LADSPA.
Months/years ago I remember similar problems but I thought it was my old slow CPU fault.
Now when I build H2 cmake doesn't find LADSPA so I am not using it (even if I have some plugins installed, just didn't spend time to solve the problem since I didn't use the feature)

oddtime:
There was another thing I didn't understand and probably you can explain, if I take advantage of your knowledge :)
When I load a bigger drumkit I see that the used ram memory increases (from ubuntu task manager), but if I switch to the previous kit again, the used memory does not decrease

theGreatWhiteShark:
> When I load a bigger drumkit I see that the used ram memory increases (from ubuntu task manager), but if I switch to the previous kit again, the used memory does not decrease

That's most probably a bug (and luckily one we can make a unit test of). Can you open an issue?

cme:
> That being said, there is one part of the application I found to be almost unusable because it's so inefficient: the LADSPA plugins. Using `Non-Mixer` as a host I run over 100 of such plugins when producing music without any xruns. But using just the four ones in Hydrogen brought my laptop to its knees. But since no one has complained yet, this issue can wait for later.

This is interesting. The plugins all get run sequentially in the audio thread, and there'll be some buffer copying overhead, but I wouldn't have expected the difference between 4 and 100. The host doesn't have much say in what happens other than arranging buffering and execution time -- the actual processing time is down to each individual plugin. Does your laptop have many cores? I can imagine Non-Mixer doing a better job of scheduling and parallelising, but there doesn't seem to be that much scope for significant difference other than that, unless I'm missing something?

theGreatWhiteShark:
Okay. So, the full story is: During a congress (I think it was 34c3) I decided to write some LADSPA plugins myself in Rust (without any knowledge of Rust but in the physical vicinity of a Rust contributor/developer group). I did port a very simplistic delay plugins from the LADSPA examples. But when I used just a single instance with Hydrogen I got cracking and xruns. First, I thought I messed something up. Deeper inspection showed that it was just about half as efficient as the C equivalent (I don't recall the actual number but almost negligible with respect to the simplicity of the plugin). Also, when loading it into `non-mixer` I could run a couple of them without any problems. But soon after the NSM and JACK support of H2 become increasingly annoying to me and I decided to fix them. Well, and since then other things are just keep coming. So, I hadn't a change to look into the subject again (I gave the topic a low priority since no user complained).

> Does your laptop have many cores?

Just two.

cme:
> Okay. So, the full story is: During a congress (I think it was 34c3) I decided to write some LADSPA plugins myself in Rust (without any knowledge of Rust but in the physical vicinity of a Rust contributor/developer group). I did port a very simplistic delay plugins from the LADSPA examples. But when I used just a single instance with Hydrogen I got cracking and xruns. First, I thought I messed something up. Deeper inspection showed that it was just about half as efficient as the C equivalent (I don't recall the actual number but almost negligible with respect to the simplicity of the plugin). Also, when loading it into `non-mixer` I could run a couple of them without any problems. But soon after the NSM and JACK support of H2 become increasingly annoying to me and I decided to fix them. Well, and since then other things are just keep coming. So, I hadn't a change to look into the subject again (I gave the topic a low priority since no user complained).

Ah, that sounds like there's a genuine bug in there somewhere. I don't suppose you would have been using the ALSA driver at that point? 'cause #881 would be very likely to cause such behaviour.

theGreatWhiteShark:
> I don't suppose you would have been using the ALSA driver at that point?

Yeah, probably I did.

cme:
One of the potential behaviours of that bug (depending on hardware) was to vastly reduce the amount of time available to do audio processing in, so hopefully that's what it was. I'll try later on to see how many plugins I can get running on my raspberry pi without it underrunning!

cme:
Ugh, so this video is awful because Kazam doesn't seem to be capable of processing such a large screen area on something as humble as a RasPi4, and also it didn't seem to capture the sound at all either (though I had the volume turned down also). But in this you can see the Hydrogen CPU meter (as well as wmcpu and top), with  4 plugins (including the largest EQ I could find). Console output visible showing no XRUNs, and top showing Hydrogen using about 15% of a CPU.

Also it's a debug build, so a release build would have lower CPU usage (though that won't affect the plugins as they're release builds from Raspbian).

[Hydrogen on Raspberry Pi 4](https://youtu.be/AF3RXIyzeyU)

theGreatWhiteShark:
@cme How did your patterns in the SongEditor became so colorful? :smile:

But it seems to work now.

I had the idea of increasing the number of LADSPA effects in the `CMakeLists.txt` and see how much I could use. But the GUI can not handle more than 4 of these. See #1151

cme:
> @cme How did your patterns in the SongEditor became so colorful? 😄

image

It's the rather oddly named "Coloring method for Song Editor elements" preferences item.

theGreatWhiteShark:
Oh, an existing feature. It's surprising how many of them I still do not know

Issue #192, opened 2014-08-06
Better visual feedback for current note & pattern being triggered as they are playing


ViktorNova:
Something I am constantly finding myself doing is squinting to see where in the pattern or song editor that tiny little cursor at the top is floating to see where in the song or pattern I am at the moment. It is also hard to line it up iwhile watching the thing play when there is a page and a half of instruments or patterns loaded and the instrument or pattern being played is nowhere near the only position indicator

I propose 2 new things to address this:
1. A vertical line that extends down from the cursor and goes to the bottom of the song/pattern editor area, moving with the cursor and showing the current position through the whole page. Almost every sequencer does this and it should be pretty easy to implement, I'm going to take a stab at this one this week
2. As a sample is being triggered, that note glows a different color for a moment to let you know it is currently sounding. In the song editor, any patterns being played glow a different color until they are finished playing. Maybe they even flash or pulsate with the beat =)

Any thoughts about this?

theGreatWhiteShark:
1. was implemented in #1544 

For the 2. part: it would be in principle possible to implement such a thing but it would probably introduce a significant increase of Hydrogen's CPU load. I can see where you are coming from but I think highlighting the current note is sufficiently done using the full-height playhead. 

I will close this issue for now. But feel free to reopen it if there is something missing or the implementation #1544 is not sufficient for your needs.

Issue #1103, opened 2021-01-16
Show other pattern notes in Stacked Mode / Virtual Patterns


cme:
Hydrogen version *          :
Operating system + version  : 
Audio driver + version      : 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Editing a pattern in "stacked" mode (or editing notes in the 'owner' pattern of a virtual pattern group), we can hear the notes of the other patterns active during playback. This is great for editing patterns that go together, but can lead to some trial and error to get exactly what you want.

It would be handy if the Pattern Editor would *show* the notes from the other patterns (in some different representation to make it clear they can't be interacted with) that are active.

This would also be really useful for teaching, and figuring out handing.

theGreatWhiteShark:
I never used the `stack` mode before and it was a bit trial and error first. We should add in the manual that you have to press [Ctrl]+left mouse to add/remove patterns to/from the stack. Is there more to it I missed out?

> It would be handy if the Pattern Editor would show the notes from the other patterns (in some different representation to make it clear they can't be interacted with) that are active.

How about showing them in a light greyish color shifted some 2-3 pixels left and up? Note: the patterns could have different lengths too. Maybe continuing the grid in a light color while keeping the general background of the pattern editor could indicate this difference in length very well.

oddtime:
> I never used the stack mode before and it was a bit trial and error first. 

I didn't know either.
How about adding "add to stack" and "remove from stack" in the pop up menu of the song editor?

cme:
> I didn't know either.
> How about adding "add to stack" and "remove from stack" in the pop up menu of the song editor?

I was thinking something similar.

Also since stacked mode doesn't apply to Song Mode, perhaps the button should be greyed out in Song Mode.

theGreatWhiteShark:
And a short tooltip like "play back other selected patterns while editing" would be nice too.

cme:
#1310 applied to master

Issue #1018, opened 2020-12-13
Design of Playback track


theGreatWhiteShark:
Hydrogen version *          : 1.0.1
Operating system + version  : Linux

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

I just used our Playback Track functionality and I was a little bit underwhelmed by the way it is displayed right now:

![orig](https://user-images.githubusercontent.com/14164547/102003573-4bc2f500-3d09-11eb-8de0-3ae7f45c6db7.png)

Problems:

1. The ruler is not visible
2. The playhead is not visible
3. One can not see, interact or toggle the Timeline although it can be either active or not
4. The fader for the Playback track is not located in the Mixer and initialized with both volume set to 0 and muted

Possible solutions:

1. How about we add the ruler at the bottom of the playback track
![scale](https://user-images.githubusercontent.com/14164547/102003630-d0157800-3d09-11eb-8d7b-0b9312f7cdf8.png)

2. The playhead I would make bigger and cover the whole playback track
![scale_playhead](https://user-images.githubusercontent.com/14164547/102003641-f20efa80-3d09-11eb-94d1-8923dd61aeb6.png)

While at it: how about we highlight all the patterns of the pattern group currently playing? It's more eyecatching than searching for the little playhead

![scale_playhead_marked_patterns](https://user-images.githubusercontent.com/14164547/102003652-0f43c900-3d0a-11eb-93c7-880e1016da39.png)

3. For the different speds of the Timeline I am not sure right yet. We could make a different ruler at the top (similar as for not having the Playback track active right now).
![scale_playhead_marked_patterns_tempo](https://user-images.githubusercontent.com/14164547/102003681-66e23480-3d0a-11eb-8563-9b02b7f54bdc.png)

Or it could be integrate in the bottom ruler

![scale_playhead_marked_patterns_tempo](https://user-images.githubusercontent.com/14164547/102004564-1b338900-3d12-11eb-9667-b123a043bd12.png)

4. I would like to move the Mixer fader for the Playback track into the Mixer. It does feel odd it is not present there. Maybe to the left of the Master. Instead, I would reintroduce the buttons of the Timeline. I would also initialize it not muted and with volume set to 20%.

mauser:
Hi! 

good points! The current implementation is quite rough, i was pretty happy to get it ready for the 1.0 release as time was quite short,as always :smile: 

[...]

PR #1107, opened 2021-01-19
Show a full-height cursor for song position

Ratfink:
Previously, Hydrogen had only a small marker on the song editor position ruler to indicate the playback head position in the song.  This PR extends the line all the way down the song editor, as discussed in #192 (item 1 in the first comment).

Screenshot from 2021-01-18 22-51-46

Still a work-in-progress, as I know of a few minor issues it has still.

- [x] The line is offset by half a (logical) pixel from the arrow on my HiDPI monitor.
- [ ] Occasionally, and mostly at very high BPM, the line will get slightly out-of-sync between the ruler and the song editor.
- [x] The arrow and line are different colors, which is more obvious now that the line is bigger.
- [ ] The cursor's movement is quite jerky, but that's not new; ideally we would change the update timer's tick rate based on the amount of time represented by one pixel on the song editor.
- [ ] The line still needs to be extended over the automation path, though the dark line won't show up very well on its dark background.

### DIFF
--- a/src/gui/src/SongEditor/SongEditor.cpp
+++ b/src/gui/src/SongEditor/SongEditor.cpp
@@ -824,9 +824,9 @@ void SongEditor::paintEvent( QPaintEvent *ev )
 			// nessun pattern, uso la grandezza di default
 			fPos += (float)pHydrogen->getTickPosition() / (float)MAX_NOTES;
 		}
-		uint x = (int)( m_nMargin + fPos * m_nGridWidth - 11 / 2 );
+		uint x = (int)( m_nMargin + fPos * m_nGridWidth );
 		painter.setPen( QColor(35, 39, 51) );
-		painter.drawLine( x + 5 , 0, x + 5 , height() );
+		painter.drawLine( x, 0, x, height() );
 
 		// Update the position ruler to redraw the cursor there as well.
 		// Without this they can get separated from each other if e.g. user
@@ -2116,11 +2116,11 @@ SongEditorPositionRuler::SongEditorPositionRuler( QWidget *parent )
 
 	createBackground();	// create m_backgroundPixmap pixmap
 
-	// create tick position pixmap
-	bool ok = m_tickPositionPixmap.load( Skin::getImagePath() + "/patternEditor/tickPosition.png" );
-	if( ok == false ){
-		ERRORLOG( "Error loading pixmap" );
-	}
+	// Create tick position QPainterPath
+	m_tickPositionPath = QPainterPath( QPointF( 0, 6 ) );
+	m_tickPositionPath.lineTo( -4, 2 );
+	m_tickPositionPath.lineTo( 4, 2 );
+	m_tickPositionPath.closeSubpath();
 
 	update();
 
@@ -2360,10 +2360,13 @@ void SongEditorPositionRuler::paintEvent( QPaintEvent *ev )
 	painter.drawPixmap( ev->rect(), *m_pBackgroundPixmap, srcRect );
 
 	if (fPos != -1) {
-		uint x = (int)( m_nMargin + fPos * m_nGridWidth - 11 / 2 );
+		uint x = (int)( m_nMargin + fPos * m_nGridWidth );
 		painter.setPen( QColor(35, 39, 51) );
-		painter.drawLine( x + 5 , 0, x + 5 , height() );
-		painter.drawPixmap( QRect( x, height() / 2, 11, 8), m_tickPositionPixmap, QRect(0, 0, 11, 8) );
+		painter.drawLine( x, 0, x, height() );
+		painter.translate( x, height() / 2 );
+		painter.fillPath( m_tickPositionPath, Qt::black );
+		painter.drawPath( m_tickPositionPath );
+		painter.resetTransform();
 	}
 
 	if ( pIPos <= pOPos ) {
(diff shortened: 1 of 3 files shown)


theGreatWhiteShark:
> The arrow and line are different colors, which is more obvious now that the line is bigger.

I think this is more of a feature than a bug. The line could be slightly transparent while the arrow is solid black. This way it highlights the current transport location without being to intrusive.

Just for the sake of completeness: In pattern mode there is also an arrow in the pattern editor indicating the transport position. I was going write that I thought a full-height cursor is not required there. But I'm already not sure anymore. In fact, I just realized ruler can't be used for relocating the arrow hat. :thinking: Maybe they deserves to be aligned (also in the light of #1103 )

theGreatWhiteShark:
> The line still needs to be extended over the automation path, though the dark line won't show up very well on its dark background.

Maybe using a more light grey in there instead?

There are also some complications with the Playback track (see #1018) but that's probably out of scope.

src/gui/src/SongEditor/SongEditor.cpp
```diff
@@ -811,6 +813,28 @@ void SongEditor::paintEvent( QPaintEvent *ev )
...
+		float fPos = pHydrogen->getPatternPos();
+		if ( pHydrogen->getCurrentPatternList()->size() != 0 ) {
+			int nLength = pHydrogen->getCurrentPatternList()->longest_pattern_length();
+			fPos += (float)pHydrogen->getTickPosition() / (float)nLength;
+		}
+		else {
+			// nessun pattern, uso la grandezza di default
+			fPos += (float)pHydrogen->getTickPosition() / (float)MAX_NOTES;
+		}
+		uint x = (int)( m_nMargin + fPos * m_nGridWidth - 11 / 2 );
```
cme, line 827: 818-827 here are duplicated from the ruler's paintEvent. Can it be factored out? Possibly by moving to the SongEditorPanel? Or even a static function as it doesn't look like it needs anything from the SongEditor or SongEditorRuler.

src/gui/src/SongEditor/SongEditorPanel.cpp
```diff
@@ -550,6 +550,7 @@ void SongEditorPanel::updatePlaybackTrackIfNecessary()
 void SongEditorPanel::updatePositionRuler()
 {
 	m_pPositionRuler->createBackground();
+	m_pPositionRuler->update();
```
cme, line 553: It looks like the only call site for `updatePositionRuler` is `updateAll`, which has a call to `updateTimelineUsage` just after this, which *also* calls createBackground. So I think the ruler is needlessly repainted (to its background pixmap) twice. 

Also the other component widgets are explicitly `->update()`ed in updateAll, so it would make sense to do that there rather than here?

src/gui/src/SongEditor/SongEditor.cpp
```diff
@@ -2357,7 +2381,9 @@ void SongEditorPositionRuler::paintEvent( QPaintEvent *ev )
 
 void SongEditorPositionRuler::updatePosition()
 {
-	HydrogenApp::get_instance()->getSongEditorPanel()->updateTimelineUsage();
+	SongEditorPanel *pSEPanel = HydrogenApp::get_instance()->getSongEditorPanel();
+	pSEPanel->updateTimelineUsage();
+	pSEPanel->getSongEditor()->update();
```
cme, line 2386: This runs constantly even when the audio engine isn't running. We should probably change that, but until now it's not been a problem -- the ruler is very fast to update() because it's small and has a pixmap as backing store. But the song editor is larger and may render more components each paint event (keyboard cursor, selection / move) so would be better to avoid redrawing it.

Keeping a static fLastPos in the SongEditorPanel and making `update()` conditional on that differing from the current pos would do the job fine :)

cme (commented):
This really reminds me that I should rename all the 'Cursor' members for the keyboard cursor support to 'KeyboardCursor' :)

src/gui/src/SongEditor/SongEditorPanel.cpp
cme: > It looks like the only call site for `updatePositionRuler` is `updateAll`, which has a call to `updateTimelineUsage` just after this, which _also_ calls createBackground. So I think the ruler is needlessly repainted (to its background pixmap) twice.

I should clarify -- that's a comment on the existing code, not yours :)

src/gui/src/SongEditor/SongEditor.cpp
Ratfink: There actually is a dependence in line 827 on the two members `m_nMargin` and `m_nGridWidth`, whose values are duplicated between the `SongEditor`, `SongEditorPositionRuler`, and in the case of the grid width, `AutomationPathView`.  So maybe that block should be a member of the SongEditorPanel, and those members moved as well?

cme: > There actually is a dependence in line 827 on the two members `m_nMargin` and `m_nGridWidth`, whose values are duplicated between the `SongEditor`, `SongEditorPositionRuler`, and in the case of the grid width, `AutomationPathView`. So maybe that block should be a member of the SongEditorPanel, and those members moved as well?

Ah, sorry I didn't spot those! Yes. What *really* wants to happen here is that there should be a common ancestor class of AutomationPath, SongEditor, SongEditorRuler etc. but failing that, cut-and-paste is fine (this is probably a controversial opinion!) because it means the code that should be commonised is easier to spot and easier to unify when such a class exists :)

theGreatWhiteShark:
Superseded by https://github.com/hydrogen-music/hydrogen/pull/1544. Thanks for the idea!

PR #1114, opened 2021-01-23
Toggle playback at cursor

theGreatWhiteShark:
When pressing [Ctrl] + [space] the playback gets toggled at the current cursor position in the song editor, the pattern editor, and the note properties ruler (it's now possible to start playback of a pattern not just from the start but from an arbitrary position).

In addition, when the cursor is within the `SongEditor` song mode is activated. If the cursor is in the `PatternEditor` (or `NotePropertiesRuler`) pattern mode is activated. I found this behaviour quite intuitive because it saves a shortcut for switching between song and pattern mode and one does not have to fear to loose the current playback position in the song editor either (using the new command). When switching to pattern mode the current tempo of the timeline is maintained.

@cme I have not incorporated anything for checking whether the cursor is hidden or not. Because a) I wasn't sure where to query if the cursor in the *Editor is hidden and b) I think it's useful to start playback at the position last clicked too. What do you think?


cme:
I like this, this will be really handy!

Only issue is that on macOS the Command+Space sequence is the default binding for Spotlight (global search). Yet more motivation for keyboard map editing! :D

> In addition, when the cursor is within the `SongEditor` song mode is activated. If the cursor is in the `PatternEditor` (or `NotePropertiesRuler`) pattern mode is activated. I found this behaviour quite intuitive because it saves a shortcut for switching between song and pattern mode and one does not have to fear to loose the current playback position in the song editor either (using the new command). 

Agreed, I really like this. Using the shortcut in either editor has a clear unambiguous user intention, and switching mode meets that intention. 👍 

> @cme I have not incorporated anything for checking whether the cursor is hidden or not. Because a) I wasn't sure where to query if the cursor in the *Editor is hidden and b) I think it's useful to start playback at the position last clicked too. What do you think?

I think it's right the way it is. Last click position is almost certainly the right place to start from if there's no keyboard cursor shown. 


Question about structure: since this does different things for PatternEditor and SongEditor, and has to use `inherits()` to figure out which, would it be better if this could live in the `keyPress()` methods of the PatternEditor and SongEditor classes? There isn't currently a `keyPress()` method on the PatternEditor class itself, just on all the individual derived classes, but one could be added and a super call added. Which, now I come to think of it, is probably a good idea anyway.

theGreatWhiteShark:
> Only issue is that on macOS the Command+Space sequence is the default binding for Spotlight (global search). Yet more motivation for keyboard map editing! :D

Ah, of course. On Windows this is probably something like [Super] + [Space] and [Alt] + [Space] is already taken in at least XFCE and Unity. Hmm... Shall we introduce OS specific bindings? Or switch to another key? (I slightly tend towards the first).

Should this command indeed **toggle** the playback? After all there is no different between a general "stop playback" and "stop playback at cursor position". Instead, the command could just locate to a different location during playback without stopping.

I also experience some weirdness when using the JACK driver. But I don't want to touch this right before the release. Especially, because I intend to make the transport within Hydrogen and the one of JACK compatible as soon as the refactoring of the AudioEngine is done. So, this might be an experimental feature. (Or could wait after the release)

> Question about structure: since this does different things for PatternEditor and SongEditor, and has to use inherits() to figure out which, would it be better if this could live in the keyPress() methods of the PatternEditor and SongEditor classes? There isn't currently a keyPress() method on the PatternEditor class itself, just on all the individual derived classes, but one could be added and a super call added. Which, now I come to think of it, is probably a good idea anyway.

As far as I understood the key handling in Qt the key event passes `MainForm` (at a quite early stage) and is then propagated into the other classes/widgets as well. Since the original command bound to [space] does eat the event I thought it would be a good idea to do the same. Maybe there are some hidden things to be discovered don't do it.

theGreatWhiteShark:
> I also experience some weirdness when using the JACK driver. But I don't want to touch this right before the release.

Well, who am I kidding? Of course I look into this (since I want to use it myself) :D

- [ ] Relocation in the presence of an external Timebase master has to be done twice. After the first one the timebase master sends the tempo for this particular relocation position, this might differ from the current one in Hydrogen, triggers a recalculation of the ticksize which moves the playhead to another location. Afterwards, a second relocation is done to insist on the location asked for by the user. (This should cover for most cases. In theory this relocation should be done until the transport location converges.)
- [ ] The code does not seem to work with `non-timeline`. But I'm 95% sure this is not my fault but a bug in `non-timeline` since the relocation does work when done via click on the song position ruler. The only difference is that for the latter transport was already rolling while the JACK message of the relocation is identical. JACK transport shouldn't be markovian. And with Ardour there is no problem at all.

theGreatWhiteShark:
> The code does not seem to work with non-timeline. But I'm 95% sure this is not my fault but a bug in non-timeline since the relocation does work when done via click on the song position ruler. The only difference is that for the latter transport was already rolling while the JACK message of the relocation is identical. JACK transport shouldn't be markovian. And with Ardour there is no problem at all.

This indeed was a bug in `non-timeline` https://github.com/theGreatWhiteShark/non/pull/1

theGreatWhiteShark:
The shortcut does now only starts the transport and just relocates if it's already rolling.

@cme For MacOS I changed the shortcut to [Alt] + [space]. Does this work or is there some other system binding shadowing it?

theGreatWhiteShark:
> Relocation in the presence of an external Timebase master has to be done twice. After the first one the timebase master sends the tempo for this particular relocation position, this might differ from the current one in Hydrogen, triggers a recalculation of the ticksize which moves the playhead to another location. Afterwards, a second relocation is done to insist on the location asked for by the user. (This should cover for most cases. In theory this relocation should be done until the transport location converges.)

Pfff. I just realized the bug in the relocation in the presence of an external JACK timebase master does not depend on this keyboard shortcut. In my test scenario it happens when relocating via mouse too. The JACK support is just too brittle. I will take care of this in v1.2 (by somehow making the transport frame-based instead of tick-based without altering the current behavior or the interface to the `Hydrogen` class). This will make JACK transport and the internal one compatible.

Since the only difference between the relocation provided in here and the relocation via mouse is that starting the JACK transport and requesting for a position change does happen at the same time and not a process cycle apart, I would vote for merging it anyway. Nothing which works will be broken and the user can disable the timebase support in the Preferences in order to seemlessly use Hydrogen with Ardour.

PR #1118, opened 2021-01-24
UX tweaks (as discussed in #1097 )

cme:
UX tweaks for the pattern editors.
  - alt + up/down arrows to fine-tune note properties 
  - alt + click in pattern editors for fine-grained placement of notes
  - show messages on LCD for note properties edited with the arrow keys (but not for multiple notes)
  - Also fix (theoretical) bugs with qBound argument ordering.

### DIFF
--- a/src/gui/src/PatternEditor/DrumPatternEditor.cpp
+++ b/src/gui/src/PatternEditor/DrumPatternEditor.cpp
@@ -165,7 +165,7 @@ void DrumPatternEditor::mouseClickEvent( QMouseEvent *ev )
 	if (row >= nInstruments) {
 		return;
 	}
-	int nColumn = getColumn( ev->x() );
+	int nColumn = getColumn( ev->x(), /* bUseFineGrained=*/ true );
 	int nRealColumn = 0;
 	if( ev->x() > m_nMargin ) {
 		nRealColumn = ev->x() / static_cast<float>(m_nGridWidth) - m_nMargin;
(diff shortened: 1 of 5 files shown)


theGreatWhiteShark:
Works fine!

Just a note: When selecting a note and switching to a different instrument, the note properties of this instrument can not be altered anymore.
npr_sel

I was a little bit confused for a second. But it is quite plausible what's happened there. Then again, how about always allowing to alter all properties of a selected instrument unless some notes of this very instrument are selected?

cme:
> I was a little bit confused for a second. But it is quite plausible what's happened there. Then again, how about always allowing to alter all properties of a selected instrument unless some notes of this very instrument are selected?

Not sure I'm interpreting the wording right. You're proposing showing all *selected* notes in the NotePropertiesEditor, unless the selection includes the current instrument?

I think if the NotePropertiesEditor were to show selected notes instead of current instrument notes, it would probably be better to show *only* selected notes if a selection exists, and the currently selected instrument's notes if no selection exists; that way the only rule / state the user needs to worry about is whether there's anything selected or not?

theGreatWhiteShark:
> You're proposing showing all selected notes in the NotePropertiesEditor, unless the selection includes the current instrument?

No. The idea was to allow editing of a property if none of the notes of the current instrument were selected. In the picture above: I would be able to alter all four notes of the first instrument, the one of the second, and just the right one of the fourth.

But upon revisiting I like the idea of your current implementation way more than mine. Only allow altering of selected notes and many ways to cancel the selection, like [esc] or a single click. Straight-forward and simple.

oddtime:
Can I ask what is the use of fineGrain?

cme:
> Can I ask what is the use of fineGrain and how it works basically in the code?

The "fine grained" note dragging / clicking in pattern editor (while holding Alt) is intended to allow fine-grained timing adjustment or placement, just bypassing the grid res to place at / move by single-tick resolution.

> another thing, in case you missed: if you select a note, then move the note drawing to a different position and you press ctrl when the mouse is still pressed the note is pasted and the old note not canceled, but the '+' icon doesn't appear unless you move the note drawing again (still pressing the mouse).

Hmm, good spot, I'll take a look.

PR #1156, opened 2021-02-02
Fix drumkit part of SoundLibraryPanel

### LINKED ISSUE #1139: Dialog: "The current loaded song missing his soundlibrary. Please load a existing soundlibrary first"
Hydrogen version * : 1.0.1
Operating system + version : Windows 10 Home v2004 build 19041.746
Audio driver + version : Microsoft generic USB audio v10.0.19041.388, via PortAudio

In the instrument rack, when I right-click on any kit and select 'Properties', I get this error message:

"The current loaded song missing his soundlibrary. Please load a existing soundlibrary first"

It is not clear what this means, or how to fix it. The current song was created from scratch, every instrument in the song drumkit has sample(s) assigned, and the song's drumkit was saved into the sound library successfully in this editing session.

I discovered that right-clicking any kit and selecting 'Load' in the context menu 'fixes' this issue for _all_ kits, at the expense of overwriting the current song's drumkit.

The song is not under .hydrogen, in case that matters.

### LINKED ISSUE #1141: Drumkit -> Save automatically uses same name as system kit / Drumkits with same name as system kits cause bugs
Hydrogen version * : 1.0.1
Operating system + version : Windows 10 Home v2004 build 19041.746
Audio driver + version : Microsoft generic USB audio v10.0.19041.388, via PortAudio

- Create a new project
- Drumkits -> New
- Clear all instruments? [OK]
- Drumkits -> Save

A new user drumkit appears in the sound library, with the same name as the previously loaded system drumkit (e.g. GMRockKit)

Try to load the previously loaded system drumkit:
- Open the sound library
- Right-click on a differently-named drumkit (e.g. TR808EmulationKit), Load
- Right-click the previously loaded system drumkit (e.g. System drumkits -> GMRockKit), Load
- The empty user drumkit is loaded instead

Try to delete the empty user drumkit:
- Right-click the user drumkit (e.g. User drumkits -> GMRockKit), Delete
- Dialog: "It is not possible to delete the currently loaded drumkit: "GmRockKit". To delete this drumkit first load another drumkit."

Since there is no way to rename user drumkits without copying them, the only way to delete the user kit or to load the shadowed system kit, is to select the *other* built-in kit (e.g. TR808EmulationKit) then delete the user kit.

theGreatWhiteShark:
There are quite a lot of things going wrong in the `SoundLibraryPanel`

- [x] If a drumkit does exist with the same name in both the user and the system folder, the user one was always returned. Even if the other was explicitly selected. Fixes #1141
- [x] Use proper popup windows if a drumkit name does not match the ones in the system and user dir. Fixes #1139 
- [x] Check whether drumkit of the current Song in `SoundLibraryPanel::on_drumkitPropertiesAction()` is a system or user drumkit (and choose the right one) (and figure out why this is required in the first place)
- [x] `SoundLibraryPanel::on_drumkitDeleteAction()` will always create the user level drumkit if two drumkits with the same name are present on system and user level
- [x] `SoundLibraryPanel::on_drumkitDeleteAction()` will not delete a user level drumkit if a system level drumkit is present with the same name
- [x] If there are two drumkits with the same name on system and user level, both are highlighted in the panel
- [x] The dropdown list in the export window does not highlight which drumkit is on user level and which on system level. In case our scenario the name will the listed twice
- [x] In `SoundLibraryExportDialog::on_exportBtn_clicked()` always the user level drumkit will be exported
- [x] Drag inserting of instrument does not work for system drumkits
- [x] Check the MainForm -> Drumkit items. Maybe there are some additional bugs.

The general behavior of loading user level drumkits over system ones I do not want to touch. E.g. if there are two `GMRockKit`, one on user level and one on system level, and Hydrogen is invoked using `hydrogen -k GMRockKit` the user level one will be loaded. The changes are only intended for parts of the GUI where the user is able to explicitly select the system level one.

---

Edit:
- [x] ~Fix memory leakage when adding an instrument via drag and drop.~ Fixed to some degree. The remainder was delegated to #1190


theGreatWhiteShark:
Hey @cme,

I'm rather clueless about a bug I encountered in here.

I would like to change 

```c++
void PatternEditorInstrumentList::dragEnterEvent(QDragEnterEvent *event)
{
	INFOLOG( "[dragEnterEvent]" );
	if ( event->mimeData()->hasFormat("text/plain") ) {
		Song *song = (Hydrogen::get_instance())->getSong();
		int nInstruments = song->getInstrumentList()->size();
		if ( nInstruments < MAX_INSTRUMENTS ) {
			event->acceptProposedAction();
		}
	}
}
```

into

```c++
void PatternEditorInstrumentList::dragEnterEvent(QDragEnterEvent *event)
{
	INFOLOG( "[dragEnterEvent]" );
	if ( event->mimeData()->hasFormat("text/plain") ) {
		Song* pSong = Hydrogen::get_instance()->getSong();
		int nInstruments = pSong->getInstrumentList()->size();
		if ( nInstruments < MAX_INSTRUMENTS ) {
			event->acceptProposedAction();
		} else {
			QMessageBox::critical( this, "Hydrogen", tr( "Unable to insert further instruments. Maximum possible number" ) +
								  QString( ": %1" ).arg( MAX_INSTRUMENTS ) );
			event->ignore();
		}
	}
}
```

but when the `QMessageBox` is closed (and the `dragEnterEvent()` function already has returned) Hydrogen segfaults. The same does happen when I introduce (just for the sake of debugging) a `QMessageBox` in `SongEditorPatternList::dragEnterEvent()`.

I couldn't find anything in the Qt5 doc indicating that this is the expected behavior but couldn't find anything related on stackoverflow either. Am I using this the wrong way or is it a bug? If the latter, we should ensure it's a problem of the Qt5 code and not of our side.

cme:
Hmm, I'm not sure what the expected behaviour there would be. The trouble is that QMessageBox::critical runs its own event loop, so that doesn't return until after the dialog box is closed. That re-enters the event processing system, so I'm not sure that `event` inside here would necessarily be valid anymore. You can probably sidestep that issue by just moving the `event->ignore()` call above the `QMessageBox`, so that the event is fully handled before entering another event loop.

It's possible this is a Qt bug that's an edge case, because `dragEnterEvent` is already called from a non-top-level event loop (from the `exec()` method of the QDrag, so exec'ing a QMessageBox means the event loop is re-entered twice, that might be rare.

theGreatWhiteShark:
> It looks like I was wrong about the lifetime of event being an issue. It looks like any QMessageBox from inside a QDrag causes a crash. This happens on Linux but doesn't seem to happen on macOS. Almost certainly a Qt bug I think.

At least it works in `dropEvent`. I will put the code there instead.

I'll have a look at Qt's bug tracker. If it's not too much effort I will be a good boy and report it. Thanks a lot for the clarification.

theGreatWhiteShark:
I also fixed the MainForm > Drumkit > Open dialog.

There is still some minor convenience: When drag&drop inserting an instrument and modifying its content, <kbd>Ctrl</kbd> + <kbd>Z</kbd> followed by <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Z</kbd> does not restore the modified instrument but the (original) inserted one. But would wait till someone complaints about this before fixing it. Feels like efforts for a very minor edge case.

PR #1163, opened 2021-02-04
Change "Copy" into "Duplicate" in song editor pattern right-click menu

oddtime:


theGreatWhiteShark:
Hey @oddtime,

Could your rename the underlying function as well?

(I think "duplicate" makes it indeed more clear. In general, we should not alter the translatable strings too much since it will invalidate parts of existing translations. For languages with many active supports this might be no big deal but others had been one time contributors and those translations keep slowly falling apart. Well, unfortunately, I forget those things every other day too :/ )

oddtime:
Yes, I agree with everything.

theGreatWhiteShark:
Thanks. Could you do the same for the `SE_copyPatternAction()` function called by `patternPopup_duplicate()`?

oddtime:
Ops I forgot to see. I have renamed that and another function

theGreatWhiteShark (approved):
thanks!

PR #1169, opened 2021-02-07
UX tweaks continued

cme:
* Immediate update of mouse cursor when pressing/release Ctrl while dragging notes or patterns
* Note properties editors will show and edit all *selected* notes


oddtime:
May a "clear notes (all patters)" be possible for every instrument as well?

cme:
Now with overwriting of duplicate notes on deselect. 
Todo:
  - [ ] deselect on save etc.

cme:
Hey @oddtime 
> May a "clear notes (all patters)" be possible for every instrument as well?

Currently the instrument menu has:
 - Clear notes
 - Fill notes...
 - Randomize velocity
 - Select notes
 - Copy notes (all patterns)
 - Paste notes (all patterns)
 - Rename instrument
 - Delete instrument

I think adding a third all-patterns option makes it a good idea to restructure a little, to something like:
  - Delete notes        (change name)
  - Fill notes...
  - Randomize velocity
  - Select all notes
  - Edit all patterns...
     - Delete notes
     - Copy notes
     - Paste notes
 - Rename instrument
 - Delete instrument

though I'm worrying now that the extra level of menu hierarchy might end up being unweildy. Hmm.

oddtime:
> though I'm worrying now that the extra level of menu hierarchy might end up being unweildy. Hmm.

Maybe it slows down the editing. However it is fine for me, make it as you prefer.
Alternatively we could have only "Select notes (all patterns)" instead of "Delete..." and "Copy...": then you press ctrl+c to copy or canc to delete.

cme:
> > though I'm worrying now that the extra level of menu hierarchy might end up being unweildy. Hmm.
> 
> Maybe it slows down the editing. However it is fine for me, make it as you prefer.
> Alternatively we could have only "Select notes (all patterns)" instead of "Delete..." and "Copy...": then you press ctrl+c to copy or canc to delete.

I realised we can use a section header in menus, so now they're grouped without needing an extra click (since it's still a fairly compact menu)

image

oddtime:
That's clean!

Issue #946, opened 2020-09-26
[Audio Export with Timing Humanize] unwanted samples at the very end of .wav


oddtime:
Hydrogen version *         v1.0.0 rc1 :
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
This seems to happen using the humanized timing:
I have put many empty patterns at the end of the pattern sequence.
But at the very end of the exported song .wav file (fractions of seconds) sometimes it appears something like the beginning of a sample.

Maybe it could be one note in the first pattern in the sequence being anticipated because of humanized timing.
Turning off the loop song button doesn't help because iit turns on automatically when exporting.

mauser:
Hi oddtime,

it seems plausible that this has something to do with humanization. Could you share a song with us (maybe with a stock drumkit) with which the behavior is (at least sometimes) reproducible?

oddtime:
here it is
https://we.tl/t-jV956zcE51

oddtime:
Hi Mauser, may be this a solution?

in `src/core/Hydrogen.cpp` line 3110 in` void Hydrogen::startExportSession(`
>	pSong->set_loop_enabled( true );

replaced by

>	pSong->set_loop_enabled( false );

It seems to work, but I don't know if that causes other problems

mauser:
Hi @oddtime ,

yes, this seems to help! I've play around with loop_enabled(false) and stumbled directly into another problem (#1005 ), but this shouldn't be related to this export situation, it also appears in song mode. I will see if deactivating the loop mode has other negative issues..

oddtime:
Hi @mauser 
Can we integrate the change now, since #1005 seems independent?
I can make the pull req if you don't have time

theGreatWhiteShark:
This was introduced in a fix of a problem https://github.com/hydrogen-music/hydrogen/commit/1b285966bb3c87c3d0bb672606ec34a6623965bd (which one is unfortunately not mentioned in the commit message).

PR #1170, opened 2021-02-10
remove ERRORLOG in Hydrogen::findPatternInTick

### LINKED ISSUE #1005: Error message: "[findPatternInTick] tick = 192. No pattern list found" when playing in Song mode without loop enabled"
Hydrogen version *          :  Hydrogen 1.1.0-'6d8ca1a3'

When Hydrogen plays a simple pattern with size 8 and res 32 in song mode, but the loop mode is deactivated, an error message appears:

`(E) ::int H2Core::findPatternInTick(int, bool, int*) [findPatternInTick] tick = 192. No pattern list found` 

theGreatWhiteShark:
fixes #1005

### DIFF
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -1983,8 +1983,6 @@ inline int findPatternInTick( int nTick, bool bLoopMode, int* pPatternStartTick 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


oddtime:
Thanks!
Since this bug was found initially in #946, do you think we can merge the solution discussed there for the other bug as well?

theGreatWhiteShark:
> Since this bug was found initially in #946, do you think we can merge the solution discussed there for the other bug as well?

I'll have a look

Issue #1007, opened 2020-12-06
Tweaking INSTALL.md


theGreatWhiteShark:
Hydrogen version *          : 362d745

--- 

There are some things I do want to change about the `INSTALL.md` file.

- [ ] Split the essential and the optional packages to install on Linux systems
- [ ] Fix Windows requirements statement (11 years old and suggesting 198MB RAM + keyboard and mouse...) 
- [ ] Move the "create binary package" paragraph to the `DEVELOPERS` file. IMHO this goes beyond the scope of this file and only scares users.
- [ ] Add packages required to generate the documentation
- [ ] Add packages required to support C++17

About the second one: Do we support Windows right now or are the corresponding components still "fallen into despair"?

Opinions?

mauser:
About the Windows compatibility: In my opinion, we have now reached a state where the quality of Linux, OS X and Windows builds differ only slightly. The INSTALL file was written in a time when we had to compile the windows version (including the complete QT environment) on our PCs. I remember that it sometimes took more then one day on my machine to compile QT :) So there were times when there were years between two windows releases..

theGreatWhiteShark:
> I remember that it sometimes took more then one day on my machine to compile QT :)

Yeah, I know. I tried that too once. After more than 20 hours compilation I only got an error about insufficient memory :D

theGreatWhiteShark:
In #972 it was mentioned the `liblash-compat-dev` package could not be found. But this does not seem to be correct https://packages.debian.org/sid/liblash-compat-dev.

> Split the essential and the optional packages to install on Linux systems

They already are (in verbose form). Breaking the apt install command into individual pieces is not necessary and can be done by the user herself (the verbose form lists both the general name and the corresponding package).

> Move the "create binary package" paragraph to the DEVELOPERS file. IMHO this goes beyond the scope of this file and only scares users.

I'll leave it in the INSTALL.md . Packagers are no developers either.

> Add packages required to support C++17

I think it's out of the scope of this file (and not a good idea) to tell the user to get a specific version `gcc`. If their OS does not supply a current one, we can still help them via the issues and discussion.

PR #1171, opened 2021-02-10
tweak INSTALL.md

theGreatWhiteShark:
addresses #1007

@trebmuh & @cme 

There are still some open points:

- Which Windows versions are supported?
- I removed the line stating one has to build Hydrogen from source in order to have JACK support on Mac OS X. Is this correct?
- I would vote for dropping the build on [Linux wiki page](https://github.com/hydrogen-music/hydrogen/wiki/Building-Hydrogen-from-source-(LINUX)) since it's an out of sync copy of INSTALL.md
- What to do with the packaging on [Windows wiki page](https://github.com/hydrogen-music/hydrogen/wiki/Packaging-for-Windows)? It was out of date in 2015 and probably not helpful at all anymore. Shall we keep it or drop it (since we now have the Windows binaries via Appveyor)?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -22,7 +22,7 @@ Hydrogen is supported on the following operating systems:
 
   * Linux/Unix/BSD
   * Mac OS X
-  * Windows 
+  * Windows 10 (maybe 7)
 
 ## Download
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -22,20 +22,7 @@ Hydrogen is supported on the following operating systems:
 
   * Linux/Unix/BSD
   * Mac OS X
-
-Hydrogen is *not* supported on Windows because there are currently no
-Windows developers on the project... and so the Windows components
-have fallen into disrepair.
-
-Hydrogen is known to work well on fairly old systems like this (as of
-2009):
-
-- Pentium III 500 MHz
-- 198 MB RAM
-- Consumer-Grade (cheap) audio card
-- Keyboard and Mouse
-- 1x1 USB MIDI Interface (optional)
-
+  * Windows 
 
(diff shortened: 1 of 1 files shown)


mauser:
> There are still some open points:
> 
>     * Which Windows versions are supported?

Windows XP is *not* supported (recent QT versions have no support for XP). Windows 7 might be supported, but I do not test the binaries on Windows 7. So the "main" version should be Windows 10. 

>     * I removed the line stating one has to build Hydrogen from source in order to have JACK support on Mac OS X. Is this correct?

Yes. The current OS X packages have jack support.

>     * I would vote for dropping the build on [Linux wiki page](https://github.com/hydrogen-music/hydrogen/wiki/Building-Hydrogen-from-source-(LINUX)) since it's an out of sync copy of INSTALL.md

I don't think that this was a real copy, the wiki page was more less maintained side-by-side with the INSTALL.md. The wiki page was migrated from our older trac and assembla wikis, this carried always the current build instructions (and sometimes a bit more detailed then in INSTALL). But it is true that we only need one source, so i would be ok with dropping it. But we should keep a link there to the INSTALL file if the content is of the page gets removed, others may still link there..

> 
>     * What to do with the packaging on [Windows wiki page](https://github.com/hydrogen-music/hydrogen/wiki/Packaging-for-Windows)? It was out of date in 2015 and probably not helpful at all anymore. Shall we keep it or drop it (since we now have the Windows binaries via Appveyor)?

I was already working on new instructions in december, this was more or less the last work i did on Hydrogen before stopping.  The instructions are based on the new msys workflow which were using with appveyor. I have written it locally and need to verify it on a clean machine to see if anything is amiss..

PR #1175, opened 2021-02-13
Update Author list and Donation dialog

theGreatWhiteShark:
How about the following idea: only the active developers (with merge/push privileges) will be listed and for a list of all contributors a link will point to Github. In addition, there is a list of active translators including an email address for users of that particular language to refer to. When calling for translations after the string freeze we will also asked the contributors of the particular translations whether they want to be listed as active ones. I would not show any of the former contributors without asking for permission.

@trebmuh are you fine with this design? Also, how about instead of "French translation" you put the translated (*not* translatable) string in there? This is a little bit inconvenient since a dev instead of a translator has to do it but a new name would need to be added anyway.

---

`DonationDialog` : 

Since we will only have a single Paypal account from now on, I discarded the custom DonationDialog widget and used a default QMessageBox instead.

donate

`AboutDialog` : 

- A listing of the currently active developers was introduced
- the author list was replaced by a list of active translators (which hopefully grows during the string freeze) including their email
- The author text is now contained in a `QTextBrowser` instead of a `QTextEdit` to allow following external links
- Instead of listing all former contributors, a link will now point to the contributor page on Github (which will omit the contributors predating the git version of the repository)
- Translatable strings have been stripped of all the HTML

author1
author2


cme:
I know brevity is important, but I think it's also important to give significant recent contributors their due kudos, particularly those who contributed to *the current release*. So maybe list contributors to the release, with historical contributors via github?

The former list can be extracted automatically from git with:
```
git log releases/1.0..HEAD | grep '^Author:' | sort | uniq -c | sort -n -r | sed 's/^.*Author: //'
```

cme:
> I know brevity is important, but I think it's also important to give significant recent contributors their due kudos, particularly those who contributed to _the current release_. So maybe list contributors to the release, with historical contributors via github?
> 
> The former list can be extracted automatically from git with:
> 
> ```
> git log releases/1.0..HEAD | grep '^Author:' | sort | uniq -c | sort -n -r | sed 's/^.*Author: //'
> ```

Oh heh, I'm in that list three times because I have committed from too many different machines >_<

theGreatWhiteShark:
Alright. Sounds reasonable. I will prepare some sort of script in order for the developers to only call

```bash
script.sh PREVIOUS_GIT_TAG CURRENT_GIT_TAG
```

> Oh heh, I'm in that list three times because I have committed from too many different machines >_<

With some more SED/AWK magic we can filter these out as well :wink: 

```bash
git log origin/releases/1.0..HEAD | grep '^Author:' | uniq -c | sort -n -r | sed 's/^.*Author: //' | awk '!visited[toupper($1)]++' | awk 'BEGIN { FS = " <" }; !visited[toupper($2)]++ {print $1}'
```

theGreatWhiteShark:
@cme I added a separate class and bash script in 9f81822 . Is this o.k. or a little bit too dirty?

cme:
> @cme I added a separate class and bash script in [9f81822](https://github.com/hydrogen-music/hydrogen/commit/9f81822192570a23c5deb8f3fc9e9870ca84c804) . Is this o.k. or a little bit too dirty?

I think it's fine -- if we were generating anything more structured than just text with the data it would be worthwhile to do something more considered, but since it's just text... whatever does the job :D

oddtime:
I missed this, thank you for the mention! I appreciate it :)

trebmuh:
This is good like that. Thanks @theGreatWhiteShark (and sorry for the lag in answering, I'm catching up with internet stuff after 2 weeks off the internet).

Issue #863, opened 2020-05-06
UI refresh


elpescado:
While I was working on [support for high resolution screens](https://github.com/hydrogen-music/hydrogen/pull/487) I noticed that current UI - that is based on pixmaps - has some shortcomings:

* It's not translatable, which means that some parts of UI cannot be localised at all. And even if there was a possibility to load different pixmaps for different languages, translating would require creating separate images, thus a lot of pixel pushing. See instrument editor, most of text there is not translated.

* It's not scalable, so it looks pixelated on hide resolution screens. While technically doable, fixing this requires recreating all pixmaps in high resolution, which is a lot of work.

* Developing new features might require creating lots of graphic assets, which is tedious task and it's hard to achieve consistency (look at pixel font throughout Hydrogen UI, there are slight differences everywhere)

* It looks a bit dated, like a holdout from WinAmp era;)

It would be nice to have UI redesigned. I have some ideas how it could be implemented:

* In code, using [`QPainter`](https://doc.qt.io/qt-5/qpainter.html) API (might be complicated)

* Using SVG (not sure about performance)

* Using QML (not sure if it's possible to use QML from plain Qt app)

* Using Qt CSS implementation (not sure what are limitations)

Of course, that's huge undertaking, but maybe someone is looking for ideas:)

What do you think?

Cheers!

mauser:
Hi! 

Good that you bring this up. I had a discussion with @theGreatWhiteShark recently where we talked also about this topic. With highres display getting more widespread every year, this has been also a big downside of Hydrogen. But we were also talking about the costs and that a major rewrite would be more or less impossible because of the sheer size. 

But maybe we could work on the problem issue by issue? I suspect that using SVG for buttons would be sth. that can be done as a first start. Especially if we do not create a complete new design and try to mimic the look of the current buttons. 

I'm absolutely unsure about these things, and i don't know how other modern QT applications tackle these problems. I suppose that i also can't offer a lot of help.. my personal playground is LV2 support for Hydrogen at the moment, and even the time for that is quite limited :) .

trebmuh:
[HS]

> my personal playground is LV2 support for Hydrogen at the moment

Nice to read :sunglasses:

cme:
Replacing existing images with SVG seems a reasonably straightforward change and worth considering even if it's not regarded as a final solution:

  * Doesn't require major rework or restructuring of the code

[...]

PR #1176, opened 2021-02-16
Improve the LCDSpinBox widget and use it in the instrument editor

Ratfink:
Recently while setting Hydrogen up to control DrumGizmo using MIDI, I got frustrated by how there doesn't seem to be a nicer way to set an instrument's MIDI note number than clicking a tiny button dozens of times.  I got to wondering why the LCD spin boxes didn't support the scroll wheel, and dug into the code to see what it would take to add scroll wheel support.  I found the LCDSpinBox widget, and noticed that it actually *does* support the scroll wheel.  Puzzled, I searched through the code and found that this widget is only used for the big BPM spinbox, not for any of the little ones in the instrument editor.  It doesn't currently support other styles of LCD, and its scroll wheel code lacks bounds checking.

This PR is my attempt to address this issue, fixing the issues with the LCDSpinBox widget and using it in the instrument editor for better interaction.  I've fixed all the bugs in the widget itself that I know of, improved its API so that it's easier to connect increment/decrement buttons, and added an extra type of spinbox for integer values with "Off" as the minimum value.  Using these improvements, I've ported the spin boxes in the instrument editor to use the LCDSpinBox widget.  Also, as an added bonus, now the BPM spinbox can't be made to show values outside the acceptable range of [10, 400] using the scroll wheel.


src/gui/src/Widgets/LCD.cpp
```diff
@@ -318,15 +318,15 @@ void LCDDisplay::digitClicked()
 const char* LCDSpinBox::__class_name = "LCDSpinBox";
 
 // used in PlayerControl
-LCDSpinBox::LCDSpinBox( QWidget *pParent, int nDigits, LCDSpinBoxType type, int nMin, int nMax )
+LCDSpinBox::LCDSpinBox( QWidget *pParent, LCDDigit::LCDType lcdType, int nDigits, LCDSpinBoxType type, int nMin, int nMax )
```
theGreatWhiteShark, line 321: How about making both the minimum and maximum values `float`s? This would allow for both non-integer and negative values. In addition, it would make the intention of disabling the limits more explicit by requiring them to be set to e.g . NaN.

theGreatWhiteShark (commented):
Awesome! Thanks a lot.

Two remarks:
- In the `Rotary` class used for most of the knobs the speed of scrolling can be reduced by a factor of 5 by pressing the [Ctrl] key while scrolling. https://github.com/hydrogen-music/hydrogen/blob/master/src/gui/src/Widgets/Rotary.cpp#L231 How about supporting this behavior in here as well to make the UX more consistent?
- We now have two sets of LCD displays. One with scrolling and one without. But by looking at them one can not tell the two apart. We could alter some color/font etc. But I think it would be better to make the `LCDDisplay`s representing the state of `Rotary` scrollable as well. After all, the LCDs used for selection can be scrolled as well.

src/gui/src/Widgets/LCD.cpp
Ratfink: Good idea.  It would probably make more sense to use positive and negative infinity than NaN though, since then there wouldn't even need to be a special case when the bounds are disabled.

Ratfink:
Good point about making the LCDs that show the value of the Rotary knobs respond to the scroll wheel as well.  I suppose that having their scrolling behavior exactly match that of the Rotary would make sense, so that scrolling over the two widgets doesn't act differently.  Really, it seems like the right approach would be to have a common base class for both widgets, and have some way for them to be connected together to share a single value.  Then in the future, we could even make a single widget that provides the + and - buttons of a spinbox, which could be connected to the LCDSpinBox in the same way.  This is starting to seem like a part of the work for #863.

I feel less sure about making the LCDSpinBox respond to the <kbd>Ctrl</kbd> key by using a smaller increment.  For the QSpinBox, at least on Linux, scrolling the wheel while holding <kbd>Ctrl</kbd> makes the value change *faster* (by 10 instead of 1), so it might be confusing for the LCDSpinBox to change slower.

src/gui/src/Widgets/LCD.cpp
theGreatWhiteShark: Even better. How about making -/+ infinity the default arguments so people could omit them if they don't require bounds?

theGreatWhiteShark:
> Good point about making the LCDs that show the value of the Rotary knobs respond to the scroll wheel as well. I suppose that having their scrolling behavior exactly match that of the Rotary would make sense, so that scrolling over the two widgets doesn't act differently. Really, it seems like the right approach would be to have a common base class for both widgets, and have some way for them to be connected together to share a single value. Then in the future, we could even make a single widget that provides the + and - buttons of a spinbox, which could be connected to the LCDSpinBox in the same way. This is starting to seem like a part of the work for #863.

+1 Indeed.

I have also thought about the coexistence of interactive and non-interactive LCD boxes a little longer. E.g. the ones displaying the peak values in the mixer would not benefit from scrolling at all. So, we are more or less bound to have these two groups present and it would be nice to be able to tell them apart.

How using different background images for the `LCDSpinBox` and the `LCDDisplay`. While `LCDSpinBox` remains blue `LCDDisplay` could have a grey / more greyish color to more clear state that this widget is a passive one.

theGreatWhiteShark:
> I feel less sure about making the LCDSpinBox respond to the Ctrl key by using a smaller increment. For the QSpinBox, at least on Linux, scrolling the wheel while holding Ctrl makes the value change faster (by 10 instead of 1), so it might be confusing for the LCDSpinBox to change slower.

Good point. I wasn't aware of this. I think in order to achieve a consistent scrolling UX more parts of the GUI have to be changed as well. So, this goes beyond the scope of this PR. I will move it to the discussions section.

theGreatWhiteShark:
P.S. how do you insert the symbols for the Ctrl keys? Is this part of the emoji set?

src/gui/src/Widgets/LCD.cpp
Ratfink: Done!

Ratfink:
>I have also thought about the coexistence of interactive and non-interactive LCD boxes a little longer. E.g. the ones displaying the peak values in the mixer would not benefit from scrolling at all. So, we are more or less bound to have these two groups present and it would be nice to be able to tell them apart.
>
>How using different background images for the LCDSpinBox and the LCDDisplay. While LCDSpinBox remains blue LCDDisplay could have a grey / more greyish color to more clear state that this widget is a passive one.

Seems like a good idea to me.  It would be pretty easy to make a new LCDDisplay font set that's gray instead of blue (the backlight is turned off 🙂), and use it for all the non-interactive LCDs.

Crazy idea: I see that there's a red LCD font set too.  What if we used that for the mute group, MIDI channel, and hihat group when they're turned off?  Would the visual distinction be useful, or just distracting?

>Good point. I wasn't aware of this. I think in order to achieve a consistent scrolling UX more parts of the GUI have to be changed as well. So, this goes beyond the scope of this PR. I will move it to the discussions section.

Sounds good.

>P.S. how do you insert the symbols for the Ctrl keys? Is this part of the emoji set?

It's the `<kbd>` HTML tag.  Writing `<kbd>Ctrl</kbd>+<kbd>z</kbd>` results in <kbd>Ctrl</kbd>+<kbd>z</kbd>.

Ratfink:
I just added a small gray font set, and went through the program so that now all the non-interactive LCDs should be gray instead of blue.  Screenshot below:

Screenshot from 2021-02-18 21-14-56

Seeing it in action, I have to say that it looks rather bland compared to the blue LCDs we currently have, but it *is* clear which LCDs are interactive and which aren't.  If we decide to move forward with this solution, there are still a few tweaks to make (the blue around the Master mixer LCD, most obviously), and I can still make the remaining LCDs in the instrument editor into LCDSpinBoxes of course.

theGreatWhiteShark:
> Crazy idea: I see that there's a red LCD font set too. What if we used that for the mute group, MIDI channel, and hihat group when they're turned off? Would the visual distinction be useful, or just distracting?

I haven't used MIDI with Hydrogen yet. Under which conditions is it turned off?

> Seeing it in action, I have to say that it looks rather bland compared to the blue LCDs we currently have, but it is clear which LCDs are interactive and which aren't.

I agree. It looks rather dull. But in principle I like it a lot! We just have to get the colors right (we could also move this one to the Discussions section to get more input). AFAICS there are no shadows in LCD fonts. So, changing color will just be a mapping all 2-3 colors to different ones in Gimp (pretty fast).

> I can still make the remaining LCDs in the instrument editor into LCDSpinBoxes of course.

That would be nice. Although the pitch one will probably be a little bit more involved since it needs to move two rotaries.

The "size" LCD in the pattern editor does not support scrolling yet too. I would suggest it could increment/decrement the nominator by 1 and thus easily enlarge or shrink the pattern. But I can implement this too. You don't have to do everything by yourself :wink: .

Ratfink:
> > Crazy idea: I see that there's a red LCD font set too. What if we used that for the mute group, MIDI channel, and hihat group when they're turned off? Would the visual distinction be useful, or just distracting?
> 
> I haven't used MIDI with Hydrogen yet. Under which conditions is it turned off?

I meant when the MIDI channel LCDSpinBox is set to "Off" instead of a channel number.  We could also consider disabling the MIDI note LCDSpinBox at the same time.  I'm really just thinking out loud here; it would probably make more sense for this to be a separate discussion/PR after this one is finished.

> > Seeing it in action, I have to say that it looks rather bland compared to the blue LCDs we currently have, but it is clear which LCDs are interactive and which aren't.
> 
> I agree. It looks rather dull. But in principle I like it a lot! We just have to get the colors right (we could also move this one to the Discussions section to get more input). AFAICS there are no shadows in LCD fonts. So, changing color will just be a mapping all 2-3 colors to different ones in Gimp (pretty fast).

That's exactly how I made the small gray style currently on this branch 😁 .  Okay, I'll try to come up with some better colors and start a discussion about it with screenshots.

> > I can still make the remaining LCDs in the instrument editor into LCDSpinBoxes of course.
> 
> That would be nice. Although the pitch one will probably be a little bit more involved since it needs to move two rotaries.

Probably not much more complicated, really.  The coarse rotary moves in integer steps, so I could just make the scroll wheel on the LCD change the coarse rotary (except at the limits, where the current behavior of the LCDSpinBox might make it need to change the fine adjustment too, e.g., scrolling up from 24.00 would presently change the value to 24.50).

theGreatWhiteShark:
> I meant when the MIDI channel LCDSpinBox is set to "Off" instead of a channel number. We could also consider disabling the MIDI note LCDSpinBox at the same time. I'm really just thinking out loud here; it would probably make more sense for this to be a separate discussion/PR after this one is finished.

Ah, I see. I thought you were going for something like if MIDI input/output is not set in the preferences the red font will indicate to  the user that the setup is not ready yet.  But I agree. This is probably best in a different PR.

> That's exactly how I made the small gray style currently on this branch grin . Okay, I'll try to come up with some better colors and start a discussion about it with screenshots.

Awesome!

Ratfink:
I've created several options, and posted them in a discussion at #1192.  Have a look, give feedback, etc.!

Ratfink:
I finally came back to work on this some more, and see #1282 seems to have made it obsolete!  We truly live in exciting times.

theGreatWhiteShark:
Hey @Ratfink ,

Your work here was actually one of the main motivations to do the large overhaul in #1282 since the UX felt so much more convenient and I thought other widgets could benefit from being streamlined and modernized as well. Thanks a lot for all the discussion and for bringing the UX forward. Sorry that your commits got superseded and sorry I didn't point this our right away. I did plan to make a note in here as soon as the other PR was done/accepted/merged. But I had to hit the break for a some time and will only start to work on it again end of this month.

BTW. The discussion about the coloring is still open. I used a slightly different one in #1218 but I'm not pleased by it yet.

Ratfink:
Oh, no problem at all!  I agree that using proper Qt widgets makes more sense in the long run, and I'm glad I could help motivate that work.  Thank you for undertaking such a big project; it really does look like it's shaping up nicely.

PR #1177, opened 2021-02-16
Use radio buttons for View > Timeline and View > Playback Track

Ratfink:
Previously, the Timeline and Playback Track items in the View menu were shown as checkboxes, but they functioned as radio buttons.  Also, while the actual state was saved in the preferences, the Timeline menu item was always shown as checked on startup.

This PR changes the menu items to be a radio button group, with an extra separator above them to set them apart from the checkboxes further.  It also correctly loads the state on program startup so that the correct item is shown as selected.


theGreatWhiteShark (approved):
Nice. Thanks a lot!

Issue #1184, opened 2021-02-20
[Bug] note out of boundary in pattern editor


oddtime:
Hydrogen version *: 010e1c05d5cb24788c7a09c81b18d0f249e7396c
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
With resolution set to "off" I can draw this note out of the boundary:
![screenshot-00f8fde3](https://user-images.githubusercontent.com/54219881/108590363-6ba71280-7363-11eb-9e3d-114aec6ab96d.jpg)

However it doesn't play

oddtime:
created PR #1193

PR #1193, opened 2021-02-24
Fix Bug #1184

oddtime:
in function `getColumn()` which returns the position (in ticks) of the nearest grid line to the position of click event,
added constraint: output >= 0 ticks


cme:
Thanks!

PR #1196, opened 2021-02-27
Experimental change of visual representation of fixed length notes

cme:
I've long been a bit dissatisfied with the way fixed-length notes are displayed in the pattern editors. The transition from round note marker to rectangle is jarring and abrupt. 

  * Suddenly going from round to having corners. 
  * Gives more visual emphasis to longer notes, which doesn't necessarily represent the aural impact.
  * Usually the *start* of the note is the most important thing, with time spent in the 'sustain' of the note less significant.

So, as a proposal and a request for comment (and not completely implemented -- only in the Drum Pattern Editor, not the Piano Roll Editor), I've tried a few things and like this representation more: a round note like the usual notes, but with a tail representing the length of the note:

image

Any thoughts or opinions?


theGreatWhiteShark:
Looks awesome!

cme:
I think I might apply this for 1.1 after all since changing the shape of the note to be closer to the usual blob makes intersection detection simpler, especially for very short notes.

trebmuh:
Great UX improvement!

PR #1204, opened 2021-03-08
Updated Spanish translation

ignotus666:


theGreatWhiteShark (approved):
Thanks a lot!

theGreatWhiteShark:
Hey @ignotus666,

I totally forgot about it: starting with the next release of Hydrogen we offer translators to be listed in the Info > About dialog as active translators with a name/handle and their email address. The idea is to allow people to hint translation issues without the daunting atmosphere when putting them on public display on the dev mailing list or a separate Github issue. But this is purely optional and there is no obligation resulting from your commitment of translating the GUI.

ignotus666:
Hi @theGreatWhiteShark,

I've been doing the Spanish translations for the last 10 years (at least) - but this is the first time I've submitted it via a PR as I only started using GitHub relatively recently. I used to just post the .ts file to the dev mailing list. 
Anyhoo, I think I'm on a translators list somewhere (Daryl Hanlon); my email address is darylo1@hotmail.com. I don't have a problem if this info is included in the About dialog.

theGreatWhiteShark:
> Anyhoo, I think I'm on a translators list somewhere (Daryl Hanlon)

Unfortunately the list of contributors was not maintained for quite some time. So, we decided to put only active developers and translators and a list of all contributors via Github since the last release (fetched in an automated way) into the dialog.

PR #1216, opened 2021-03-20
Implement basic screenshot mode.

cme:
Screenshot mode for creating widget screenshots for documentation in a scriptable manner:

  -t 'widget1 widget2 ...' to screenshot matching widgets and then quit.

The whole named widget will be grabbed, so in the cases of things like the Song Editor, that's a very large widget because the whole thing is rendered, rather than just the contents of the viewport.

As it stands, there's no way to grab transient or lazily built widgets using this method, since an intermediate action or command is needed to create the widget before it can be grabbed.

To do:
  - add a way to send signals to named widgets before grabbing so transient widgets can be created and grabbed.
  - add objectNames to widgets that we want to grab for documentation.

As an example,
```
  % ./hydrogen -t PatternEditorPanel
```
gets:

image

in `PatternEditorPanel.png`


cme:
The first commit here also encompasses a few changes to remove unneeded conditionality of updates (these also caused a bug which annoyed me by its transientness when the pattern editor would show garbage for an instant sometimes on load!)

cme:
I realise I may end up with something that's almost a mini scripting language, and have asked myself the very valid question: "Would it be better just to integrate QScript?"

I've come to the conclusion that, for the moment, the answer is 'no', because the use case is pretty straightforward and linear and integrating with QScript looks like it'll end up with just as much new code.

But it's tempting.

theGreatWhiteShark:
Nice. So, by using `setObjectName()` with elements like `QVBoxLayout` we can grab all subsets of the widgets we need too, right? (Well, apart from such monoliths as the Instrument Editor, which would benefit to be recreated using the Qt Designer anyway).

> The whole named widget will be grabbed, so in the cases of things like the Song Editor, that's a very large widget because the whole thing is rendered, rather than just the contents of the viewport.

There is an option in the Preferences using which one can adjust the maximum number of bars in a song.

cme:
> Nice. So, by using `setObjectName()` with elements like `QVBoxLayout` we can grab all subsets of the widgets we need too, right?

Yep.

I think for scrolled views it's maybe better to name and grab the containel scroll area, and perhaps pair up the grabbing process with a special purpose hydrogen.conf that defines standard window areas too.

PR #1218, opened 2021-03-21
Use a condition variable for logger sleep/wake.

cme:
Logger previously used wall clock time to sleep / wake. This had
two(ish) unpleasant side effects.

  1. logger traffic would be excessively bursty, which makes it
     harder to read the log on the console in real time, and
     particularly makes it harder to associate particular log lines
     with user actions.
  2. quitting Hydrogen had to wait for the thread to wake up and
     terminate itself before quitting. This adds on average 0.5s
     to the time taken to quit. Not annoying for most users, but
     probably fairly annoying for developers (hi).
     The sleep period was for some reason only 0.1 seconds on
     Windows, but 1.0 everywhere else. I suspect this was a typo.
     Actually there also seems to have been a pointless extra 1s
     wait (0.1s on Windows) after flushing the log. I think this
     is probably a historical artefact.

It also wasted a very tiny bit of CPU time by waking the thread up
every second to check the queue, but it would be tiny.

This change uses a condition variable instead to block the logger,
so it only wakes up when messages are present, and it does so
immediately. It also wakes up immediately to terminate and flush
buffers.

### DIFF
--- a/src/core/Logger.cpp
+++ b/src/core/Logger.cpp
@@ -27,14 +27,6 @@
 #include <QtCore/QDir>
 #include <QtCore/QString>
 
-#ifdef WIN32
-#include <windows.h>
-#define LOGGER_SLEEP Sleep( 100 )
-#else
-#include <unistd.h>
-#define LOGGER_SLEEP do { usleep( 500000 ); usleep( 500000 ); } while (0)
-#endif
-
 namespace H2Core {
 
 unsigned Logger::__bit_msk = 0;
@@ -69,7 +61,8 @@ void* loggerThread_func( void* param ) {
 	Logger::queue_t::iterator it, last;
 
 	while ( logger->__running ) {
-		LOGGER_SLEEP;
+		pthread_cond_wait( &logger->__messages_available, &logger->__mutex );
+		pthread_mutex_unlock( &logger->__mutex );
 		if( !queue->empty() ) {
 			for( it = last = queue->begin() ; it != queue->end() ; ++it ) {
 				last = it;
@@ -94,7 +87,7 @@ void* loggerThread_func( void* param ) {
 #ifdef WIN32
 	::FreeConsole();
 #endif
-	LOGGER_SLEEP;
+
 	pthread_exit( nullptr );
 	return nullptr;
 }
@@ -114,11 +107,13 @@ Logger::Logger() : __use_file( true ), __running( true ) {
 	pthread_attr_t attr;
 	pthread_attr_init( &attr );
 	pthread_mutex_init( &__mutex, nullptr );
+        pthread_cond_init( &__messages_available, nullptr );
 	pthread_create( &loggerThread, &attr, loggerThread_func, this );
 }
 
 Logger::~Logger() {
 	__running = false;
+        pthread_cond_broadcast ( &__messages_available );
 	pthread_join( loggerThread, nullptr );
 }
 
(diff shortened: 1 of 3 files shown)


theGreatWhiteShark:
Awesome. This way one could actually use the *LOG macros for debugging :D

cme:
> Awesome. This way one could actually use the *LOG macros for debugging :D

Whoah hold on, I wouldn't go *that* far :D

But, seriously, perhaps some equivalent of `fsync` for the logger might be useful, to make sure the log is flushed before doing something which may cause a crash?

theGreatWhiteShark:
> But, seriously, perhaps some equivalent of fsync for the logger might be useful, to make sure the log is flushed before doing something which may cause a crash?

Would be great. But is this possible in C++? In Go writing a `defer` statement is the first thing to do after creating a file handler. But I thought in C++ one can only raise and catch exceptions. Segfaults caused by trying to access `nullptr` or read/write memory not allocated on the other hand just have to be avoided. But I'm not a C++ native :) .

cme:
> Would be great. But is this possible in C++? In Go writing a `defer` statement is the first thing to do after creating a file handler. But I thought in C++ one can only raise and catch exceptions. Segfaults caused by trying to access `nullptr` or read/write memory not allocated on the other hand just have to be avoided. But I'm not a C++ native :) .

There's no C++ language support for it, but it's well-defined in UNIXen to raise a SEGV which can be handled with `signal`; having a signal handler which gives the logger a chance to shut down gracefully (and maybe also give a backtrace in a debug build) is totally do-able.

Windows is a whole different ballgame I guess.

theGreatWhiteShark:
> There's no C++ language support for it, but it's well-defined in UNIXen to raise a SEGV which can be handled with signal; having a signal handler which gives the logger a chance to shut down gracefully (and maybe also give a backtrace in a debug build) is totally do-able.

How about using this graceful shutdown to save the song with a different file extension, like `*.h2song.bak`, writing it's path in a dedicated member variable in `Preferences`, saving this one too, and showing during the next startup a popup window like: "Hydrogen has died unexpectedly. Do you which to restore the last song?".

cme:
> How about using this graceful shutdown to save the song with a different file extension, like `*.h2song.bak`, writing it's path in a dedicated member variable in `Preferences`, saving this one too, and showing during the next startup a popup window like: "Hydrogen has died unexpectedly. Do you which to restore the last song?".

I'm not sure about that. If we hit a SEGV or other fatal signal, it's because something, somewhere, has gone wrong, so it's impossible to know which bits of program state can be trusted or not. Flushing the `Logger` is fairly safe as there's almost no potential way for that to cause any damage, but the more things we do on such a signal, particularly with lasting (filesystem) effects, the more scope there is for permanent damage.

It might make sense to *attempt* writing out the current song, but maybe needs a bit more thought, there's a lot that could potentially go wrong in the process. Writing `Preferences` is probably a bad idea since that's state that *must* be read before starting the app. So if the auto save is successful, the normal startup should detect and offer recovery based only on the state of the autosave (and maybe somehow check it for integrity too?)

theGreatWhiteShark:
> I'm not sure about that. If we hit a SEGV or other fatal signal, it's because something, somewhere, has gone wrong, so it's impossible to know which bits of program state can be trusted or not. Flushing the Logger is fairly safe as there's almost no potential way for that to cause any damage, but the more things we do on such a signal, particularly with lasting (filesystem) effects, the more scope there is for permanent damage.

> It might make sense to attempt writing out the current song, but maybe needs a bit more thought, there's a lot that could potentially go wrong in the process. Writing Preferences is probably a bad idea since that's state that must be read before starting the app. So if the auto save is successful, the normal startup should detect and offer recovery based only on the state of the autosave (and maybe somehow check it for integrity too?)

You are right. Maybe the autosave could be triggered but in general this is not a good idea.

PR #1226, opened 2021-04-02
add fallback for drumkit lookup

### LINKED ISSUE #1222: Error messaging when displaying drumkit properties via the Soundlibrary panel
Hydrogen version *          :  Hydrogen 1.1.0-'d7d3db9a'

Hydrogen shows an error message when using the "Properties" dialog of a drumkit via the soundlibrary panel. It does not matter if a user or a system drumkit has been selected.. 

image

The dialog prints out "GmRockKit" for each time the dialog has been opened.

theGreatWhiteShark:
fixes #1222

### DIFF
--- a/src/core/Basics/Song.cpp
+++ b/src/core/Basics/Song.cpp
@@ -928,10 +928,10 @@ Song* SongReader::readSong( const QString& sFileName )
 				// and no information whether it is found at user- or
 				// system-level.
 
-				if ( ! Filesystem::drumkit_path_search( sDrumkit, Filesystem::Lookup::user ).isEmpty() ) {
+				if ( ! Filesystem::drumkit_path_search( sDrumkit, Filesystem::Lookup::user, true ).isEmpty() ) {
 					iLookup = 1;
 					WARNINGLOG( "Missing drumkitLookup: user-level determined" );
-				} else if ( ! Filesystem::drumkit_path_search( sDrumkit, Filesystem::Lookup::system ).isEmpty() ) {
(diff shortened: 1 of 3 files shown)


mauser:
Hi,

thanks for the fix, i can confirm that it solves my problem! 

There is a small issue coming up now which is related to this fix. I get now the message `(E) Filesystem::drumkit_path_search drumkit GMRockKit not found using lookup type [1]` printed one time for each instrument, so 16 times for my current drumkit. 
If you are ok with this, i would like to turn this into a message of category "WARNING" instead of error. In my opinion the caller should throw an error if Filesystem::drumkit_path_search was not successful but in that specific situtation a successful result was exptected (like you did in your fix).

theGreatWhiteShark:
> There is a small issue coming up now which is related to this fix. I get now the message (E) Filesystem::drumkit_path_search drumkit GMRockKit not found using lookup type [1] printed one time for each instrument, so 16 times for my current drumkit.
If you are ok with this, i would like to turn this into a message of category "WARNING" instead of error. In my opinion the caller should throw an error if Filesystem::drumkit_path_search was not successful but in that specific situtation a successful result was exptected (like you did in your fix).

Your right. I fixed it.

> so 16 times for my current drumkit

I occasionally think about making those log messages more concise. Especially if there are multiple versions of the same strings over and over again. During startup it not clear at first sight whether there is a problem or just the regular stuff.

mauser:
Perfect, works fine! 

Yes, that's definitely true.. Hydrogen always had the tendency to be a bit overcautious and produces error messages even if a warning might have been ok.

Issue #1190, opened 2021-02-23
Memory leakage when drag&drop inserting an instrument from the soundlibrary in the pattern editor


theGreatWhiteShark:
Some leakage has been fixed in 722877d . But according to the debug log there are still some instrument components alive.

I searched for it but couldn't find the cause of this leakage yet.

I think some time soon I will use shared pointers for the drumkits and instruments. This will mitigate some leakage but probably not this one.

theGreatWhiteShark:
Fixed in #1248

Issue #1113, opened 2021-01-22
unfreed memory loading a different drumkit?


oddtime:
Hydrogen version: 1.1.0     5089fab47aa6f26eb53dffbebe9e5644b949f9a8

Operating system + version  : Lubuntu 20 (I remember the same happening in lubuntu 16)
Audio driver + version      : 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
Way to reproduce
Run H2 => GMRockKit drumkit is loaded by default, used ram: 156.8 MB (seen from task manager).
Now load a bigger drumkit: Flac_BigMono => used ram: 340.2 MB.
Now load GMRockKit drumkit => used ram is still 340.2 MB

theGreatWhiteShark:
@oddtime does this still occur with the latest code on branch `development`?

oddtime:
Yes it is happening, but some memory is freed now (like 20-40 MB)

oddtime:
@theGreatWhiteShark, actually I have tried now the exact steps mentioned in the opening comment and they are reproduced in the same way

cme:
Is it genuinely memory being leaked or retained at the application level? Or could it be just not handed back by the runtime libraries?

mauser:
Can you start Hydrogen from commandline with the `-VDebug` argument? This shows you a list of objects which are "alive" at the time Hydrogen is shut down.

oddtime:
@cme I am sorry but I don't know, all I can see is the used ram in LXTask 0.1.9, task manager for the LXDE...

oddtime:
@mauser there are no objects shown to be alive

```
Objects map :                        class      constr   destr   alive
        [            AudioEngineInfoForm ]           1       1       0               
        [                       Director ]           1       1       0               
        [             FilesystemInfoForm ]           1       1       0               
        [              SimpleHTMLBrowser ]           1       1       0               
        [                    HydrogenApp ]           1       1       0               
        [               InstrumentEditor ]           1       1       0               
        [          InstrumentEditorPanel ]           1       1       0               
        [                   LayerPreview ]           1       1       0               
        [                    WaveDisplay ]           2       2       0               
        [                 InstrumentRack ]           1       1       0               
        [                       MainForm ]           1       1       0               
        [                          Mixer ]           1       1       0               
        [              DrumPatternEditor ]           1       1       0               
[...]

PR #1230, opened 2021-04-07
Test for memory leakage

theGreatWhiteShark:
To get rid off most of the memory leakage in Hydrogen, I decided to add some unit tests covering most occurrences. 

- [x] test all constructors and copy constructors of core/Basic classes
- [x] test all loading functions and other member functions of core/Basic classes invoking a `new` command
- ~[ ] check for other parts in core and gui which use create new core/Basic classes using `new`, encapsulate them in a dedicated routine in `CoreActionController`, and test them as well~

To ease debugging, I introduced new helper function to the `Object` class. One can now retrieve a copy of the object map and compare it to current one at some future time.

```c++
auto mapSnapshot = H2Core::Object::getObjectMap();

...

H2Core::Object::printObjectMapDiff( mapSnapshot );
```

```
 Objects map :                        class	constr   destr   alive
	[                    AudioEngine ]	     0	     0	     0
	[       AutomationPathSerializer ]	     0	     0	     0
	[                           ADSR ]	   160	   159	     1
	[                 AutomationPath ]	     0	     0	     0
	[               DrumkitComponent ]	     7	     7	     0
	[                     Instrument ]	    82	    81	     1
	[            InstrumentComponent ]	    83	    83	     0
	[                InstrumentLayer ]	   311	   311	     0
	[                 InstrumentList ]	     9	     9	     0
	[                           Note ]	     2	     2	     0
	[                        Pattern ]	     0	     0	     0
	[                    PatternList ]	     0	     0	     0
	[                       Playlist ]	     0	     0	     0
	[                         Sample ]	   306	   306	     0
	[                           Song ]	     0	     0	     0
	[                     SongReader ]	     0	     0	     0
	[           CoreActionController ]	     0	     0	     0
	[                     EventQueue ]	     0	     0	     0
	[                        Effects ]	     0	     0	     0
	[                   LadspaFXInfo ]	     0	     0	     0
	[                        XMLNode ]	   650	   650	     0
	[                         XMLDoc ]	     5	     5	     0
	[                       Hydrogen ]	     0	     0	     0
	[                 AlsaMidiDriver ]	     0	     0	     0
	[               DiskWriterDriver ]	     0	     0	     0
	[                 FakeDiskDriver ]	     0	     0	     0
	[                  TransportInfo ]	     0	     0	     0
	[                     SongWriter ]	     0	     0	     0
	[                     MidiAction ]	     0	     0	     0
	[                  ActionManager ]	     0	     0	     0
	[                        MidiMap ]	     0	     0	     0
	[                      NsmClient ]	     0	     0	     0
	[                      OscServer ]	     0	     0	     0
	[                    Preferences ]	     0	     0	     0
	[               WindowProperties ]	     0	     0	     0
	[                        UIStyle ]	     0	     0	     0
	[                     H2RGBColor ]	     0	     0	     0
	[                        Sampler ]	     0	     0	     0
	[                      SMFHeader ]	     0	     0	     0
	[                       SMFTrack ]	     0	     0	     0
	[                            SMF ]	     0	     0	     0
	[                SMFWriterSingle ]	     0	     0	     0
	[                 SMFWriterMulti ]	     0	     0	     0
	[                     SMF0Writer ]	     0	     0	     0
	[                      SMFBuffer ]	     0	     0	     0
	[          SMFTrackNameMetaEvent ]	     0	     0	     0
	[           SMFSetTempoMetaEvent ]	     0	     0	     0
	[    SMFCopyRightNoticeMetaEvent ]	     0	     0	     0
	[      SMFTimeSignatureMetaEvent ]	     0	     0	     0
	[                 SMFNoteOnEvent ]	     0	     0	     0
	[                SMFNoteOffEvent ]	     0	     0	     0
	[                          Synth ]	     0	     0	     0
	[                       Timeline ]	     0	     0	     0
Total :   1048 objects.
```

A number of tests do fail. Except of one they are disabled by commenting.

~Also, I did not add file required to test the `Legacy` load of `Drumkit`, `Pattern`, and `Playlist`.~


theGreatWhiteShark:
I'm not sure whether it's a good idea to fix all the leakage in this PR as well or to better put them in separate ones.

mauser:
It's great to have that check automated, nice work! 

> I'm not sure whether it's a good idea to fix all the leakage in this PR as well or to better put them in separate ones.

I would say it depends on the amount of leakage and the size of the required fixes :smiley: I don't have an idea if the current leak is just a missing "delete" somewhere or if its an architectural issue. Have you been looking into that issue? Otherwise i could offer to have a look at the weekend.

theGreatWhiteShark:
> I would say it depends on the amount of leakage and the size of the required fixes smiley I don't have an idea if the current leak is just a missing "delete" somewhere or if its an architectural issue. Have you been looking into that issue? Otherwise i could offer to have a look at the weekend.

Well, unfortunately it's not just leakage. Something is off in the copy constructor of `Drumkit` causing a segfault. It feels like we tackled this already two times but it doesn't work yet. Similar with the `Note` constructors. The version implemented right now either gives a leakage when not deleting the `Instrument` used for creating the Notes or a double free segfault when doing so.

Apart from this I just looked into one leakage yet: when creating a `Note` via `Note::load_from` using a `InstrumentList` which does not contain an instrument of the right ID, an empty one will be created in `Note::map_instrument` but never deleted. But when setting a flag to do so right after the creation of the empty one I got a segfault as well. Didn't looked too much into this and decided to proceed with the tests but things may be more subtle to fix here.

Also there is a wired `Song::readTempPatternList` function creating new `Pattern`s and leaking some of them. But this one is only used when pressing the **CLEAR** button at the top left of the Song Editor (it was the first time I actually used it). Here we should drop the whole function since the functionality of deleting multiple patterns while being able to restore them was also implemented by @cme in the 1.1 update of the Song Editor (probably in a much better way).

So, I would suggest I do finish the tests tomorrow and things can be fixed in different PRs. And it would be really nice if you could have a look at them.

theGreatWhiteShark:
@mauser @cme do you have by any chance some old .h2song, .h2pattern, .h2playlist and drumkits laying around to test the legacy loading code? Installing Qt4 in order to create some old files is a little bit over the top.

cme:
I don't seem to have any left I'm afraid.

theGreatWhiteShark:
AFAICS I'm done with the core part. Strangely, I couldn't reproduce #1113.

I have not added tests for any GUI function as I'm not an expert in `cmake` and didn't found time to introduce myself yet. But we could just concentrate on the core stuff for now and only add the GUI stuff if we encounter any problems.

mauser:
Hi! 

you can find some old songs and drumkits (from 2007) in our Repository. Just check out 7be7489ee50c85a8b2d0cc9e40609a96ef118117 and have a look at data/demo_songs and data/drumkits. 

I suppose i can't help with old patterns and playlist.

mauser:
> Well, unfortunately it's not just leakage. Something is off in the copy constructor of `Drumkit` causing a segfault. It feels like we tackled this already two times but it doesn't work yet. 

I had a look at this specific issue. This is sth. which is specific to the way the unit test is build. In the copy constructor of `Drumkit`,  the copy construtor of `InstrumentList` is called. The latter assumes that its operand is not a nullptr. But a drumkit is constructed by default with an not set InstrumentList. This should not happen in real live, since the method which constructs a `Drumkit` sets the InstrumentList in any case. Nontheless, i've fixed this issue by initialising the `InstrumentList` of a `Drumkit` with an empty list in its constructor. Half of the code in the `Drumkit` class did not expect that `__instruments` might be a `nullptr` :)

mauser:
You can find my fix for this here: https://github.com/hydrogen-music/hydrogen/tree/theGreatWhiteShark-phil-test-for-memory-leakage

mauser:
> Similar with the Note constructors. The version implemented right now either gives a leakage when not deleting the Instrument used for creating the Notes or a double free segfault when doing so.

Wow, that one got me pretty confused :exploding_head: During debugging, it seemed that Instrument::copy_adsr() returned an adress which pointed to the same memory location as the thing it should copy. So i inserted an assert(): 

`	ADSR* pTmpADSR = new ADSR( __adsr );
	assert( pTmpADSR != __adsr);`

And in your instance, the assert code evaluated to true. That took a moment to figure out, but it also boils down to an issue with the unit test. The thing is that `__adsr` had already been deleted, because it was used in an earlier part of the test function (for a different Instrument). The memory was released, but the pointer pointed still to the old location and was then re-used to construct an Instrument for the Note test. So it is valid that the next new() returns the same memory address. I've added some brackets around each constructor-part of the test to introduce anonymous scopes, this should help to minimize those problems.

mauser:
>Also there is a wired Song::readTempPatternList function creating new Patterns and leaking some of them. But this one is only >used when pressing the CLEAR button at the top left of the Song Editor (it was the first time I actually used it). Here we should >drop the whole function since the functionality of deleting multiple patterns while being able to restore them was also implemented >by @cme in the 1.1 update of the Song Editor (probably in a much better way).

@cme: Colin, could you adress this issue?

theGreatWhiteShark:
> you can find some old songs and drumkits (from 2007) in our Repository

Yeah, you are right. There is even a dedicated repo https://github.com/hydrogen-music/Song-and-pattern-repository/ containing old patterns. :smile:

theGreatWhiteShark:
> I had a look at this specific issue. This is sth. which is specific to the way the unit test is build. In the copy constructor of Drumkit, the copy construtor of InstrumentList is called. The latter assumes that its operand is not a nullptr. But a drumkit is constructed by default without an InstrumentList (__instruments = nullptr).

Ah, I see. There are probably some other instances where  the state of the involved objects is crucial. I'll add  test for the copy constructor of `Drumkit` with a non-empty `InstrumentList` too.

cme:
> Also there is a wired `Song::readTempPatternList` function creating new `Pattern`s and leaking some of them. But this one is only used when pressing the **CLEAR** button at the top left of the Song Editor (it was the first time I actually used it). Here we should drop the whole function since the functionality of deleting multiple patterns while being able to restore them was also implemented by @cme in the 1.1 update of the Song Editor (probably in a much better way).

It's also used in undoing deletion and loading (overwriting) patterns using the pattern list widget -- the whole sequence state is snapshotted to a temp file and restored on undo. 

This can probably be replaced fairly easily with existing undo actions and cleaned up a lot. Excellent!

theGreatWhiteShark:
@cme @mauser I'm trying to add some tests for the GUI methods creating and destroying objects of the basis class. But I'm not getting past `cmake`. My current attempt is 

``` diff
$ git diff src/tests/CMakeLists.txt 
diff --git a/src/tests/CMakeLists.txt b/src/tests/CMakeLists.txt
index 6d319314..138f63cc 100644
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -1,9 +1,9 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
 add_definitions()
 include_directories(
-    ${CMAKE_SOURCE_DIR}/src/core/include            # core headers
     ${CMAKE_SOURCE_DIR}/src
     ${CMAKE_BINARY_DIR}/src                         # generated config.h
+    ${CMAKE_BINARY_DIR}/src/gui/hydrogen_autogen/include/                     # ui_ headers
     ${QT_INCLUDES}                                  # TODO be able to remove this
     ${JACK_INCLUDE_DIRS}
     ${CPPUNIT_INCLUDE_DIRS}
@@ -20,8 +20,8 @@ SET_PROPERTY(TARGET tests PROPERTY CXX_STANDARD 17)
 target_link_libraries(tests
        hydrogen-core-${VERSION}
        ${CPPUNIT_LIBRARIES}
-       Qt5::Core
        Qt5::Test
+       Qt5::Widgets
 )
 
-add_dependencies(tests hydrogen-core-${VERSION})
+add_dependencies(tests hydrogen hydrogen-core-${VERSION})
diff --git a/src/tests/MemoryLeakageTest.cpp b/src/tests/MemoryLeakageTest.cpp
index 1e9bbb38..b8f6fa9f 100644
--- a/src/tests/MemoryLeakageTest.cpp
+++ b/src/tests/MemoryLeakageTest.cpp
@@ -21,6 +21,9 @@
 #include <core/Sampler/Sampler.cpp>
 #include <core/Helpers/Xml.h>
 
+#include <gui/src/HydrogenApp.h>
+#include <gui/src/SongEditor/SongEditor.h>
+
 #include "TestHelper.h"
 
 CPPUNIT_TEST_SUITE_REGISTRATION( MemoryLeakageTest );
@@ -411,6 +414,10 @@ CPPUNIT_ASSERT( pPlaylist != nullptr );
        
 }
 
+void MemoryLeakageTest::testGui() {
+       auto pApp = HydrogenApp::get_instance();
+}
+
 void MemoryLeakageTest::tearDown() {
        if ( H2Core::Filesystem::drumkit_exists( "testKitLadida" ) ) {
                QString sPath = H2Core::Filesystem::drumkit_usr_path( "testKitLadida" );
diff --git a/src/tests/MemoryLeakageTest.h b/src/tests/MemoryLeakageTest.h
index 8e6a0f2e..96e6bfde 100644
diff --git a/src/tests/MemoryLeakageTest.h b/src/tests/MemoryLeakageTest.h
index 8e6a0f2e..96e6bfde 100644
--- a/src/tests/MemoryLeakageTest.h
+++ b/src/tests/MemoryLeakageTest.h
@@ -7,6 +7,7 @@ class MemoryLeakageTest : public CppUnit::TestCase {
        CPPUNIT_TEST_SUITE(MemoryLeakageTest);
        CPPUNIT_TEST(testConstructors);
        CPPUNIT_TEST(testLoading);
+       CPPUNIT_TEST(testGui);
        CPPUNIT_TEST_SUITE_END();
 
 public:
@@ -23,6 +24,9 @@ public:
         */
        void testLoading();
 
+       /** Calls all routines of the GUI creating and destroying basic
+               objects on the heap.*/
+       void testGui();
 };

```

But it gives me 

> /usr/bin/ld: CMakeFiles/tests.dir/MemoryLeakageTest.cpp.o: in function `MemoryLeakageTest::testGui()':
/home/phil/git/hydrogen/src/tests/MemoryLeakageTest.cpp:418: undefined reference to `HydrogenApp::get_instance()'
collect2: error: ld returned 1 exit status
make[2]: *** [src/tests/CMakeFiles/tests.dir/build.make:449: src/tests/tests] Error 1
make[1]: *** [CMakeFiles/Makefile2:315: src/tests/CMakeFiles/tests.dir/all] Error 2
make: *** [Makefile:152: all] Error 2

I also tried using the `add_library` function in `src/gui/CMakeLists.txt` to create a custom object I can link using `target_link_libraries` in `src/tests/CMakeLists.txt` but it yielded the same result.

Could you point me into a direction how to fix this?

mauser:
Hi! 

>Could you point me into a direction how to fix this?

Currently you're only linking against libhydrogen-core-1.1.0.so (via `target_link_libraries`). This is build from the contents of the src/core folder, so it does not contain the symbols of HydrogenUI. Furthermore, our current cmake setup does not build a library for the UI-part, since that part contains the main() function. 

You could add all gui files (minus main.cpp) from src/gui to your compilation unit.. you would also need to handle all the include/link dependencies for those file (have a look at the CMakeLists.txt in the gui folder). This is definitely not a clean solution :-/ We could also introduce a new library which includes all ui parts apart from main.cpp, but this would add a bit more complexity and overhead to Hydrogen's build system, and i suppose that the binaries would be larger. Maybe someone else has a better idea?

theGreatWhiteShark:
> You could add all gui files (minus main.cpp) from src/gui to your compilation unit.. you would also need to handle all the include/link dependencies for those file (have a look at the CMakeLists.txt in the gui folder).

Ah, I see. So it was the `main.cpp` file hindering me from doing so.

> We could also introduce a new library which includes all ui parts apart from main.cpp, but this would add a bit more complexity and overhead to Hydrogen's build system, and i suppose that the binaries would be larger.

Hmm. At least it feels like something that should be done in a separate PR. I think I will have a look at the remaining failing tests and deal with the GUI tests later (Maybe the loading ones will already solve https://github.com/hydrogen-music/hydrogen/issues/1190).

theGreatWhiteShark:
Alright. There are just two failing ones left.
1. the `Song::readTempPatternList`
2. and one affecting the loading of a `Note` from file.

The 1. we already discussed. 

The 2. is a sneaky one.  When loading a `Note` using `Note::load_from( XMLNode* node, InstrumentList* instruments )` and providing an `InstrumentList` which does not contain an `Instrument` with the ID found in `node` `Note::map_instrument()` will create a new `Instrument` on the heap and assigns it. The destructor of `Note` does not delete the `__instrument` member since those usually outlive the `Note`s. And it should neither do so when a fallback `Instrument` was created in `Note::map_instrument()` since another `Note` could be constructed from the first one via the copy constructor and carry the same fallback `Instrument`. So, right now, the fallback `Instrument` is deemed to live on. AFAIKS this can be fixed by using shared pointer instead of raw pointers for `Song::__instrument`. But this I will do in a separate PR.

Issue #1240, opened 2021-04-18
[Bug] Velocity Automation works only for patterns with length = 192


oddtime:
Way to reproduce: write a velocity automation ramp with value from 0 to 1 during one pattern of length = 8/8 (192*2 ticks), previously filled with notes.

I hear like if there were TWO ramps from 0 to 1

I think that the problem is here in `Hydrogen.cpp`:
```
		float velocity_adjustment = 1.0f;
		if ( pSong->getMode() == Song::SONG_MODE ) {
			float fPos = m_nSongPos + (pNote->get_position()%192) / 192.f;
			velocity_adjustment = vp->get_value(fPos);
		}
```

where that modulo `%192 `resets the positions periodically, instead that should depend on the biggest pattern size at that moment in the sequence.
If somebody confirms the bug I can work on it.

oddtime:
I have opened pull req #1241

theGreatWhiteShark:
Closed with #1241

PR #1241, opened 2021-04-18
fix bug: make velocity automation valid for any pattern length

oddtime:
See #1240

### DIFF
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -946,8 +946,7 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 		framepos = pHydrogen->getRealtimeFrames();
 	}
 	
-	AutomationPath *vp = pSong->getVelocityAutomationPath();
-	
+	AutomationPath *pVelAutomationPath = pSong->getVelocityAutomationPath();
 
 	int nSongLength = 0;
 	if ( pSong->getMode() == Song::SONG_MODE ) {
@@ -957,24 +956,6 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 	// reading from m_songNoteQueue
 	while ( !m_songNoteQueue.empty() ) {
 		Note *pNote = m_songNoteQueue.top();
-
-		float velocity_adjustment = 1.0f;
-		
-		// Velocity Adjustment Automation
-		if ( pSong->getMode() == Song::SONG_MODE ) {
-			// we need to calculate the total length (in ticks) of pattern columns BEFORE the playing column
-			int nPreviousColumnsTotalLength;
-			// findPatternInTick() writes the needed value into nPreviousColumnsTotalLength
-			findPatternInTick( m_nPatternStartTick, false, &nPreviousColumnsTotalLength );
-		
-			//get position in the pattern columns scale (refers to the pattern sequence which can be non-linear with time)
-			float fPos = m_nSongPos // this is the integer part
-						+ ( pNote->get_position() // this note_position is monotonic with time (never reset in loop mode )
-							- nSongLength * /*floor*/( pNote->get_position() / nSongLength ) // use floor if ticks become float
-							- nPreviousColumnsTotalLength
-						   ) / (float) pHydrogen->getCurrentPatternList()->longest_pattern_length(); // divide to get fractional part (<1)
-			velocity_adjustment = vp->get_value(fPos);
-		}
 
 		// verifico se la nota rientra in questo ciclo
 		unsigned int noteStartInFrames =
@@ -994,11 +975,17 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 		bool isOldNote = noteStartInFrames < framepos;
 
 		if ( isNoteStart || isOldNote ) {
-			// Humanize - Velocity parameter
-			pNote->set_velocity( pNote->get_velocity() * velocity_adjustment );
+			// Velocity Automation Adjustment
+			if ( pSong->getMode() == Song::SONG_MODE ) {
+				// position in the pattern columns scale (refers to the pattern sequence which can be non-linear with time)
+				float fPos = static_cast<float>( m_nSongPos ) // this is the integer part
+							+ ( static_cast<float>( pNote->get_position() % nSongLength - m_nPatternStartTick )
+								/ static_cast<float>( pHydrogen->getCurrentPatternList()->longest_pattern_length() ) );
+				pNote->set_velocity( pNote->get_velocity() * pVelAutomationPath->get_value( fPos ) );
+			}
 			
-			/* Check if the current note has probability != 1
-			 * If yes remove call random function to dequeue or not the note
+			/* Check if the current note has probability != 1.
+			 * If yes call a random function to choose whether to dequeue the note or not
 			 */
 			float fNoteProbability = pNote->get_probability();
 			if ( fNoteProbability != 1. ) {
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -962,30 +962,16 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
 		
 		// Velocity Adjustment Automation
 		if ( pSong->getMode() == Song::SONG_MODE ) {
-			// calculate the total length (in ticks) of pattern columns BEFORE the playing column
-
-			/////////// the following lines are inspired to findPatternInTick()
-			int nPreviousColumnsTotalTicks = 0;
-			std::vector<PatternList*> *pPatternColumns = pSong->getPatternGroupVector();
-			int nColumns = pPatternColumns->size();
-
-			// Sum the lengths of all pattern columns and use the macro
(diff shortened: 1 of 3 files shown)


theGreatWhiteShark (commented):
how about using a call to  `findPatternInTick()`  and a new `Song::lengthInTicks()` function instead of the inspired code regions?

The latter would be basically the first added section and could prove helpful at other places in the code too (IIRC I did a workaround for such a thing at one point too).

oddtime:
Hi, thanks for the clue! I wrote the comment to highlight the possible factorization indeed

oddtime:
In `findPatternInTick()` there is an additional statement inside the loop to return, so `Song::lengthInTicks()` can't be used there. Do you remember where you used the workaround?

oddtime:
> how about using a call to findPatternInTick()

By calling we would get the value that is already stored in `m_nSongPos`, instead the second "inspired region" is to calculate the length (in ticks) of pattern columns before the current time position

theGreatWhiteShark:
> In findPatternInTick() there is an additional statement inside the loop to return, so Song::lengthInTicks() can't be used there. Do you remember where you used the workaround?

I can't remember (not even if it was in a code that got merged eventually). Anywatm the way you implemented it is just as I had in mind :+1: 

> By calling we would get the value that is already stored in m_nSongPos, instead the second "inspired region" is to calculate the length (in ticks) of pattern columns before the current time position

But wouldn't `findPatternInTick( m_nPatternStartTick - 1, false, &nPreviousColumnsTotalTicks )` be exactly what your code is doing?

oddtime:
> But wouldn't findPatternInTick( m_nPatternStartTick - 1, false, &nPreviousColumnsTotalTicks ) be exactly what your code is doing?

You are right! (but I think without `- 1`)
 I used a `int tmp` variable for the return value of `findPatternInTick()`, even if it should be safe to write the return to `m_nSongPos`. What do you think? maybe in this way it is more evident that we are interested in the value written via the pointer, and `m_nSongPos` is already up-date.

I see now that there is another similar function:
` long Hydrogen::getTickForPosition( int pos )`

oddtime:
maybe `nPreviousColumnsLength = pHydrogen->getTickForPosition( m_nSongPos );` makes it simpler?

theGreatWhiteShark:
> You are right! (but I think without - 1)
I used a int tmp variable for the return value of findPatternInTick(), even if it should be safe to write the return to m_nSongPos. What do you think? maybe in this way it is more evident that we are interested in the value written via the pointer, and m_nSongPos is already up-date.

How about omitting the `int tmp = ` as you are not using this variable and just write.

```c++
findPatternInTick( m_nPatternStartTick, false, &nPreviousColumnsLength );
```

> I see now that there is another similar function:
long Hydrogen::getTickForPosition( int pos )

Yes. `findPatternInTick` is a global function defined in the `Hydrogen.cpp` file (on `master`. In `development` it now lives in `AudioEngine.cpp` as a member function thanks to the refactoring of this mess by @mauser ). I introduced `getTickForPosition()` to make it accessible to other classes as well. But since `find*` is now a proper member maybe the latter one can be removed again because it just adds redundancy (but not here but at some point in the future). Long story short: `find*` is the better one to use :smiley: 

> I don't know if this can be merged directly into development, I remember that many lines moved to AudioEngine

Don't worry about it. I intend to merge the changes in `master` little by little via the PRs I open against `development`. This way we can avoid a huge mess by doing just a single merge after the release.

src/core/Hydrogen.cpp
```diff
@@ -945,17 +945,34 @@ inline void audioEngine_process_playNotes( unsigned long nframes )
...
 
 		float velocity_adjustment = 1.0f;
+		
+		// Velocity Adjustment Automation
 		if ( pSong->getMode() == Song::SONG_MODE ) {
-			float fPos = m_nSongPos + (pNote->get_position()%192) / 192.f;
+			// we need to calculate the total length (in ticks) of pattern columns BEFORE the playing column
+			int nPreviousColumnsLength;
+			 // findPatternInTick() writes the needed value into nPreviousColumnsLength
+			int tmp = findPatternInTick( m_nPatternStartTick, false, &nPreviousColumnsLength );
```
theGreatWhiteShark, line 968: This line does not seem to depend on the `Note`s  the while loop is iterating over. How about moving it in front of the loop?

oddtime: Are you sure  that notes can't be in different pattern columns?

oddtime:
> How about omitting the int tmp = as you are not using this variable and just write...

Ok. I didn't know that you can call a function like if it was `void`

src/core/Hydrogen.cpp
oddtime: Ah ok, the function is called for every tick, (I thought it was for the audio buffer)

oddtime: I set some debug `printf()`  and I see that:
` audioEngine_process_playNotes() ` is called every nframes. Sometimes the notes in the loop belong to different pattern column.
So if I am not wrong now, I was wrong before when I used `m_nSongPos` trusting in its value. Instead that value has to be calculated as well in the loop

theGreatWhiteShark: > I set some debug printf() and I see that:
audioEngine_process_playNotes() is called every nframes. Sometimes the notes in the loop belong to different pattern columns.

Yes. The `audioEngine_process()` function is registered in the audio driver which in turns calls it on every `nframes` (buffer size) of audio that needs to be processed. After - and among other things - this function calls `audioEngine_updateNoteQueue`  (which sets `m_nSongPos` and `m_nPatternStartTick`) it calls `audioEngine_process_playNotes()`. These calls for handling the audio do live in a single thread and are called sequentially. Only by interaction with the GUI, by MIDI or OSC - which all live in different threads - other core routines can be run outside this sequential order. So, whether or not the Note belongs to a different column those two variables will be the same for all Notes in the note queue.

> Sometimes the notes in the loop belong to different pattern columns.

I haven't looked into this part too closely. But in general you can have Instruments with very long samples. A Note associated with them might play for several (short) patterns - e.g. a gong or a ride.

oddtime: > So, whether or not the Note belongs to a different column those two variables will be the same for all Notes in the note queue

So do you confirm that we cannot use `m_nSongPos` to calculate `velocity_adjustment` because `m_nSongPos` stores the starting position of the buffer and the buffer size may extend to the next pattern column in the song?

theGreatWhiteShark: > So do you confirm that we cannot use m_nSongPos to calculate velocity_adjustment because m_nSongPos stores the starting position of the buffer and the buffer size may extend to the next pattern column in the song?

I took a deeper look at the code. This should work:

```c++
	AutomationPath *vp = pSong->getVelocityAutomationPath();
	
	int nSongLength = 0;
	if ( pSong->getMode() == Song::SONG_MODE ) {
		nSongLength = pSong->lengthInTicks();
	}

	// reading from m_songNoteQueue
	while ( !m_songNoteQueue.empty() ) {
		Note *pNote = m_songNoteQueue.top();
		float velocity_adjustment = 1.0f;
		
		// Velocity Adjustment Automation
		if ( pSong->getMode() == Song::SONG_MODE ) {
			float fPos = static_cast<float>( m_nSongPos )
				+ ( static_cast<float>( pNote->get_position() % nSongLength - m_nPatternStartTick )
					/ static_cast<float>(pHydrogen->getCurrentPatternList()->longest_pattern_length()) );
			velocity_adjustment = vp->get_value(fPos);
		}

```

`m_nSongPos` is always the current pattern number and the second `findPatternInTick()` call can be omitted since `m_nPatternStartTick` and `nPreviousColumnsLength` will always be the same.

oddtime: Ok I tested and it works indeed. So some notes of the queue don't start in the current tick.

>`+ ( static_cast<float>( pNote->get_position() % nSongLength - m_nPatternStartTick )
`

I wanted to use a formula in the way that it could directly work for `double` positions and length ( without `%` )... but of course until it gets merged the result is the same.

Since the velocity automation is used only `if ( isNoteStart || isOldNote )` (the other notes in the queue are skipped if I understand now), can we put the velocity adjustment calculation into that scope?

theGreatWhiteShark: > I wanted to use a formula in the way that it could directly work for double positions and length ( without % )... but of course until it gets merged the result is the same.

Okay. But the equation should work for `double` positions too. The only thing that has to be adapted is to change the casts into `static_cast<double>` and the modulo operator `%` into the [std::fmod](https://en.cppreference.com/w/cpp/numeric/math/fmod) function

> Since the velocity automation is used only if ( isNoteStart || isOldNote ) (the other notes in the queue are skipped if I understand now), can we put the velocity adjustment calculation into that scope?

Sure

theGreatWhiteShark (approved):
Alright, everything's fine now. Thanks again!

oddtime:
Thanks to you!

PR #1248, opened 2021-04-23
Use shared pointer for Instrument, InstrumentLayer, InstrumentComponent, ADSR

### LINKED ISSUE #1190: Memory leakage when drag&drop inserting an instrument from the soundlibrary in the pattern editor

theGreatWhiteShark:
Fixes #1190


cme (commented):
Just a thought but... could we just remove the Instrument pointer from notes altogether and get the pointer from the instrument_id and the Song?

theGreatWhiteShark:
> Just a thought but... could we just remove the Instrument pointer from notes altogether and get the pointer from the instrument_id and the Song?

In principle this should be no problem. But maybe there are some strange side effects in the elderly and naturally grown parts of the core. Bjarne Stroustrup made me believe replacing pointers by shared pointers has almost no disadvantages and so I tended to the above version of the fix. But you are right, if there are unused Notes lurking around shared pointer will be of not much help.

PR #1251, opened 2021-04-27
Float note positions and tuplets

oddtime:
After some tests I am pushing the experimental work (in progress) with `double` note tick positions.
I have not implemented a class for note positions, just using the 192/whole note resolution as a unit.

It includes the minimal gui used to test tuplets in #1127 that can be used to input `double` position notes.
Note: reloading a file using in the other branch versions, transforms all fractional positions to zero, because of `QLocale::toInt()` in `readXMLInt()`

Note: piano roll is not working for tuplets


oddtime:
Along with many things to discuss, there is the printing precision of `double` type values in XML nodes, affecting both Song export and copy/cut & paste. This is bound together with the tolerance that the GUI accepts while searching the note (to overwrite or remove the note for example).

For the moment I am using 10 significant digits for the XML `double `nodes, and defined a `POS_EPSILON` in `Pattern` for the tolerance (wider than the XML precision).

oddtime:
I see Windows is not building the last commit. I have no idea, there were very few changes (maybe function `fabs` ?).

theGreatWhiteShark:
> I see Windows is not building the last commit. I have no idea, there were very few changes (maybe function fabs or missing a
parenthesis ?).

Hey @oddtime ,

I think this was due to one of my commits which slipped into master and caused the Windows builds to break.

oddtime:
Hi @theGreatWhiteShark
this seems good ;) thanks for it
and thanks for the information here, send the link when you upload the thing.
I am afraid that I will have to rewrite all this PR at the end...

theGreatWhiteShark:
> I am afraid that I will have to rewrite all this PR at the end...

Well, at least the AudioEngine stuff I'm doing right now is more or less orthogonal to your PR. But there was indeed quite some progress during the last year.

oddtime:
@theGreatWhiteShark is your patch of Hydrogen's AudioEngine fully merged now? Thanks

theGreatWhiteShark:
> @theGreatWhiteShark is your patch of Hydrogen's AudioEngine fully merged now? Thanks

Hmm. So, the audio engine already uses double precision ticks internally. That's merged. But there are still a number of glitches due to me getting some things not right the first time and some conceptional legacy still prevailing. I'm in the process of fixing these things and make the engine (hopefully) run as smooth as it gets. It still needs some more testing but it feels like its almost done. https://github.com/theGreatWhiteShark/hydrogen/tree/phil-introduce-transport-position

That aside, `Note` placement is still only possible on integer tick positions. I did not touch this in order to not change too many things at the same time. From the perspective of the audio engine, however, this should take only some minor changes to support double precision note placement. But the remainder of the core and the GUI has to be adapted. This is probably quite some work.

PR #1252, opened 2021-04-27
Make the swing humanization code more explicit

oddtime:
Basically this makes clear that the Swing is applied on upbeat 16th notes, rather than 8th notes, both in the GUI and in the code for developers to maintain it.
And now the swing offset is proportional to MAX_RESOLUTION.


oddtime:
there was a TODO in those lines
```
 // TODO: incorporate the factor of 6.0 either
						// in Song::__swing_factor or make it a member
						// variable.
```
I slightly modified it to keep trace, but I would be fine to remove it

theGreatWhiteShark (approved):

Issue #1244, opened 2021-04-19
Manual can not be displayed via Main Menu > Info > Manual


theGreatWhiteShark:
Hydrogen version *          : d16cbf5
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Right now, Hydrogen will try to access a general `manual.html` - which does not exists.

https://github.com/hydrogen-music/hydrogen/blob/d16cbf5a28c486b84aa613a9572475ab18a8f95a/src/gui/src/HydrogenApp.cpp#L286

In addition, the CSS files is not loaded and the output does not look pleasing.

theGreatWhiteShark:
I fixed the display of the manual itself but for some reason it won't load the provided stylesheet. 

https://github.com/theGreatWhiteShark/hydrogen/tree/phil-view-manual

Maybe it's worth migrating this widget to a more advanced/recent Qt class like the webview? (although I somewhat dislike the concept). I'll have a further look at this tomorrow.

theGreatWhiteShark:
`Qt5` seems to only understand [HTML4](https://doc.qt.io/qt-5/richtext-html-subset.html). I did not found anything regarding the CSS capabilities but I highly doubt the `QTextDocument` and the underlying engine can interpret and apply the new CSS3 styling.

@mauser @cme How about we remove the `HelpBrowser` class and instead use [`QDesktopService`](https://doc.qt.io/archives/qt-4.8/qdesktopservices.html#openUrl) to open the manual in the default browser of the user?

cme:
Fix on master

PR #1260, opened 2021-05-03
Open manual in browser

cme:
Fix for #1244  -- opens in default system browser as per @theGreatWhiteShark 's suggestion.

Also looks for an appropriate translation for the current locale preferences, if available.

For broken installs, will fall back to the online manual.


theGreatWhiteShark (approved):
Nice!

PR #1268, opened 2021-05-09
Prefer CoreAudio driver when available in the 'Auto' order.

cme:
Also, small refactor to remove duplication, and also use the "Auto"
list as a fallback when a user's preferred driver isn't available.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,8 @@
 	* Instrument main pitch shift offset
 	* Custom pattern size support with representation in note values
 	* Custom pan law support in mixer
+	* Attempt to find fallback Audio drivers if user preferred
+	  driver fails
 
 2021-04-11 the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 	* Release 1.0.2
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,7 @@
 	* Custom pattern size support with representation in note values
 	* Custom pan law support in mixer
 	* Attempt to find fallback Audio drivers if user preferred
-	  driver fails
+	  driver fails, as if the "Auto" driver option were selected
 
 2021-04-11 the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 	* Release 1.0.2
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -2120,60 +2120,33 @@ void audioEngine_startAudioDrivers()
 
 
 	QString sAudioDriver = preferencesMng->m_sAudioDriver;
-	if ( sAudioDriver == "Auto" ) {
-	#ifndef WIN32
-		if ( ( m_pAudioDriver = createDriver( "JACK" ) ) == nullptr ) {
-			if ( ( m_pAudioDriver = createDriver( "ALSA" ) ) == nullptr ) {
-				if ( ( m_pAudioDriver = createDriver( "CoreAudio" ) ) == nullptr ) {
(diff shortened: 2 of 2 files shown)


theGreatWhiteShark:
Thanks! Could you make a note in the ChangeLog as well? This might alter the local setup of some users.

ChangeLog
```diff
@@ -11,6 +11,8 @@
 	* Instrument main pitch shift offset
 	* Custom pattern size support with representation in note values
 	* Custom pan law support in mixer
+	* Attempt to find fallback Audio drivers if user preferred
+	  driver fails
```
theGreatWhiteShark, line 15: How about adding "If the "Auto" audio driver is selected, ..."?

src/core/Hydrogen.cpp
```diff
@@ -2120,60 +2120,33 @@ void audioEngine_startAudioDrivers()
...
-							}
-						}
-					}
-				}
+	QStringList drivers = { "CoreAudio", "JACK", "ALSA", "OSS", "PulseAudio" };
+
+#ifdef WIN32
+	drivers.removeAll( "PortAudio" );
+	drivers.push_front( "PortAudio" );
+#endif
```
theGreatWhiteShark, line 2128: I like the way you handle it with lists.

I wounder if it would be useful to make the choices a little more explicit by introducing another `#ifdef` for macOS and removing the drivers from the list, which are not supported by the OS anyway. At first glance it looks like Linux will use the CoreAudio driver or the second most common choice on Windows is ALSA. But on the other hand, with a little audio context these things are also somewhat evident.

theGreatWhiteShark (commented):
Nice. This was indeed one of the parts that badly needed some refactoring

src/core/Hydrogen.cpp
cme: Thinking about it more, I think it's probably best overall for legibility and to make the intention clear to just provide the whole list for each platform separately. That should avoid confusion like this, makes the order easy to understand for each platform, and I think is probably actually fewer lines of code:
```c++
#if defined(WIN32)
    QStringList drivers = { "PortAudio", "JACK" };
#else if defined(__APPLE__)
    QStringList drivers = { "CoreAudio", "JACK", "PulseAudio" };
#else /* Linux */
    QStringList drivers = { "JACK", "ALSA", "OSS", "PulseAudio", "PortAudio" };
#endif
```

src/core/Hydrogen.cpp
```diff
@@ -2120,60 +2120,36 @@ void audioEngine_startAudioDrivers()
...
-								// use the NULL output driver
-								m_pAudioDriver = new NullDriver( audioEngine_process );
-								m_pAudioDriver->init( 0 );
-							}
-						}
-					}
-				}
+#if defined(WIN32)
+    QStringList drivers = { "PortAudio", "JACK" };
+#elif defined(__APPLE__)
```
theGreatWhiteShark, line 2125: In the remainder of the code the `Q_OS_MACX` macro is used. Are they equivalent?

theGreatWhiteShark: > I hadn't noticed that before, actually. They're almost equivalent; __APPLE__ is true for iOS as well, so if there were ever an iOS port this would be applicable here as well, and is also the direct equivalent of the WIN32 symbol used for the Windows-specific parts. But for all that it'll ever matter, Q_OS_MACX would be fine too, though it raises the question of whether we should also be using Q_OS_WINDOWS instead of WIN32.

Hmm. I don't have a strong opinion on this one. It's nice to have it consistent but that's all. If it's doing the same, it's fine.

theGreatWhiteShark (approved):

PR #1271, opened 2021-05-10
Appearance tab overhaul

theGreatWhiteShark:
appearance_tab _reworked
(note the button font and the ones in the `SoundLibrary`)

## visual/user-level changes
* Font and coloring updates take effect immediately. *With the `PreferencesDialog` still open!* **No restart required anymore!!** (except when changing the GUI layout)
* I dropped the previous font options in favor for three different levels of font (without exposing their point sizes). These are used for 1. titles, menues, the grids, ... , 2. pattern and instrument names in the `SongEditor` and `PatternEditor`, the song names in the `PlaylistDialog`, and the listed items in the `SoundLibrary`, and 3. the text on custom widgets (for which we currently only support the large buttons. But the label of the rotary will follow at some point).
* I introduced the ability to decrease or increase the overall font size (using hand crafted values) in order to increase accessibility.
* I dropped the "fixed" coloring methods and rename "steps" to "Custom".
* Via a colored button the particular line color can now be adjusted using a `QColorDialog`. This causes for one color selection button to be present per default which should nicely point the user towards this features (I wasn't aware of it for years).
* The font of the tooltips of all `Button` and derived classes does now match the one in the other tooltips.
* The restart warning (for changes to take effect) is now only shown if the "Default interface layout" is changed.

## internal changes
* I removed the `H2RGBColor` and replaced it by `QColor`.
* The update is done using the signal `preferencesChanged(bool)`. The `bool` is never used yet but is intended to indicate whether the Ok button of the `PreferencesDialog` was clicked and all prefs should be updated or if there was a change in the Appearance tab

## limitations
* I didn't implement the update of dialogs that are created upon opening. E.g. when the `SampleEditor` or a file open dialog is opened, font changes will only take action upon closing and reopening.
* Only up to 50 different colors are supported.

## Further plans
* I intend to put all the places the GUI accesses the `Preferences` class into either their constructor or an `onPreferencesChanged` public slot and store the required values in private members. This way it will be much more clear what happens when compared to picking up the changes based on a timer. Also, the code will be more consistent with changes introduced in the PR.


src/gui/src/InstrumentEditor/InstrumentEditor.cpp
```diff
@@ -1290,6 +1295,61 @@ void InstrumentEditor::update()
 	}
 }
 
+int InstrumentEditor::getPointSize() const {
```
cme, line 1298: There seem to be quite a few instances of `getPointSize` and `on_PreferencesChanged` with almost identical code. Would it be worth considering adding some common superclass that encapsulates that functionality? Like
```c++

template < int s, int n, int l >
class WidgetThatHasAFont {
protected:
  int getPointSize() {
    switch Preferences::FontSize::Small:  return s;
  [ ... etc ]
};

class InstrumentEditor : public QWidget, protected WidgetThatHasAFont< 10, 12, 14 > {
...
};
```
?

cme (commented):
I *love* this!

I've had a quick look through about half of it and have a couple of comments, and have found a potential crash too.

Awesome stuff! :)

src/core/LocalFileMgr.cpp
```diff
@@ -124,6 +124,27 @@ QString LocalFileMng::readXmlString( QDomNode node , const QString& nodeName, co
 	}
 }
 
+QColor LocalFileMng::readXmlColor( QDomNode node , const QString& nodeName, const QColor& defaultValue, bool bCanBeEmpty, bool bShouldExists, bool tinyXmlCompatMode)
+{
+	QString text = processNode( node, nodeName, bCanBeEmpty, bShouldExists );
+	if ( text == nullptr ) {
```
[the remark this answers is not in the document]
theGreatWhiteShark: Oh, I haven't even noticed. I bluntly copy-pasted the code of `readXmlString` and changed a couple of things.

Yes. I will fix it in both of them.

src/gui/src/PreferencesDialog_UI.ui
```diff
@@ -29,7 +29,7 @@
       <enum>QTabWidget::Rounded</enum>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
```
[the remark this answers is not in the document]
theGreatWhiteShark: Hahaha. I forget about this one every single time.

src/gui/src/InstrumentEditor/InstrumentEditor.cpp
theGreatWhiteShark: Ah, nice. I wasn't aware templates can be used in such an elegant way.

cme: On reflection, I must have been tired -- I have no idea why I seem to have started writing a `switch` and then skipped straight to the `case` value... :D

theGreatWhiteShark: Haha, Don't worry. I also jumped over it and didn't noticed.

All the copy&pasting for the `onPreferencesChanged`  did bug me too. But in the end it's content should contain class-specific assignments to member variables and updates, like `update()`, `createBackground()`, or the update of a label - and in the future the update of everything else that can be changed in the `Preferences` class through a session.

Maybe something like a  `std::vector<std::pair<memberVariable,functionPointerToPreferencesGetter*>` registered in the constructor + a custom overwrite implementations for classes requiring additional stuff like rerendering would do the trick. But the member variables are of different types and a `std::variant` is only available starting from C++17 while we AFAIK still work with C++14. And since C++ does not feature reflection, storing a string containing the name of the member variable wouldn't do any good either. So, I just sticked to the classical version

theGreatWhiteShark:
Hey @cme,

I would like to merge this one in the next couple of days because it will be quite helpful when reworking the labels and the buttons text in the GUI. Is this okay with you?

cme:
Go for it! Looks good, thanks for prompting me!

cme (approved):

Issue #1272, opened 2021-05-12
rotary_center_images.png anomalies


oddtime:
There are some graphic anomalies in files:
`data/img/gray/mixerPanel/rotary_center_images.png`
`data/img/gray/mixerPanel/rotary_images.png`

The first has one less frame, but most important is not centered:
you can compare both in gimp and see that if you shift the first image by two frames to the right, the central (green) frame corresponds to the second.

I don't know if this is correct, however it was not described in the code that one image had less frames.
Currently there's a bug at development caused by #1013, where I dropped out some lines in the rotary class (now I understand).
Way to reproduce: pan an instrument at full right, save, close and reload the song => the knob image is broken.

I plan to fix it, if you are ok I will do it together with the deprecation of members `pan_L` `pan_R` which is almost ready to be pulled

theGreatWhiteShark:
> if you are ok I will do it together with the deprecation of members pan_L pan_R which is almost ready to be pulled

Since the bug only affects development that's fine.

> There are some graphic anomalies in files:
data/img/gray/mixerPanel/rotary_center_images.png
data/img/gray/mixerPanel/rotary_images.png

Nice catch!

But I also think those pictures are quite likely to die prior to the 1.2 release as they need an update to SVGs.

oddtime:
Ok, until the SVG or other type of GUI redesign will come the current png images are ok.
The bug is fixed in #1273 where the different number of frames is taken into account.

theGreatWhiteShark:
implemented in #1282

Issue #1242, opened 2021-04-19
More about the the pan_l pan_r members


oddtime:
When we decided to keep `pan_r` and `pan_l` members in `Note` and `Instrument` class, we did it to maintain compatibility with the old song/drumkit files (see #665).

But actually we could maintain the compatibility deprecating those members (replaced by a unique pan member in [-1;1]) in `Note` and `Instrument` class, but still saving and importing the pan in the old way (using the pan law functions we have in` Sampler` class).
This avoids some conversions in the sampler, and simplifies the pan input/displaying in the GUI code reducing chance of errors.

Right?

theGreatWhiteShark:
With the design decision of early days to have two values and by maintaining backward compatibility we are bound to have some additional complexity somewhere in the code.

I agree that keeping both members will cause confusion and in the worst case even bugs. But having the classes and their file counterparts contain differing fields will cause confusion too.

With more XML knowledge up my sleeve than when I commented in #665 I would suggest to hide this complexity in the XSD file for validating the `.h2pattern` and `drumkit.xml` files and the method loading them.

oddtime:
So you would save single "pan" in xml files?

theGreatWhiteShark:
> So you would save single "pan" in xml files?

Jupp. (Since fusing them to a single value is what I thought you were intending to do).

oddtime:
Yes

Issue #1167, opened 2021-02-07
Inconsistent use of PAN_MAX in the pan law (in NotePropertiesRuler )


oddtime:
I have noticed one inconsistent use of PAN_MAX in src/gui/src/PatternEditor/NotePropertiesRuler.cpp line 487:
```
	if ( fValue > PAN_MAX ) {
		fPanL = 2*PAN_MAX - fValue;
		fPanR = PAN_MAX;
	} else {
		fPanL = PAN_MAX;
		fPanR = fValue;
	}
```

I guess it is an oversight , since there PAN_MAX = 0.5 = middle point in pan parameter domain for the widget ( i.e. [0;1] ). 
The computation is correct but PAN_MAX refers to the gain weights and not to the pan parameter, in addition if we want to use PAN_MAX explicitly in the pan law, the expressions should be different

I have modified it in pull req #1015 which is still open.

cme:
Excellent. I think this is my fault. All the more argument for commonising code like this! :)

oddtime:
> All the more argument for commonising code like this!

We could actually, we already have the pan law as static functions in `Sampler`

oddtime:
Actually this part could get even simpler... #1242

theGreatWhiteShark:
Fixed in #1273

PR #1273, opened 2021-05-13
Single pan parameter element deprecates pan_L, pan_R

oddtime:
This deprecates the member `pan_L` and `pan_R` in `Note` and `Instrument` classes, replaces them by a single `pan` member.
Discussed in #1242.
Old drumkits and songfiles are imported as well. There is a custom setter/getter for the GUI widgets issue: the pan value is there in range [0;1], while in the `Note`/`Instrument` class and `Sampler` pan law functions it is used the range [-1;1].

Fixes a bug described in #1272. Fixes some incoeherent formulas described in #1167.


src/core/CoreActionController.cpp
```diff
@@ -244,14 +232,13 @@ void CoreActionController::setStripPan( int nStrip, float fPanValue, bool bSelec
 	InstrumentList *pInstrList = pSong->getInstrumentList();
 
 	auto pInstr = pInstrList->get( nStrip );
-	pInstr->set_pan_l( fPan_L );
-	pInstr->set_pan_r( fPan_R );
+	pInstr->setPanWithRangeFrom0To1( fValue );
```
theGreatWhiteShark, line 235: Is there a benefit in keeping the range of the pan in the Mixerline from 0 to 1? How about we make it [-1,1] instead and note it done in the CHANGELOG? Since the values in existing songs will be properly mapped to the new values and the rotary is still a centered one, old users won't probably notice the difference while for new one the symmetric behavior might be more intuitive.

src/core/CoreActionController.cpp
```diff
@@ -294,20 +281,10 @@ void CoreActionController::initExternalControlInterfaces()
...
-			float fPanValue = 0.0;
-			if (fPan_R == 1.0) {
-				fPanValue = 1.0 - (fPan_L / 2.0);
-			}
-			else {
-				fPanValue = fPan_R / 2.0;
-			}
-		
-			setStripPan( i, fPanValue, false );
+			float fValue = pInstr->getPanWithRangeFrom0To1();
```
theGreatWhiteShark, line 286: I see. By changing the range we might break existing scripts relying on MIDI/OSC feedback of the pan in the range of [0,1]. But I think it's not as crucial as a mismatch between the pan value the user sets in the Mixerline and the one that is written in the .h2song file. What's your opinion on this?

theGreatWhiteShark (commented):
Looks good. Could you update the two drumkits and the demo songs in `data/` which we ship along with Hydrogen too?

src/core/CoreActionController.cpp
oddtime: I think it is ok to adopt [-1,1].

src/core/CoreActionController.cpp
oddtime: Well, we need those getters and setters for the midi at least.
I am not a user of osc, but I imagined that I may have broken it. So I trust on your opinion

What about the note pan properties ruler? Maybe the way it is now is ok as it has not an associated widget value?

oddtime:
> Looks good. Could you update the two drumkits and the demo songs in data/ which we ship along with Hydrogen too?

Yes ;)

data/DefaultSong.h2song
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <song>
- <version>1.0.0</version>
+ <version>1.1.0-'d2081b16'</version>
```
oddtime, line 3: Should I write 1.2.0 ?

theGreatWhiteShark: Yes

src/core/CoreActionController.cpp
oddtime: Ok.
You mean to write a new function `setStripPan()` for the range [-1;1]?

theGreatWhiteShark: > You mean to write a new function setStripPan() for the range [-1;1]?

Yes. Having a dedicated function associated with the new symmetric PAN endpoints will make things more consistently than adding an argument in `setStripPan`.

oddtime: Ok. Not to demand the job, but some info/example would speed up the process since I don't know how the osc syntax works.
I see that `OscServer` depends on `MidiAction` functions.
Is Osc some kind of messages that the user writes on the shell? If so, can values be fractional? I see they are generally assumed as `int` in [0;127]

theGreatWhiteShark: Okay, so first of all in order to easily test OSC commands I do recommend the `oscsend` shell program. It's bundled in the **liblo-tools** repo (AFAIR you are using a Debian-based Linux).
The following command will set the pan value of the first mixer strip to 0.9 (note that you have to enable OSC support in the PreferencesDialog (and maybe restart) first).

```bash
oscsend localhost 9000 /Hydrogen/PAN_ABSOLUTE/1 f 0.9
```

> I see that OscServer depends on MidiAction functions.
Is Osc some kind of messages that the user writes on the shell? If so, can values be fractional? I see they are generally assumed as int in [0;127]

You can think of OSC as something like an extension of MIDI to allow for various other message types, like floats, strings, and arbitrary blobs of data, to easily achieve control via a network and some other stuff. 

Unfortunately, the OSC interface is not as clean as it could be. All numerical input has to be provided as floats, even if Hydrogen just wants integer, and also functions requiring no input value at all, like stopping the playback, are registered as requiring a float input too. This is to mitigate limitations of certain OSC implementations, like TouchOSC.

`PAN_ABSOLUTE/<STRIP_NUMBER>` requires a value between 0.0 and 1.0. `PAN_RELATIVE/<STRIP_NUMBER>` either 1 or -1 (or any value other than 1). In the new version of the documentation I noted the input ranges, types, and basic information about all OSC commands. But it's not available online yet but still dormant in the other repo,

oddtime: Thanks. I have pushed a commit for this...
What would be the benefit of `PAN_RELATIVE_SYM`?

theGreatWhiteShark: > What would be the benefit of PAN_RELATIVE_SYM?

You're right. I guess none.  Sorry for the noise :)

theGreatWhiteShark:
@oddtime One of the test still fails since one of the shipped drumkits is not valid.

oddtime:
restoring
`<?xml version="1.0" encoding="UTF-8"?>`
in the drumkit files makes the test passing...

theGreatWhiteShark:
> restoring
\<?xml version="1.0" encoding="UTF-8"?>
> in the drumkit files makes the test passing...

could you push this fix? We changed the PR policy in the repo a couple of weeks ago and I can only merge if all tests are passing

theGreatWhiteShark (approved):
Looks like we are done (except for the xml version in the `drumkit.xml` files). Nice work. Thanks a lot!

oddtime:
Thanks! Fix pushed

Issue #884, opened 2020-05-26
Pulseaudio oddities


theGreatWhiteShark:
Hydrogen version *          : d29d9e0b9e81c5bca5e04100e5e86c227ad58b06
Operating system + version  :  Ubuntu 20.04
Audio driver + version      :  Pulseaudio 13.99.1

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

I just tried the Pulseaudio driver and found it is not working properly on my machine.

1. When using it in combination with the JACK server (via `module-jack-sink`), it only produces the expected output when both sample rate and buffer size match the ones of the JACK server. If this is a general thing, why not provide and fix these values for the users (apart from the additional complexity)?

2. When using it standalone, it floods the log with messages like 

```
[...]
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 760, new size = 756
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 756, new size = 268
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 268, new size = 487
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 487, new size = 537
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 537, new size = 218
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 218, new size = 269
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 269, new size = 486
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 486, new size = 538
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 538, new size = 217
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 217, new size = 761
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 761, new size = 755
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 755, new size = 269
(I) ::int H2Core::audioEngine_process(uint32_t, void*) Buffer size changed. Old size = 269, new size = 486
[...]

PR #1279, opened 2021-05-17
Fix #1275 - CoreAudio assumes nFrames equivalent to buffer len

### LINKED ISSUE #1275: CoreAudio driver does not verify audio format / sample rate parameters
Hydrogen version *          : 1.1 beta + all before
Operating system + version  :  MacOS X
Audio driver + version      :  CoreAudio

The CoreAudio audio driver for OS X attempts to set the sample rate and format, but does not 
  - check for an error returned from `AudioUnitSetProperty` when setting the format, or
  - verify that it actually got the sample rate and format it asked for.

I discovered this because my new bluetooth headphones don't support the sample rate that Hydrogen had been configured for.

That was unpleasant.

cme:
Fix #1275 - CoreAudio assumes nFrames equivalent to buffer len
    
Device buffer length is not directly related to render buffer length (as passed to audioEngine_process()), so don't use this to set the number of frames to process. Also, don't use it to update the song timer, for the same reason.
    
Reviewing the lifetime of the pointers `m_pMainBuffer_{L,R}`, the lifetimes of these buffer pointers are actually contained entirely within the lifetime of audioEngine_process(), so these pointers have been removed and replaced with local calls to getOut_{L,R}() when the buffers are actually needed.
    
This then removes the dependency on size of buffers, which solves the problem of dynamically sized buffers which can occur due to mismatched device and AudioUnit sample rates.

### DIFF
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -249,26 +249,7 @@ Instrument *			m_pMetronomeInstrument = nullptr;
 
 // Buffers used in the process function
 unsigned			m_nBufferSize = 0;
-/**
- * Pointer to the audio buffer of the left stereo output returned by
- * AudioOutput::getOut_L().
- *
- * Initialized to NULL in audioEngine_init(), assigned in
- * audioEngine_startAudioDrivers(), reset in
- * audioEngine_process_clearAudioBuffers(), and populated with the
- * actual audio in audioEngine_process().
- */
-float *				m_pMainBuffer_L = nullptr;
-/**
- * Pointer to the audio buffer of the right stereo output returned by
- * AudioOutput::getOut_R().
- *
- * Initialized to NULL in audioEngine_init(), assigned in
- * audioEngine_startAudioDrivers(), reset in
- * audioEngine_process_clearAudioBuffers(), and populated with the
- * actual audio in audioEngine_process().
- */
-float *				m_pMainBuffer_R = nullptr;
+
 /**
  * Current state of the H2Core::AudioEngine. 
  *
@@ -342,8 +323,7 @@ unsigned int			m_naddrealtimenotetickposition = 0;
  * -# It sets #m_nSongPos = -1.
  * -# It sets #m_nSelectedPatternNumber, #m_nSelectedInstrumentNumber,
       and #m_nPatternTickPosition to 0.
- * -# It sets #m_pMetronomeInstrument, #m_pAudioDriver,
-      #m_pMainBuffer_L, #m_pMainBuffer_R to NULL.
+ * -# It sets #m_pMetronomeInstrument, #m_pAudioDriver to NULL.
  * -# It uses the current time to a random seed via std::srand(). This
       way the states of the pseudo-random number generator are not
       cross-correlated between different runs of Hydrogen.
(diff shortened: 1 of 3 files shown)


cme:
@theGreatWhiteShark , @mauser if either of you have time to take a look over this, I'd appreciate the extra eyes! Particularly since it changes the lifetime of the buffers used to accumulate audio in.

src/core/Hydrogen.cpp
```diff
@@ -1453,7 +1429,7 @@ int audioEngine_process( uint32_t nframes, void* /*arg*/ )
 			+ ( finishTimeval.tv_usec - startTimeval.tv_usec ) / 1000.0;
 
 	if ( m_audioEngineState == STATE_PLAYING ) {
-		AudioEngine::get_instance()->updateElapsedTime( m_pAudioDriver->getBufferSize(),
+		AudioEngine::get_instance()->updateElapsedTime( nframes,
```
theGreatWhiteShark, line 1432: There is https://github.com/hydrogen-music/hydrogen/blob/master/src/core/IO/PulseAudioDriver.cpp#L280 which might be the cause of #884. Haven't looked into PulseAudio yet and don't know if the change is something bad or rather a cure.

theGreatWhiteShark (commented):
I can not really judge the CoreAudio specific stuff or try it on dedicated hardware. But the remainder looks good

src/core/Hydrogen.cpp
cme: Oh boy, that's a weird one.

So, that `std::min` call looks as if what it's *trying* to do is account for the possibility of PulseAudio asking for more bytes than a single buffer will hold. It's not actually doing that though because

 1. I don't think PulseAudio would actually do that, because it's been told that's the *max* buffer size
 2. It doesn't actually push the data out to PulseAudio for each call to `audioEngine_process` -- it just chunks up generation of the data, and 
 3. it takes the minimum of the number of *frames* requested, and the max number of *bytes* of the buffer. Since PulseAudio should already be ensuring the number of frames will fit in the buffer, and the number of bytes is 4 times the number of frames, the min should always just be `num_samples`.

I don't think it's the actual *cause* of either of the issues in #884 though, since I don't think it ever does anything at all. For part 2, I think I will remove that informational message as part of this change -- because it wall happen *all* the time under CoreAudio with mismatched device/AudioUnit sample rates, and it looks like PulseAudio is entirely intentionally passing variable numbers of samples to the write callback (this makes perfect sense for an audio system that may depend on network for data). In general we just shouldn't expect a fixed size buffer. The only trouble remains the use of buffer size as a proxy for latency. Hmm.

theGreatWhiteShark: > The only trouble remains the use of buffer size as a proxy for latency

The `AudioEngine::updateElapsedTime()` method is only used to display the correct time that passed since the start of the playback in the Player Control. I introduced this because the former version based on the tick size showed nonsense when either a tempo marker was passed or the tempo/position is being controlled by the JACK server. It should have no problems with dynamic sizes at all. It only cares about counting the frames passed since playback started.

cme: Yes, that use is fine -- changing it to nFrames there is, I think, exactly the right thing to do; there's another use somewhere else I meant. Hold on... [this one](https://github.com/hydrogen-music/hydrogen/blob/11b32c45f8fc88fee2423286b204c390d764462c/src/core/Hydrogen.cpp#L2775).

I think the right thing to do with this is to remove `getBufferSize` from the driver interface and add a method that returns an approximate latency. PulseAudio has an interface to get this, and drivers which use a constant buffer size can just return that by default.

cme:
Further changes after following the scope of what `getBufferSize()` actually gets used for. One inapproprate misuse, and one unneeded use for BPM in the GUI. After cleaning this up, it turns out there's no real functional dependency on buffer size after all (but maybe there *should* be but that's another issue). Excellent.

PR #1291, opened 2021-06-04
introduce translation comments, update trans. + german

theGreatWhiteShark:
@mauser I've introduced some comments describing the meaning of the translatable strings in the context of the session management (and a few others).

### DIFF
--- a/data/i18n/hydrogen_fr.ts
+++ b/data/i18n/hydrogen_fr.ts
@@ -786,6 +786,10 @@ Use &apos;Save as&apos; to enable autosave.</source>
         <translation>Le morceau est en lecture seule.
 Utilisez &apos;Sauvegarder sous&apos; pour activer la sauvegarde automatique.</translation>
     </message>
+    <message>
+        <source>Input Midi Note</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>InstrumentEditor</name>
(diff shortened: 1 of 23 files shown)


data/i18n/hydrogen_fr.ts
```diff
@@ -3183,6 +3182,7 @@ Le chemin vers le script et le nom du script doivent être sans espaces.</transl
     </message>
     <message>
         <source>Use &amp;LASH</source>
+        <extracomment>LASH (acronym for LASH Linux Audio Handler) is a protocol allowing for session management in Linux (recalling songs, preferences, and JACK port connections).</extracomment>
```
trebmuh, line 3185: LASH = Linux Audio Session Handler (if I'm not wrong)

trebmuh (changes requested):
Nice move !

data/i18n/hydrogen_fr.ts
theGreatWhiteShark: Your right, I mixed that up. But according to the https://wiki.linuxaudio.org/apps/categories/lash it's a recursive one `LASH Audio Session Handler`

trebmuh: That is right : [http://git.savannah.nongnu.org/cgit/lash.git/tree/README].

PR #1310, opened 2021-07-01
Stacked mode enhancement

cme:
Improvements to Stacked Mode.

  * Show notes from other active patterns in the "background" of the pattern editor

  * Disable / grey-out the 'stacked mode' button in Song mode. This provides the user with a good strong hint about the relationship between stacked mode and song mode!
      - grey-out Buttons when they're disabled
      - change core method from 'toggle' semantics to 'set' semantics
      - This also fixes a small bug: previously, clicking it while in Song Mode would toggle the GUI state, but not the core state, so they would get out of sync

  * Interpret clicks in the leftmost part of pattern lines ('over' the active triangles) as the same as Ctrl-clicking, since Ctrl-clicking has the direct effect of making those visible or invisible.

  * Add empty active triangle next to patterns in SongEditorPatternList when in stacked mode -- this is to suggest to the user that some new action can be carried out with the lines. Discoverability is good.

(Internal:)

  * Add core event emitted when stacked mode is enabled/disabled (the terminology for this could stand to be clarified...)

  * Emit SONG_MODE_ACTIVATION_EVENT when mode is changed from the GUI as well as when changed from OSC. This should keep everything in sync.

Also some fixes for the Song Editor pattern list, found during testing of the above:

  * Dragging a pattern now records *which* pattern is being dragged, and displays the name so the user can have more confidence in what they're doing.

      - This fixes an issue where a drag could be accidentally started without changing the selected pattern (eg. with ctrl-click), and then ended at the same place, causing the *selected* pattern to be moved to wherever was clicked

  * Drag start position was never updated, which made it look like all drags started at (0,0) so would start immediately on a mouse move.

Still to do:
  * [x] show background notes in Piano Roll editor

Pattern editor in stacked mode with notes in 'background':

image

Stacked mode button greyed out in Song mode:
image


theGreatWhiteShark (commented):
I've just been toying with it a little bit and noticed some things

- The pattern lasted played in song mode seems to be always selected
- The pattern lasted played in song mode is marked with a black triangle while all others have grey ones. When clicking the black one it turns grey but it won't disappear without toggling playback
- When deselecting all patterns none is playing -> I think this is intended, isn't it?
- Marking some patterns with grey triangles with playback stopped and switching to "single pattern mode" -> empty triangles are gone but grey ones persist
- The grey triangles can be activated in "single pattern mode" too
- The greyed out notes are not shown in the NotePropertiesRuler. But I don't think they have too.
- Greyed out notes are only shown over the length of the currently selected pattern. Even, if the other ones are longer. It's fine if it's too much work. But it made me wonder for a moment what was going on since the playhead disappeared while playback was rolling (moved into the scroll area)

Awesome feature. I like it a lot! (this will finally make me use the stacked mode :D )

cme:
> I've just been toying with it a little bit and noticed some things

Heh, most of these are things that apply to the current Stacked Mode behaviour, just they're perhaps more noticable now :)


>     * The pattern lasted played in song mode seems to be always selected

Yes, "stacked mode" really just changes how it treats the internal state of the current / next playing patterns lists. It's a bit odd perhaps but I think it's fairly useful behaviour -- you can play in Song Mode, and if you find a spot something you want to change, you can flip to pattern + stacked mode and loop what was playing when you stopped.

>     * The pattern lasted played in song mode is marked with a black triangle while all others have grey ones. When clicking the black one it turns grey but it won't disappear without toggling playback

Yeah, this is a bit odd. It's because toggling the patterns on / off doesn't take effect immediately, but at the end of the current pattern. It's shown as a light triangle to indicate it's *going* to change state, but if the player isn't running, it won't get to the event that changes its state.

I guess this is fine when the player is running, but if it's not running maybe changing the visual state immediately is better.
 
>     * When deselecting all patterns none is playing -> I think this is intended, isn't it?

I think so yes.

>     * Marking some patterns with grey triangles with playback stopped and switching to "single pattern mode" -> empty triangles are gone but grey ones persist

Another instance of the above. Definitely feels it would be better to remove them immediately when stopped.


>     * The grey triangles can be activated in "single pattern mode" too

Yes, that's a bit confusing, but again is current behaviour. Trouble is that I can see it being useful for live use by activating a fill pattern over a repeating structure for a bar or so, otherwise I think it would be easy to argue to remove it.


>     * The greyed out notes are not shown in the NotePropertiesRuler. But I don't think they have too.

I considered that, but came down on the side of leaving them out since the properties ruler already filters out a subsection of the current pattern, so it seemed reasonable and less confusing to also filter out the other patterns.

>     * Greyed out notes are only shown over the length of the currently selected pattern. Even, if the other ones are longer. It's fine if it's too much work. But it made me wonder for a moment what was going on since the playhead disappeared while playback was rolling (moved into the scroll area)

I didn't consider that! Hmm, extending the Pattern Editor in response to other patterns being played probably isn't *too* much work, but there will be lots of wrinkles I think (keyboard cursor etc) to consider. I'll look into it.

> Awesome feature. I like it a lot! (this will finally make me use the stacked mode :D )

Thank you :)

theGreatWhiteShark:
> most of these are things that apply to the current Stacked Mode behaviour, just they're perhaps more noticable now :)

Okay. Haven't really used it up to now but it explains why it's quite rough around the edges.

> Yeah, this is a bit odd. It's because toggling the patterns on / off doesn't take effect immediately, but at the end of the current pattern. It's shown as a light triangle to indicate it's going to change state, but if the player isn't running, it won't get to the event that changes its state.

I see. Takes some time to get used to it. But that's quite reasonable when using it for live performances.

Maybe it was the symbols that confused me at first since both the additional notes and the triangle (without playback rolling) are grey. Do you think a symbol representing something like "queued", "scheduled", "waiting", like [this](https://st2.depositphotos.com/1432405/12246/v/950/depositphotos_122469928-stock-illustration-sign-waiting-download-on-internet.jpg) or [that](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Ftse1.mm.bing.net%2Fth%3Fid%3DOIP.IZVFBm_SJsBU7owvK-5BCwAAAA%26pid%3DApi&f=1), would make things more clear? We could replace the pixel graphics with flat buttons having the same background color as the labels and change their icon depending on the current state (or create a bunch of buttons and toggle hide and view. Should be more efficient.) In addition, this would allow for tooltips explaining what's going on.

> Yes, that's a bit confusing, but again is current behaviour. Trouble is that I can see it being useful for live use by activating a fill pattern over a repeating structure for a bar or so, otherwise I think it would be easy to argue to remove it.

Hmm. I think I don't get it. How is this different from being in "stacked mode" and having only a single pattern selected.

> I considered that, but came down on the side of leaving them out since the properties ruler already filters out a subsection of the current pattern, so it seemed reasonable and less confusing to also filter out the other patterns.

+1


One thing that also confused me at first is that the selected pattern is not played back automatically anymore. But this is very reasonable. How about we rename "single pattern mode" into "selected pattern mode"? Since one can just select a single one this might be more descriptive.

I also liked the black triangles indicating which pattern is played back during song mode. How about them being activated regardless of which flavor of pattern mode is selected? This would be a great help to quickly get an overview in large songs and it's already implemented.

cme:
> > Yes, that's a bit confusing, but again is current behaviour. Trouble is that I can see it being useful for live use by activating a fill pattern over a repeating structure for a bar or so, otherwise I think it would be easy to argue to remove it.
> 
> Hmm. I think I don't get it. How is this different from being in "stacked mode" and having only a single pattern selected.

The only difference is that in Single Pattern Mode, an additional pattern that's activated will automatically be deactivated when playback gets to the end of the pattern, effectively making it a 'one-shot', while in stacked pattern mode, it would have to be manually deactivated.

> One thing that also confused me at first is that the selected pattern is not played back automatically anymore. But this is very reasonable. How about we rename "single pattern mode" into "selected pattern mode"? Since one can just select a single one this might be more descriptive.

This seems pretty reasonable, yes!

> I also liked the black triangles indicating which pattern is played back during song mode. How about them being activated regardless of which flavor of pattern mode is selected? This would be a great help to quickly get an overview in large songs and it's already implemented.

I like this! +1

PR #1312, opened 2021-07-03
namespace closing should go inside #ifdef #endif

charbeljc:
I catched this one while generating python bindings.


theGreatWhiteShark (approved):
Thanks

PR #1314, opened 2021-07-05
Sample editor instant feedback

### LINKED ISSUE #811: sample editor : end marker jumps to start marker position
Hydrogen version *          : 7283fcabce84e5575104746135fc6d7cede29e51
Operating system + version  : ubuntu studio 16
Audio driver + version      : jack

to reproduce : 

- open sample editor
- grab the start marker
- End-marker jumps to the Start-marker position

This causes previous edits to be lost : 

- tweak the end-marker position of a sample
- save your edit
- open the same sample again in the editor
- grab the start-marker
- your end-marker jumps to the start and your previous work is lost

### LINKED ISSUE #1304: Further SampleEditor tweaks
-  Open the SampleEditor -> enter a large value in the spinbox of the "Start" and press enter -> both the values in the spinboxes of "Start" and "Loop" get updated but only the "S" line gets moved in the wave display while "L" remains at zero.
- Open the SampleEditor -> change something and hit the "Close" button -> a dialog "Unsave changes..." will pop up and the user has to consent to discard the changes -> clicking "Ok" and reopen the SampleEditor by clicking "Edit Layer" again -> the same popup shows up again. I think the second one is a dud.
- When entering a fresh instance of the SampleEditor the value of the EndFrameSpinBox looks perfectly fine. But as soon as the mouse pointer enters the MainSampleWaveDisplay it is reset to 0 by SampleEditor::getAllFrameInfos() what causes some other oddities. The slider, however, stays at it's correct position.
- The first value change in the Start, Loop, and End spinboxes is discarded and, instead, the slider is shown in the DetailWaveDisplay. It would be nice to update the DetailWaveDisplay as soon as the spinboxes have focus and make the first change affect the value as well.

charbeljc:
Update target wave display when loop, rubberband or envelope are edited.

Also should close #811
Also should close #1304 

There are still strange things  with rubberband, but we should be close to something usable.


src/core/Basics/Sample.cpp
```diff
@@ -51,18 +50,18 @@ static RubberBand::RubberBandStretcher::Options compute_rubberband_options( cons
 
 
 /* EnvelopePoint */
-EnvelopePoint::EnvelopePoint() : Object( EnvelopePoint::__class_name ), frame( 0 ), value( 0 )
```
theGreatWhiteShark, line 54: How come you removed the `H2Core::Object` in here? It's used for debugging and shouldn't add too much weight.

charbeljc: > How come you removed the `H2Core::Object` in here? It's used for debugging and shouldn't add too much weight.

Hum, I was maybe a bit to quick with this. I think EnvelopePoint should go away in favor of a proper Envelope class.
Besides, I have a pending patch which removes the H2Core::Object's overhead... 

My point was to get rid of the unique_ptr around EnvelopePoint in envelope, so I can access them. I don't remember the details right now, but pybind11 caused me some troubles with those std::shared_ptr<std::unique_ptr<EnvelopePoint>> ...

ihmo, EnvelopePoints should be plain dummy structs with copy semantics, and let the (to be written) Envelope let deal with them.

charbeljc: Hi @theGreatWhiteShark I just restored EnvelopePoint inheritance of H2Core::Object.
Regards

theGreatWhiteShark: > I think EnvelopePoint should go away in favor of a proper Envelope class.
Besides, I have a pending patch which removes the H2Core::Object's overhead...

Could you split off these changes and make a separate PR? Such fundamental changes in the code base require the consent of at least a majority of all maintainers. This will happen much faster if it's a concise one.

One thing to think about in advance (and I haven't done yet, so, there is a chance none of this makes sense): The envelope in the sample editor is quite similar to the automation path. Maybe there could be a common basis.

charbeljc: > > I think EnvelopePoint should go away in favor of a proper Envelope class.
> > Besides, I have a pending patch which removes the H2Core::Object's overhead...
> 
> Could you split off these changes and make a separate PR? Such fundamental changes in the code base require the consent of at least a majority of all maintainers. This will happen much faster if it's a concise one.
> 

Yes, I will do a separate PR on this.

> One thing to think about in advance (and I haven't done yet, so, there is a chance none of this makes sense): The envelope in the sample editor is quite similar to the automation path. Maybe there could be a common basis.

I will have a look at those automation path. If we could factor out some code, it would be nice.

charbeljc: Please see #1318

src/gui/src/SampleEditor/SampleEditor.cpp
```diff
@@ -931,11 +976,38 @@ void SampleEditor::valueChangedProcessingTypeComboBox( const QString unused )
...
+	if ( ! m_pTargetSampleView ) {
+		// qWarning() << "no TargetSampleView!";
+		return;
+	}
+	switch ( index ){
+		case 0 ://
+			m_pTargetSampleView->setEditMode( SampleEditor::EnvelopeType::VelocityEnvelope );
+			break;
+		case 1 ://
+			if ( m_pTargetSampleView ) {
```
theGreatWhiteShark, line 999: This sanity check and the one below should already covered by the one in the beginning of the function

src/gui/src/SampleEditor/TargetWaveDisplay.cpp
```diff
@@ -204,45 +206,54 @@ void TargetWaveDisplay::paintLocatorEventTargetDisplay( int pos, bool updateposi
...
+	// qWarning() << "TargetWaveDisplay::updateDisplay: layer:" << pLayer;
+	if ( ! pLayer ) {
+		return;
+	}
+	updateDisplay( pLayer->get_sample(), pLayer->get_gain() );
+}
+
+void TargetWaveDisplay::updateDisplay( std::shared_ptr<H2Core::Sample> sample, double gain )
+{
+	// qWarning() << "TargetWaveDisplay::updateDisplay: sample:" << sample;
```
theGreatWhiteShark, line 218: The original version of `updateDisplay` also included a check whether `pLayer->get_sample()` is a `nullptr`. Could you add it in here too just to be save?

theGreatWhiteShark (commented):
I like this signal-based version for communicating changes in the envelope or sliders more.

However, some parts are now broken:

- Looping doesn't seem to work anymore and if the loop count is set to a value higher than 0 there are visual artifacts in the `DetailWaveDisplay`
- Pitch settings using Rubberband does not seem to take any effect either.
- Previously, when hitting play the white slider indicating current playback position was visible in the TargetWaveDisplay too
- (unrelated to this PR but I just noticed it) when increase the loop count to very large values, like 300, Hydrogen freezes. We probably should add some sanity checks in there

charbeljc:
> I like this signal-based version for communicating changes in the envelope or sliders more.
> 
> However, some parts are now broken:
> 
>     * Looping doesn't seem to work anymore and if the loop count is set to a value higher than 0 there are visual artifacts in the `DetailWaveDisplay`

What a mess... I guess I did let this PR in the middle of nowhere... Thank you for your review, I'm finally back on track.
> 
>     * Pitch settings using Rubberband does not seem to take any effect either.
> 

Are your working with the cli version or the library one ?

>     * Previously, when hitting play the white slider indicating current playback position was visible in the TargetWaveDisplay too
>

This should be restored now.

>     * (unrelated to this PR but I just noticed it) when increase the loop count to very large values, like 300, Hydrogen freezes. We probably should add some sanity checks in there

Hum, this don't hang on my machine, but yes, this must be addressed.

theGreatWhiteShark:
Hey @charbeljc ,

Can you rebase this PR on master? It will be my next read.

PR #1316, opened 2021-07-06
small fix: Sample.is_empty() was always returning false

charbeljc:
Another catch from python bindings

```python
from h2core import Sample
sample = Sample.load_static('/path/to/sample.flac')
sample.unload()
assert sample.is_empty()
sample.load()
assert not sample.is_empty()
```


theGreatWhiteShark (approved):
Nice.

PR #1317, opened 2021-07-06
raw pointers are evil (share the Song)

charbeljc:
Use shared pointers for Song objects everywhere to avoid ownership problems with python bindings.

### DIFF
--- a/src/core/AudioEngine.cpp
+++ b/src/core/AudioEngine.cpp
@@ -294,7 +294,7 @@ int AudioEngine::start( bool bLockEngine, unsigned nTotalFrames )
 	m_nPatternTickPosition = 0;
 
 	// prepare the tick size for this song
-	Song* pSong = Hydrogen::get_instance()->getSong();
+	std::shared_ptr<Song> pSong = Hydrogen::get_instance()->getSong();
 	m_pAudioDriver->m_transport.m_fTickSize =
 		AudioEngine::computeTickSize( static_cast<float>(m_pAudioDriver->getSampleRate()), pSong->getBpm(), pSong->getResolution() );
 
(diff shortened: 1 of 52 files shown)


theGreatWhiteShark (approved):
Jej. Thanks!

charbeljc:
> Jej. Thanks!

Ur welcome!

src/core/AudioEngine.cpp
```diff
@@ -787,7 +787,7 @@ void AudioEngine::restartAudioDrivers()
 	startAudioDrivers();
 }
 
-void AudioEngine::processCheckBPMChanged(Song* pSong)
+void AudioEngine::processCheckBPMChanged(std::shared_ptr<Song> pSong)
```
Floessie, line 790: You can generally save some refcounting by passing `std::shared_ptr<T>` as `const std::shared_ptr<T>&` or `const std::shared_ptr<const T>&`.

elpescado: I'm not a fan of passing smart pointers by reference, as it's confusing and does not convey ownership semantics clearly. I'd suggest either:

1. `std::shared_ptr<Song>` if function participates in ownership of song, or
2. `Song &` otherwise.

There's nice [explaination](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/) of this on Herb Sutter's blog:

> Pass by `*` or `&` to accept a `widget` independently of how the caller is managing its lifetime. Most of the time, we don’t want to commit to a lifetime policy in the parameter type, such as requiring the object be held by a specific smart pointer, because this is usually needlessly restrictive. As usual, use a * if you need to express null (no widget), otherwise prefer to use a &; and if the object is input-only, write const widget* or const widget&.

PR #1318, opened 2021-07-09
use copy semantics for EnvelopePoint

charbeljc:
The main motivation for this patch comes from my work on python bindings. Wrapping std::vector<std::unique_ptr<Thing>> is a recipe for disaster...


theGreatWhiteShark (approved):
+1

PR #1319, opened 2021-07-09
Get rid of the __class_name parameter in H2Core::Object constructor

charbeljc:
The main motivation is the ability to expose abstract classes (AudioOutput, MidiInput, MidiOutput) to python bindings (and subclass them in python).
This also fixes the case of some H2Core::Object subclasses for which __class_name was never defined / initialized.

To keep it short, the macro H2_OBJECT is changed to take the class name as parameter
```c++
class Foo: public H2Core::Object {
    H2_OBJECT(Foo)
public:
    Foo() : Object() {}
    ...
};
```


src/gui/src/Director.cpp
```diff
@@ -65,7 +65,7 @@ using namespace H2Core;
 
 Director::Director ( QWidget* pParent )
 		: QDialog ( pParent )
-		, Object ( "Director" )
```
theGreatWhiteShark, line 68: The corresponding header file `Director.h` was not modified.

Hmm. Turns out it did not contain the `H2_OBJECT` macro. Could you add it?

src/gui/src/SongPropertiesDialog.h
```diff
@@ -32,7 +32,7 @@
  */
 class SongPropertiesDialog : public QDialog, private Ui_SongPropertiesDialog_UI
 {
-    H2_OBJECT
+    H2_OBJECT(SongPropertiesDialog)
```
theGreatWhiteShark, line 35: This one is odd. The class itself does not inherit from `H2Core::Object`. If think it's save to remove the macro.

theGreatWhiteShark (commented):
> The main motivation is the ability to expose abstract classes (AudioOutput, MidiInput, MidiOutput) to python bindings (and subclass them in python).
This also fixes the case of some H2Core::Object subclasses for which __class_name was never defined / initialized.
>
> To keep it short, the macro H2_OBJECT is changed to take the class name as parameter
>
> class Foo: public H2Core::Object {
>    H2_OBJECT(Foo)
> public:
>   Foo() : Object() {}
>    ...
> };

I like the idea. All these `__class_name` variables also do flood the terminal with warnings when building the doxygen developer documentation causing more important warnings to slip through.

I'm afraid having `Object::class_name()` as a `static` member won't work.

In the constructor of `Object` the `add_object()` method is called in order to add the `class_name` to the object map (the one shown at the very end of the log after starting hydrogen using the `-VDebug` command-line option and quitting it). That's a huge help when tracking down memory leakage. With these changes, however, all object will be just entered as "Object" and not with their actual class names. 

I'm not an expert in C++ but this [comment at Stackoverflow](https://stackoverflow.com/questions/34222703/how-to-override-static-method-of-template-class-in-derived-class/34222770#34222770) seems to hold one candidate solution.

charbeljc:
> I'm afraid having `Object::class_name()` as a `static` member won't work.
> 
> In the constructor of `Object` the `add_object()` method is called in order to add the `class_name` to the object map (the one shown at the very end of the log after starting hydrogen using the `-VDebug` command-line option and quitting it). That's a huge help when tracking down memory leakage. With these changes, however, all object will be just entered as "Object" and not with their actual class names.
> 
> I'm not an expert in C++ but this [comment at Stackoverflow](https://stackoverflow.com/questions/34222703/how-to-override-static-method-of-template-class-in-derived-class/34222770#34222770) seems to hold one candidate solution.

I did have overlooked the `add_object()` mess ... This gave me an hard time to restore counting functionality with this rather pythonic pattern which don't fit well in C++, so I ended up killing completely `add_object()` in favor of a template class
`Countable<T>` following the lovely Curiously Recursive Template Pattern you pointed at me!

The only downside is that we have to explicitly register class names with the associated counters in `init_core_object_map()` or `init_gui_object_map()`.

The new form for H2Core::Object subclasses becomes

```c++
class Foo : public H2Core::Countable<Foo>, public H2Core::Object {
  H2_OBJECT(Foo)
public:
 ...
};
```

The counting is also cumulative, so Object counters contains the grant total of constructor/destructor calls.

theGreatWhiteShark (commented):
Hey @charbeljc,

I'm not done yet looking through your changes. But I have two general questions:

- How come you are preparing the `__objects_map` using the `init_core_object_map()` and `init_gui_object_map()` functions? Shouldn't the previous code of `Object::add_object()` using the revised `class_name()` and `_class_name()` pair do the same without the extra burden of keeping the objects with the `H2_REGISTER` calls in sync?
- I understand that you want to expose e.g. `AudioOutput` to allow creating custom audio drivers using Python. But there shouldn't be raw `Object` instances but only derived classes, right? Wouldn't it then be more concise to merge `Object` and `Countable` in order to have a `Object<ClassName>` template?

src/core/Object.h
```diff
@@ -29,39 +29,119 @@
...
+			}
+#endif
+		}
+		Countable(const Countable<T> &other) {
+#ifdef H2CORE_HAVE_DEBUG
+			if( __logger && __logger->should_log( Logger::Constructors ) ) {
+				__logger->log( Logger::Debug, nullptr, T::_class_name(), "Copy Constructor" );
+			}
+			if ( __count ) {
+				++counters.destructed;
```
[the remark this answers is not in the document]
charbeljc: This one is a plain typo, it should read ++counters.constructed, of course. Strange that I didn't caught this one since I just restored the constructor/destructor counting... Hum, maybe that's could mean I introduced the first actual copy operation with my work on envelope points... I'll rebase this one to see what we get.

src/core/Object.h
```diff
@@ -29,39 +29,119 @@
...
+		static bool __count;               ///< should we count class instances
+		static std::atomic<int> __objects_count;        ///< total objects count
+		static Logger * __logger;
+};
+
+template<typename T> class Countable: public CountControl {
+	public:
+		Countable() {
+#ifdef H2CORE_HAVE_DEBUG
+			if ( __logger && __logger->should_log( Logger::Constructors ) ) {
```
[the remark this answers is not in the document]
charbeljc: Yes, I actually discovered It will playing with my python bindings. I will have a look to expose it to main.cpp/cmd line.

charbeljc:
> Hey @charbeljc,
> 
> I'm not done yet looking through your changes. But I have two general questions:
> 
>     * How come you are preparing the `__objects_map` using the `init_core_object_map()` and `init_gui_object_map()` functions? Shouldn't the previous code of `Object::add_object()` using the revised `class_name()` and `_class_name()` pair do the same without the extra burden of keeping the objects with the `H2_REGISTER` calls in sync?

The problem here is that you can't rely on sane behavior calling a virtual method in any constructor. This is completely undefined behavior in c++. So to achieve on demand class registry when first object is constructed, we should have to enforce object's creation only with a factory function. I guess it would be to much changes to the codebase. Hum, your next comment about making H2Core::Object a class template might lead to a solution here... I will try this.
> 
>     * I understand that you want to expose e.g. `AudioOutput` to allow creating custom audio drivers using Python. But there shouldn't be raw `Object` instances but only derived classes, right? Wouldn't it then be more concise to merge `Object` and `Countable` in order to have a `Object<ClassName>` template?

I guess this is quite sound, and might be the final piece of the puzzle I'm trying to solve :-) I was a bit conservative on this, I guess.

src/gui/src/Director.cpp
charbeljc: done

src/gui/src/SongPropertiesDialog.h
charbeljc: done

src/core/Object.h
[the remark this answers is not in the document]
theGreatWhiteShark: You don't have to do now. I still have on my TODO list to reorganize the log anyway. Right now it just floods the terminal as soon as the drumkits are slightly out of date and it gets hard to notice the important stuff while debugging.

src/gui/src/SongPropertiesDialog.h
theGreatWhiteShark: I wonder whether I have been a little premature on this one. Can you think of a reason why any of the classes defined in src/core and src/gui should _not_ be based on the `Object` template? In terms of debugging, especially detecting memory leakage, it would certainly be useful to derive them all.

theGreatWhiteShark (commented):
Nice. I like the changes a lot. Now the slightly increased complexity of `Object` is a little price to pay to get rid off all the `__class_name` initialization.

A minor search and replace thingy: the empty `Object()` initialization in the implementations of the derived classes isn't required anymore, isn't it? (Well, apart from the copy constructors)

charbeljc:
> Nice. I like the changes a lot. Now the slightly increased complexity of `Object` is a little price to pay to get rid off all the `__class_name` initialization.
> 
> A minor search and replace thingy: the empty `Object()` initialization in the implementations of the derived classes isn't required anymore, isn't it? (Well, apart from the copy constructors)

Right, done with my last commit.

src/gui/src/SongPropertiesDialog.h
charbeljc: Yes, you are right, now that H2Core::Object is really lightweight, all Hydrogen classes should inherit from it, so we can actually track instances accurately. I think we could merge this PR in development branch, and complete the work of making all classes from gui inherit H2Core::Object as we touch/refactor them. What do you think ?

theGreatWhiteShark: > I think we could merge this PR in development branch, and complete the work of making all classes from gui inherit H2Core::Object as we touch/refactor them. What do you think ?

I would prefer doing this for all classes at the same time. It would take quite a while to refactor/touch all the classes and there are already so many other things to do (we have a ratio of open to solved issues of 1 : 3. There is still a lot of room at the top.)

src/core/Object.cpp
```diff
@@ -133,17 +80,20 @@ void Object::write_objects_map_to( std::ostream& out, object_map_t* map ) {
...
+		map = &snapshot;
 	}
 	
 	std::ostringstream o;
 	pthread_mutex_lock( &__mutex );
 	object_map_t::iterator it = map->begin();
 	while ( it != map->end() ) {
-		o << "\t[ " << std::setw( 30 ) << ( *it ).first << " ]\t" << std::setw( 6 ) << ( *it ).second.constructed << "\t" << std::setw( 6 ) << ( *it ).second.destructed
-		  << "\t" << std::setw( 6 ) << ( *it ).second.constructed - ( *it ).second.destructed << std::endl;
+		if ( it->second.constructed || it->second.destructed ) {
```
theGreatWhiteShark, line 93: Shouldn't this be a `&&`?

theGreatWhiteShark (approved):

theGreatWhiteShark:
There seems to be something messing with the Windows tests. With our new (quite reasonable) policy merging is only possible when all tests pass.

Floessie:
Just two remarks that came to my mind when skimming through this PR:

@charbeljc 
> The problem here is that you can't rely on sane behavior calling a virtual method in any constructor. This is completely undefined behavior in c++.

Though it surely isn't a good idea, it is still well defined in C++: it will call the method in the current hierarchy level (the vtable is adjusted prior to the c'tor).

@theGreatWhiteShark 
Are you aware that double underscores prefixes are [reserved for the compiler/library implementation](https://stackoverflow.com/a/228797)?

HTH,
Flössie

theGreatWhiteShark:
Hey @Floessie,

Thanks for the insights!

> Are you aware that double underscores prefixes are reserved for the compiler/library implementation?

Well, TBH I wasn't aware of the exact reason its usage is a no-no but there is already a rough coding standard in the [DEVELOPERS](https://github.com/hydrogen-music/hydrogen/blob/master/DEVELOPERS) file telling to use `m_*` to indicate member variables in Hydrogen instead. Thing is, some parts of the code are really old (we don't know for sure how old but the first working version was released more than 20 years ago). So, it gets more rocky the further you descend. Our current policy is to only refactor it when rewriting larger parts (during development). It's a little bit more on the "don't change a running system" than on the modernization side of the tradeoff since our goal is stability over new features.

theGreatWhiteShark:
Hey @charbeljc 

Could you merge the latest commits of the `development` branch into this one? Using them the CI build and tests will pass and I can merge this PR. And could you do so without force pushing? Else all of the files will be shown as dirty again in the Github web interface.

charbeljc:
Hi @theGreatWhiteShark sorry for the delay, I've been quite busy last days. I just merged development into my branch.

src/core/Object.cpp
charbeljc: > Hey @Floessie,
> 
> Thanks for the insights!
> 
> > Are you aware that double underscores prefixes are reserved for the compiler/library implementation?
> 
> Well, TBH I wasn't aware of the exact reason its usage is a no-no but there is already a rough coding standard in the [DEVELOPERS](https://github.com/hydrogen-music/hydrogen/blob/master/DEVELOPERS) file telling to use `m_*` to indicate member variables in Hydrogen instead. Thing is, some parts of the code are really old (we don't know for sure how old but the first working version was released more than 20 years ago). So, it gets more rocky the further you descend. Our current policy is to only refactor it when rewriting larger parts (during development). It's a little bit more on the "don't change a running system" than on the modernization side of the tradeoff since our goal is stability over new features.

About the project's coding standards, I think they could use a little refresher. I find this utterly ugly for instance. 
```c++
if( condition ){
  ...
}
```
I mean, control flow are not functions.
```c++
if ( condition ) {
  ...
}
```
please :grin:

charbeljc:
> Just two remarks that came to my mind when skimming through this PR:
> 
> @charbeljc
> 
> > The problem here is that you can't rely on sane behavior calling a virtual method in any constructor. This is completely undefined behavior in c++.
> 
> Though it surely isn't a good idea, it is still well defined in C++: it will call the method in the current hierarchy level (the vtable is adjusted prior to the c'tor).
> 
> @theGreatWhiteShark
> Are you aware that double underscores prefixes are [reserved for the compiler/library implementation](https://stackoverflow.com/a/228797)?
> 
> HTH,
> Flössie

Yeah, you are right, It would be more accurate to say that they does not work as we, mortals, expect them to work :grin:

src/core/Object.cpp
charbeljc: > Shouldn't this be a `&&`?

In fact, if something goes plain wrong, `||` is better in this case. Suppose we allocate some object of some type and never delete it.

theGreatWhiteShark:
> I've been quite busy last days

As you might have noticed I've been quite busy as well. Sorry for the delay from my side too

PR #1344, opened 2021-08-06
PortAudio: select output device in preferences (fix #611)

### LINKED ISSUE #611: no way to choose device for portaudio (critical for windows version)
Device option, in settings, needs to provide list of existing devices. At least for portaudio (not bothering too much about pulseaudio/alsa). Even with missing jack it is possible to use asio jackrouter, but even current text field is blocked for all drivers besides alsa.

update: Looks like it is hardcoded to use default device.

cme:
This changes the 'audio device' text box to an editable combo box, and gives an interface for populating the menu of the combo box, and provides the list of devices for PortAudio.

image

Should be easy enough to populate for the other drivers as well.


cme:
There's lots that could be improved in this area, like
  - abstract driver details from PreferencesDialog etc. and provide some sort of driver registry / broker to list drivers and devices
  - more useful information for the display at the right.

src/core/IO/PortaudioDriver.cpp
```diff
@@ -76,6 +80,26 @@ int PortAudioDriver::init( unsigned nBufferSize )
 	return 0;
 }
 
+// List devices
+QStringList PortAudioDriver::getDevices() {
+	if ( ! m_bInitialised ) {
```
cme, line 85: This seems to cause problems on Linux if PortAudio wants to wait on the underlying ALSA device but it's in use by one of the other drivers. Yuck.

cme: It's starting to look like to safely query one driver about its available devices, we'll have to actively stop the current audio system to start the other one. While this is a bit cleaner in that it allows the listing method to be non-static (and virtual for some refactoring of the preferences dialog), it means that selecting a new audio system and device means *two* audio restarts.

theGreatWhiteShark:
@cme I thought about making the `Auto` driver not just try to connect to a selected device - which isn't set on a fresh install - and the default one while starting driver but iterate through all available devices (and storing the result). Do you think that would be worth the effort?

My motivation is #1479 which might not have occurred with a more thorough search of the Auto search.

PR #1348, opened 2021-08-17
Fix CLI driver selection

cme:
* Make driver id on command line case-insensitive
  * Add PulseAudio, CoreAudio, PortAudio to the GUI app's command line

Really there should be a driver registry or similar to handle this sort of thing

### DIFF
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -230,6 +230,7 @@ int main(int argc, char *argv[])
 			exit(0);
 		}
 
+		sSelectedDriver = sSelectedDriver.toLower();
 		if (sSelectedDriver == "auto") {
 			preferences->m_sAudioDriver = "Auto";
 		}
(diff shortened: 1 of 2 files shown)


theGreatWhiteShark:
Just a thought. Which drivers are supported for a specific version/platform of Hydrogen should be known at compile time, right? You introduced a list of available drivers in `startAudioDriver()`. Maybe we could create this list at compile time and use it for both the driver selection, the parsing of the command line arguments, and the messages printed using `hydrogen --help`. After all, the code seems to be tailored for Linux versions only but selecting the driver from CLI may be useful on Mac or Windows too.

cme:
> Just a thought. Which drivers are supported for a specific version/platform of Hydrogen should be known at compile time, right? You introduced a list of available drivers in `startAudioDriver()`. Maybe we could create this list at compile time and use it for both the driver selection, the parsing of the command line arguments, and the messages printed using `hydrogen --help`. After all, the code seems to be tailored for Linux versions only but selecting the driver from CLI may be useful on Mac or Windows too.

Yeah, after thinking about this a little today, I want to refactor these things into a single driver registry class that knows about all the drivers available, their names etc, and can also store some more status for display in the driver info text box (because dumping driver problems to the errorlog doesn't really seem the most helpful way of telling the user when things change.

Also having been in the PreferencesDialog code for adding PortAudio things, having a single source for driver info would make a lot of that code a lot cleaner too.

PR #1349, opened 2021-08-18
Further work on improving Portaudio driver device selection

cme:


src/core/IO/PortaudioDriver.cpp
```diff
@@ -80,27 +80,60 @@ int PortAudioDriver::init( unsigned nBufferSize )
...
+		HostAPI = Pa_GetHostApiInfo( Pa_GetDefaultHostApi() )->name;
 	}
 
 	QStringList devices;
 	int nDevices = Pa_GetDeviceCount();
 	for ( int nDevice = 0; nDevice < nDevices; nDevice++ ) {
 		const PaDeviceInfo *pDeviceInfo = Pa_GetDeviceInfo( nDevice );
+
+		// Filter by API
+		if ( Pa_GetHostApiInfo( pDeviceInfo->hostApi )->name != HostAPI ) {
```
elpescado, line 121: How about we don't use string compare and instead compare host APIS by [`type`](http://www.portaudio.com/docs/v19-doxydocs/structPaHostApiInfo.html#a5424bfcdf9f73b17cddc89e827a45d3f)? [`PaHostApiTypeId`](http://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a8eaebe3d39c5ea45598da8f86dc2e5ae) is a "unchanging unique identifiers for each supported host API" so it sounds as it fits the purpose here.

elpescado: ...or maybe get [`PaHostApiIndex`](http://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#aeef6da084c57c70aa94be97411e19930) of selected Host API before loop, and then compare that index to `pDeviceInfo->hostApi`?

cme: > How about we don't use string compare and instead compare host APIs by [`type`](http://www.portaudio.com/docs/v19-doxydocs/structPaHostApiInfo.html#a5424bfcdf9f73b17cddc89e827a45d3f)? [`PaHostApiTypeId`](http://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a8eaebe3d39c5ea45598da8f86dc2e5ae) is an "unchanging unique identifier for each supported host API" so it sounds as it fits the purpose here.

Yeah, I thought about this because the string compare in the loop feels unnecessary (and somehow "sloppy"), but PortAudio doesn't provide an interface to get a HostAPI by name, so finding the `PaHostApiTypeId` or `PaHostApiIndex` to compare would mean looping over HostAPI indices before looping over device indices, which makes for more complex code. Since HostAPI names are short anyway, the performance difference is a wash, and not important since this is only ever invoked by explicit user action.

In the end I came down on the side of accepting the "sloppy" feeling of a string compare in the loop, as it made the code simpler and more readable than looping beforehand. If there were a name -> HostAPI interface in PortAudio, I'd probably have gone the other way.

cme: (There are other alternatives that get might better performance, but would definitely be unwarranted complexity :D )

Issue #1352, opened 2021-08-23
Audio latency on Windows (PortAudio) is awful


cme:
Hydrogen version *          : 
Operating system + version  : Windows
Audio driver + version      :  PortAudio

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

The latency on Windows is awful.

This is because using the default PortAudio stream makes a very pessimistic assumption about the latency target, in order to be a 'safe' fallback, so uses the audio device's "high" latency parameter.

Hydrogen has fairly minimal audio processing requirements so we can almost certainly use a more optimistic default latency target, but we shouldn't do this unless we also provide a way of setting the latency target through the GUI, to make sure anyone experiencing problems can adjust it.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

*  H2 version : see 'Info - About', or in case you build Hydrogen from source use 'git rev-parse HEAD'
** if you are reporting a bug, please describe how the bug can be reproduced

cme:
Fixed with #1388

PR #1388, opened 2021-10-03
Reduce PortAudio latency (mostly for Windows' sake)

cme:
Reduce PortAudio latency by using variable-length buffers, and by allowing a target latency to be specified (rather than just accepting the default conservative latency from the driver)

This also changes the colours used for disabled widgets, because I spent 10 minutes wondering why it looked like the 'buffer' textentry wasn't being disabled 😆 

Might be good to amend this to feed back the 'actual' latency value to the target latency text box, but there's the possibility that this might lead to unstable iteration of the preferences value.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
 		  pattern inserted in SongEditor (#932)
 		- fix display of tempo marker while loading a song
 		  (introduced in 1.1.0) (#1393)
+	* Explicitly set latency target for PortAudio (Windows) audio driver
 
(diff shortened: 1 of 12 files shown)


cme:
This change is to address #1352.

theGreatWhiteShark:
Hey @cme ,

Should we do a second PR containing these changes against master or is this intended for a later point in time?

cme:
> Should we do a second PR containing these changes against master or is this intended for a later point in time?

I don't think so; 1.1.1 should be released soon and the changes merged onto master as a batch from 1.1.1, so would be a single PR, unless you think it's more pressing?

Issue #1395, opened 2021-10-24
Scripting integration


antonio-petricca:
Does hydrogen integrate a scripting language?

theGreatWhiteShark:
Hey @antonio-petricca ,

I added the new commands to the OSC API. Could you check out and test the code in #1404 ?

theGreatWhiteShark:
Closed with #1404

PR #1404, opened 2021-11-13
implement Pattern & Grid OSC commands

theGreatWhiteShark:
- logic for creating a new/loading an existing/removing a pattern has been moves to H2Core
- logic for toggling a grid cell in the song editor has been moved to H2Core
- all the above actions have been exposed via the OSC API

New OSC API commands are
- `oscsend localhost 9000 /Hydrogen/NEW_PATTERN s "patternName"`
- `oscsend localhost 9000 /Hydrogen/OPEN_PATTERN s "/absolute/path/to/pattern.h2pattern"`
- `oscsend localhost 9000 /Hydrogen/REMOVE_PATTERN f 0`
- `oscsend localhost 9000 /Hydrogen/SONG_EDITOR_TOGGLE_GRID_CELL ff 2 3`

implements #1395

### DIFF
--- a/src/gui/src/SongEditor/SongEditorPanel.cpp
+++ b/src/gui/src/SongEditor/SongEditorPanel.cpp
@@ -513,36 +513,6 @@ void SongEditorPanel::newPatBtnClicked()
 	delete pDialog;
 }
 
-
-void SongEditorPanel::insertPattern( int idx, Pattern* pPattern )
-{
-	Hydrogen	*pHydrogen = Hydrogen::get_instance();
-	std::shared_ptr<Song> pSong = pHydrogen->getSong();
-	PatternList *pPatternList = pSong->getPatternList();
-
-	pPatternList->insert( idx, pPattern );
-	pHydrogen->setSelectedPatternNumber( idx );
-	pSong->setIsModified( true );
-	updateAll();
-}
-
-void SongEditorPanel::deletePattern( int idx )
-{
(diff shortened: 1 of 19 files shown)


src/gui/src/SongEditor/SongEditorPanel.cpp
```diff
@@ -950,3 +920,7 @@ void SongEditorPanel::timelineActivationEvent( int nEvent ){
 	
 	m_pPositionRuler->createBackground();
 }
+
+void SongEditorPanel::updateSongEditorEvent( int ) {
+	updateAll();
```
cme, line 925: This has an undesirable side-effect (`updateAll` -> `resyncExternalScrollbar`) of scrolling the Song Editor viewport to bring the currently selected pattern into view, which was added so that newly created patterns are visible.

When the user clicks to set a cell in the song editor, but the current pattern isn't visible, the Song Editor will move after the click to show the currently selected pattern.

I'm making some changes to song editor scrolling, so I'll fix it there.

theGreatWhiteShark: Ah, your right. Sorry for that!

theGreatWhiteShark: But, on the other hand, for setting/removing a pattern resyncing the scrollbar is important since this also changes the current pattern number. Probably calls for different arguments provided to the event.

PR #1405, opened 2021-11-15
CLI audio driver selection + tweaks

theGreatWhiteShark:
- If the peak value supplied to the Fader is `Inf` (e.g. when using LADSPA plugins), a WARNING was shown in the log (well, a lot of them actually).
- show all available audio drivers in the CLI help message (and only them)
- store the changes in Preferences done by the CLI options. Otherwise opening the PreferencesDialog would overwrite them (and a different audio driver will be shown).
- introduce Locks log level: it will print out INFO, WARNING, ERROR, as well as all locking and unlocking of the AudioEngine including the function and line which issued the locking.
-  Mixer: fix Ladspa bypass button: It was out of sync. When checked, the effect was active. In addition to fixing this I renamed all isFxActive and similar functions to isFxBypassed to make handling them more straight forward.
-  SongEditor: use ActionMode from last session: instead of overwriting the stored Song::ActionMode value with selectMode, the one from the .h2song file will be used

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 		- Remembering paths in all export/import/save/open dialogs.
 		- Introducing keyboard shortcut for the Open Pattern dialog.
 		- Allow for opening more than one Pattern at once.
+		- All available audio drivers can now be chosen via CLI.
 	* Interface
 		- Improved scalability (most PNG images were replaced by SVGs,
 		  hardcoded PNG labels are now directly drawn by Qt, and spin boxes,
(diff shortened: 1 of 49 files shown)


theGreatWhiteShark:
@cme I remember we decided that `Song::ActionMode::selectMode` should be selected by default in SongEditorPanel during startup . (Or do I mix things up?) The way it is implemented right now is that regardless of what the user stored in her .h2song file `selectMode` will be selected. I would suggest we respect the user's choice in v1.2 (see e8aa205) as there was enough time for the users between releases to get accustomed to the new interaction mode.

PR #1416, opened 2021-12-05
Devel warning automation

cme:
Show development build warning for any build from an untagged git commit.


theGreatWhiteShark:
Nice!

Instead of reenabling the `showDevelWarning` variable we could also keep track of the Hydrogen version used by the user in the Preferences. If a release version > the last one is started, it could show the ChangeLog instead of the devel warning.

PR #1428, opened 2021-12-22
Validate temp filename template

cme:
Make sure that filename templates will yield valid filenames. This is an issue since pattern names are used as basenames when modifying sequences, and these are not necessarily valid filenames.


theGreatWhiteShark:
Well, guess what happens when renaming a drumkit to `../../../something`. :smile: 

I'll have a look at if. Maybe this is a good opportunity to fix #793 along the way.

PR #1429, opened 2021-12-24
Add linker flags for backtrace/execinfo if needed.

rrrapha:
This fixes a linker error on *BSD where the _backtrace_-functions are located in `libexecinfo.so`.

(See also https://cmake.org/cmake/help/latest/module/FindBacktrace.html)


cme:
Thanks!

PR #1434, opened 2021-12-31
Merge tempfile and backtrace changes to master

cme:


src/core/Helpers/Filesystem.cpp
```diff
@@ -558,12 +558,16 @@ QString Filesystem::tmp_dir()
 {
 	return QDir::tempPath() + "/" + TMP;
 }
-QString Filesystem::tmp_file_path( const QString& base )
+QString Filesystem::tmp_file_path( const QString &base )
 {
-	QFileInfo f( base );
-	QString templateName(tmp_dir() + "/");
+	// Ensure template base will produce a valid filename
+	QString validBase = base.left( 20 );
```
theGreatWhiteShark, line 564: Is there a specific reason you only use the first 20 characters? For larger names the suffix will be cut. E. g the test file for the velocity automation is now stored as `velocityautomation-nzemKQ.w` instead of `velocityautomation-nzemKQ.wav`

cme: The purpose was to make sure that filenames stayed valid in case of underlying filesystem length constraints since pattern name lengths are unbounded. It would have been better to truncate the `completeBaseName`. I'll change.

theGreatWhiteShark: Thanks for the quick fix!

Issue #680, opened 2018-12-14
Playback track is affected by tempo changes


NihiliJay:
Hydrogen version *          : Latest
Operating system + version  :  Windows 10
Audio driver + version      : 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Change playback track to ignore tempo changes.
After importing click/bed track as a wav file, tempo changes also shifts the playback track.

theGreatWhiteShark:
Which audio driver do you use?

NihiliJay:
Just the onboard Realtek HD
I have a somewhat dated computer Asus p6x58d-e

theGreatWhiteShark:
In the drop-down menu at the very top of _Preferences > Audio System_ you can select different audio drivers Hydrogen can use to talk to your audio device. Do these jumps in the playback track occur using all of them?

By the way: Did I get you right that you import a click track to play along with Hydrogen and still use its internal audio engine to perform tempo changes? Why don't you use the the Metronome of Hydrogen instead of a click track?

NihiliJay:
I've used guitar pro and tabbed out my songs including a click track line. Exported that as a wav. in hopes to create the drum tracks then bring that into a DAW to then properly record.
Many of the songs have varied tempo changes and all was working well till i noticed a glitch when the tempo changed. I finally realized that the tempo change seems to either skip the bed track or it speeds it up much like if you bumped a record player while its playing.
I'll try what you suggested but i've just finished re-installing my system with a SSD upgrade so give me a bit.

I appreciate the help.

theGreatWhiteShark:
Hydrogen is written to work with external audio software synchronizing the playback of various programs, like the JACK server. If such an external software informs Hydrogen about a tempo change, the playback speed of will be altered and playback position will be relocated. E.g. you change the speed from 120bpm to 240bpm after one minute. Right before the tempo change 120 beats were passed and the "temporal position" is at one minute. After changing the speed the current position of 120 beats would correspond to a "temporal position" of half a minute. To avoid this problem, Hydrogen keeps an internal representation of the transport position and some logic in the audio engine to smooth out these tempo changes. But maybe something is implemented wrong concerning the playback track in the audio driver of yours. (I can't verify it since I'm no Windows user).

[...]

Issue #103, opened 2013-06-18
incorrect position of song cursor with variable BPM


lunadog:
In a song with multiple BPM changes, when the song is stopped, clicking on the timeline will position the cursor correctly, but when play is then pressed, the song starts in the wrong position.

I can't attach patches here, so I will add it to the dev. forum.

thijz:
just retested with todays latest github code > issue is still there
if you remove the BPM markers all is ok again

lunadog:
Is it worth just implementing the solution I posted?

thijz:
looks like i missed the link to the patch :-S
i'll try to apply the patch and let you know the outcome
grtz
Thijs

theGreatWhiteShark:
I can confirm this issue and already fixed it for the Jack audio driver (which handles the Timeline differently).

@trebmuh can you assign me for this issue?

trebmuh:
@theGreatWhiteShark done

theGreatWhiteShark:
Fixed with #1435

PR #1435, opened 2022-01-01
Rework audio engine

theGreatWhiteShark:
Rewrite AudioEngine class to handle arbitrary tempo changes.

This one was really a pain. I tried to keep the changes as concise as possible but in the end I saw no other option than rewrite a lot of those you-better-don't-change-this-running-system part of the AudioEngine.

## Problem

Transport in Hydrogen is based on ticks and the corresponding frames are just kept as a convenience variable used during the audio rendering. `TransportInfo::m_nFrames` were thus the number of frames passed since the beginning of the song *given a particular tick size*. If there is a change in tempo, the tick size and, as a result, the frames do change. With the Timeline deactivated this is not an issue at all. But as soon as it is enabled and there are tempo markers present things just glitch. There were jumps in the transport position (#103), the playback track was affected by the tempo changes (#680), and various little oddities in humanization of note placement and resampling using Rubberband no one reported yet happened for notes close to the tempo marker.

## Solution

I solved these by making `TransportInfo::m_nFrames` the number of frames since the beginning of the song and thus decouple them from `TransportInfo::m_fTick`. The methods `computeFrameFromTick()` and `computeTickFromFrame()` have to be used for conversion. In addition, I converted the type for the ticks from `int` to `double`. This, unfortunately, was necessary to ensure there is no loss in information when converting between frames and ticks (since we support float-precision tempo). But the new type is only used within the `AudioEngine` and rounded to `long` by the public getter function. All other classes (apart from the `Sampler`) are not affected.

 As a byproduct
- Hydrogen is now fully compatible with frame-based transport servers/drivers. (JACK implementation did slim down and potential new ones, like Ableton Link, should be way more easy to do)
- Hydrogen can easily support arbitrary tempo changes, like ramps or even random walks. The only thing that needs to be adopted are the two functions `computeFrameFromTick()` and `computeTickFromFrame()`
- The `AudioEngine` almost supports float prevision note placement (required for e.g. true tuplet support). What's left to do is described in 91ea6a564e5f1565d1df4424694a41a79b0f3834
- The core part of the transport handling is now covered with unit tests.

## Notes

I changed the JACK timebase support to only use the tempo provided by the JACK server and use the plain frame information instead the bar, beat, tick one for relocation (this wasn't possible with the previous implementation of `TransportInfo::m_nFrames`). Hydrogen disregards the time signature information provided by the JACK server. Otherwise, this would allow external application to change the length of patterns (See https://github.com/hydrogen-music/hydrogen/discussions/1433) . If there are still problems in synchronization I would suggest to deprecate the timebase support. It was the cause of quite a number of sync problems during the years and provides only minor advantages.

I'm still in the process of doing thorough testing of everything.  While doing so I also encountered a number of bugs I fixed on the way
- the `DiskWriterDriver` stopped exporting at the last tick + a small lookahead in frames. I am not quite sure how this was not spotted or reported earlier but as a result H2 did chop off every audio signal when reaching the end of the song. For short samples, like toms, placed a couple of ticks before the end this might be okay. But for longer ones, like crash, this is just bad.
- there was a small glitch in the note playback when the end of a song was reached and loop mode was not enabled
- adding a pattern after the end of the song when the transport was already looped at least once cause the transport position to glitch
- disabling the Loop button when transport was already looped at least once caused the transport to stop immediately. Now, the song will still be played till the end.


theGreatWhiteShark:
@cme I'm hitting the wall again with this damn JACK timebase support. I think it's best to just remove it. What do you think?

### short explanation:
The way Hydrogen is designed is just not compatible with the concept of JACK BBT.

### long one:
With BBT support enabled  one JACK client - which previously registered itself as timebase master -  sends some [additional metadata](https://jackaudio.org/api/structjack__position__t.html). Next to the transport position in frames, the position is also provided as bar/beat/tick, current speed, current measure and everything required to determine the onset of the last bar in frames. If there is a changes in tempo and/or measure, the frame position alone is not enough anymore since the metadata do not hint where and how many changes there are. Sync is now depending on BBT.

But our definition of a bar (at least what we are suggesting with the SongEditor) is the longest pattern in a particular pattern group.  Adding yet another SongEditorRuler keeping track of the (JACK) bars seems to me quite an overkill (although a number of users already ask for such a thing). With a hidden version of such a ruler it would be in theory possible to keep the transport sync but only if a) either the JACK server is allowed to alter pattern lengths (according to the measure information provided) or  b) bars could be at arbitrary positions within patterns. Both scenarios aren't favorable to say the least. Similar to the current implementation we could also have some half-baked solution in which the user has to have prior knowledge about its usage and has to keep specific things consistent while glitches can't be ruled out entirely.

Instead, I would remove all the JACK master stuff as well as the additional layer of the tempo design listening for external tempo and overwriting the Timeline bpm in one commit.

### current solution:
Right now the user can decide in the Preferences to 
1. enable or disable the whole BBT support
2. promise to maintain only one type of measure in the timebase master
3. promise to keep measures in the timebase master and pattern lengths in Hydrogen consistent.

Both promises in 2. and 3.  must be kept by the user herself and without a look in the documentation most people probably won't get what's going on here.

So, how other applications solve this? By the time I wrote the current solution most other applications (DAWs) only registered themselves as master and sent those extra bits of information. If there was an external master, they just looked the other way and didn't use the data....

I really dislike the current solution as it is too complex and does not feel like we are actually supporting something. In hindsight I got myself to implement this one because I didn't want to drop (a bricked) feature and I'm a usually keep the 2. promise. But the cleanest and most intuitive solution would be to get rid off the whole thing.

### consequences of removal:
Users need to keep pattern lengths/measures and tempo changes consistent manually. That's a little (expected) extra work but there is no pain because the synchronization feature is only working partially or one has to deal with glitches.

Those user's actually using the current version of the support have to add tempo markers at the same positions as the ones in the JACK master application. It's a little inconvenient but there is no loss in information. Since  "J.MASTER" button will also disappear, version 1.2 will hint that the support is gone.

cme:
I'm going to have to go and do some background reading here to answer this!

cme:
My vague high-level thoughts are that you're right, there's no general mapping between bars and beats and pattern that works in all cases, but I think the cases where the user keeps things consistent are a useful and maybe "obvious" sub-set. So maybe the more useful approach to take is to help the users keep promises [2] and [3] by providing warnings if the timebase master's measures don't match, or if they make changes in Hydrogen that would cause a mismatch?

theGreatWhiteShark:
> So maybe the more useful approach to take is to help the users keep promises [2] and [3] by providing warnings if the timebase master's measures don't match, or if they make changes in Hydrogen that would cause a mismatch?

Problem here is a lack of information. The JACK server just provides tempo and measure for the current transport position. But to keep things in sync or to give useful feedback via the GUI we would require the position of all changes in measure and tempo. It's a bit of a mess.

Another thing: The JACK TBM API does not even specify where the change did occurred in the provided buffer. This was the part is was working on last. I had both Hydrogen and Muse in sync. Depending on whether H2 or Muse did register as TBM, the resulting audio of H2 did differ. This is because, e.g. for a buffersize of 4096 the tempo change did occurred midway. With Muse as master it handled the first half of the frames as ticks with tempo X and the second half with tempo Y. But since the API does not tell where the change happened, I ended up adding some code for whether there was a bar nearby as the tempo change might have happened there. But when thinking about it again, this turned out to be just my personal preference for setting markers. Also, other applications support gradual changes in tempo.

The whole thing just feels like a good idea but way to premature/unsuited for our code. I mean, we can of course keep it (and maybe hide it by disabling the TBM support per default). It just bugs me a lot that this part of the code - which is the one I first contributed to and probably spent the most time with - is not working properly and that there is not so much I can do about it :smile:

theGreatWhiteShark:
I included the previous BBT support but deactivated it per default.

Also, now it's possible to change pattern length while playing or toggle patterns without any glitches.

PR #1436, opened 2022-01-03
Add linker flags for backtrace/execinfo

rrrapha:
This is related to #1429.
Since commit 27ab1d11bee3d6108d0eefb833a68dabab516097, the gui needs to be linked against libexecinfo too (on *BSD).


cme:
Excellent, thank you! I hope I didn't disrupt too much for you with that commit.

PR #1439, opened 2022-01-05
Fix crash on quitting after inserting a pattern.

cme:
The redo method for inserting a pattern failed to take a new copy
of the pattern when inserting into the song, leading to double
delete of the pattern on finalisation.


cme:
Now that I come to think of it, this would also cause a crash when trying to re-do an undone insert of a pattern.

theGreatWhiteShark:
Strange I didn't catch this while testing. Sorry.

cme:
> Strange I didn't catch this while testing. Sorry.

Nae bother :)

PR #1440, opened 2022-01-07
Fix compiler warnings & smaller code issues

mauser:
This pull request contains some fixes for issues which came by using clang/clang-tidy:
- off-by-one error in Hydrogen.cpp: 
   `m_nInstrumentLookupTable[ MAX_INSTRUMENTS ]` -> `m_nInstrumentLookupTable[ MAX_INSTRUMENTS - 1]`

- "," instead of "." used as (or at least was intended to be used as) a floating point number delimiter in Fader.cpp 
  `fFaderWidth = 6,8;` -> `fFaderWidth = 6.8;` 

- "&&" erroneously used for bitmask comparison
 `(ev->buttons() && Qt::LeftButton)` -> `(ev->buttons() & Qt::LeftButton)`

### DIFF
--- a/src/core/IO/PortAudioDriver.cpp
+++ b/src/core/IO/PortAudioDriver.cpp
@@ -177,7 +177,7 @@ int PortAudioDriver::connect()
 			memset( &outputParameters, '\0', sizeof( outputParameters ) );
 			outputParameters.channelCount = 2;
 			outputParameters.device = nDevice;
-			outputParameters.hostApiSpecificStreamInfo = NULL;
+			outputParameters.hostApiSpecificStreamInfo = nullptr;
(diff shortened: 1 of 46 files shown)


src/core/IO/PortAudioDriver.cpp
```diff
@@ -177,7 +177,7 @@ int PortAudioDriver::connect()
 			memset( &outputParameters, '\0', sizeof( outputParameters ) );
 			outputParameters.channelCount = 2;
 			outputParameters.device = nDevice;
-			outputParameters.hostApiSpecificStreamInfo = NULL;
```
cme, line 180: Some old habits die harder than others... :D

PR #1443, opened 2022-01-08
Add some tooltips for things that felt like they could use them

cme:


src/gui/src/PreferencesDialog/PreferencesDialog_UI.ui
```diff
@@ -29,7 +29,7 @@
       <enum>QTabWidget::Rounded</enum>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
theGreatWhiteShark, line 32: We need a unit test for this one :D

src/gui/src/PreferencesDialog/PreferencesDialog_UI.ui
```diff
@@ -342,6 +367,9 @@
                <height>22</height>
               </size>
              </property>
+             <property name="toolTip">
+              <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Audio buffer size (in frames). This generally determines the &lt;span style=&quot; font-weight:700;&quot;&gt;minimum&lt;/span&gt; audio latency the system can achieve. Actual audio latency will be greater.&lt;/p&gt;&lt;p&gt;Selecting a buffer size that is too small may lead to audible glitches, particularly if the system becomes more heavily loaded.&lt;/p&gt;&lt;p&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Use with caution.&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
theGreatWhiteShark, line 371: There is one potential problem with HTML in the translatable text. When using a standard editor to do the translation all the HTML code is displayed in plain text (not sure whether there are more elaborated ones out there). If the translator does not know HTML she might end up translating things like "body" or "style". Not quite sure if that ever happened but at least ones I encountered mismatching brackets in a translation which made Qt unable to render them

Issue #742, opened 2019-07-13
LADSPA on mac not working


alkurop:
Hydrogen version *          : 1.0.0-beta1
Operating system + version  : Mac OS Mojave
Audio driver + version      : Jack

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

I want to enable effects in the mixer.
But Hydrogen shows error message "LADSPA effects not available".
What should I do to turn them on?

mauser:
Hey,

most probably ladspa was not available when you compiled Hydrogen.

theGreatWhiteShark:
How did you install the LADSPA plugins?

On LInux you can find them bundled in dedicated package but on Mac I couldn't find anything using `brew` and `port`. I'm not too much into LADSPA and never used my Mac at work for something serious but (apart from some FFTW and maybe LAPACK packages) I do not see why one shouldn't be able to use LADSPA plugins on Mac right away. But you might need to compile and install them on your own.

For what it's worth: `Audacity`, which is available using `port`, does provide both header and source files for the LADSPA API but does not install any plugins for you.

mauser:
Hi! 

@theGreatWhiteShark : The LADSPA Plugins which are installed on the machine are not the problem. The problem is that his hydrogen version has been compiled without LADSPA support (technically: `H2CORE_HAVE_LADSPA`  is not defined). This happens if during compile time the ladspa header cannot be found.

theGreatWhiteShark:
@mauser did LADSPA on Mac ever worked? I struggle to find any hints in the net except for programs shipping their own `ladspa.h` (shall we do the same and only use it with macOS?).

At least on the macOS laptop I have access to I do not find any alternative to the `noise` library in Linux. Sure, one can get the source code of various LADSPA plugins and compile and install them on macOS. But since there is no LADSPA SDK, `FindLadspa.cmake` is not able to list and include them. 

I'm also surprised to find that one is able to install a version of Hydrogen via `sudo port hydrogen +lrdf`, which compiles against `liblrdf`. But since the `H2CORE_HAVE_LRDF` are wrapped with `H2CORE_HAVE_LADSPA`, which is unset, this variant has no addiional features at all.  But the corresponding [Portfile](https://github.com/macports/macports-ports/blob/master/audio/hydrogen/Portfile) is easy to understand and it should be no problem fixing it

theGreatWhiteShark:
[...]

PR #1453, opened 2022-01-11
Revisit Windows LADSPA support

theGreatWhiteShark:
I just found time (and a computer) to have a look at the LADSPA support on Windows.

The situation looks quite similar to the one on [Mac OS X](https://github.com/hydrogen-music/hydrogen/issues/742). In principle Hydrogen does support LADSPA plugins on both platforms. But - and that's a big but - there are **none**. Some time ago one could at least use the ones that were shipped with Audacity but they seem to use another plugin format nowadays.

On Windows both the older Audacity LADSPA plugins and the ones shipped with LMMS do not work with Hydrogen. I don't know why. When inspecting them with `winedump` on Linux it looks like they are exposing the right symbol. Maybe the compile options [expected by Qt](https://doc.qt.io/qt-5/qlibrary.html#resolve) were not provided.

On **Mac OS X** it's more or less straight forward to get the source code of LADSPA plugin bundles and compile them. It's not ideal but I would still feel comfortable with describing the process in the docs/wiki and to leave it to the user. But on **Windows**  that's not straight forward at all. I just found a single LADSPA bundle I was able to compile using MSYS2 without making too many changes to project's Makefile/CMakeLists.txt. I would suggest to either ship some plugins or to remove the FX support incl. all buttons on Windows for now.

With this proof-of-concept draft I try to compile the (GPL2 licensed) [CMT LADSPA bundle](https://www.ladspa.org/cmt/) in AppVeyor and ship it with Hydrogen on Windows platforms. A unit test was added to verify the plugins are valid and can be imported.

@cme , @mauser what do you think about the idea in general?

P.S. This code might be quite dirty as I only have a very superficial knowledge of AppVeyor and build systems in general and only use Windows for two days. The plugin library built using AppVeyor is still invalid and can not be loaded yet, But locally using the same commands everything works. If we decide to proceed in this direction, I will have a look at it.

Associated artifacts can be found [here](https://ci.appveyor.com/project/mauser/hydrogen/builds/42155299/job/ct1wet4x59lmu3ck/artifacts)


cme:
I haven't really looked into the plugin support properly but keep meaning to.

I had a few thoughts:
  1. We should probably provide a GUI for finding plugins, rather than relying on environment variables. Either to just set paths to something external, or to copy to the user plugin directory. 
  2. If we include plugins from another project in releases, we should probably fork the project and include as a submodule in hydrogen to make pulling from upstream easier
  3. Being able to test the plugin infrastructure is great! When I was debugging the audio processing for plugins with variable-size buffers, I really wanted to be able to create a unit test that used a plugin. Perhaps always including a fairly trivial plugin (say, a delay line) in the Hydrogen codebase would be useful for this, and for maintaining the build system and plugin finding in a testable way.

theGreatWhiteShark:
> 1. We should probably provide a GUI for finding plugins, rather than relying on environment variables. Either to just set paths to something external, or to copy to the user plugin directory.

For LADSPA plugins alone I would consider this an overkill. AFAIK all projects excepts [LSP](https://github.com/sadko4u/lsp-plugins) are either discontinued or still maintained but not actively developed anymore. The usual way of getting them is through the package manager and this one is putting them in a proper place.

If we intend to support other format too one day this would be indeed a good idea. But I don't know if we should. When looking at ChangeLogs and issues in Linux DAWs I get the impression supporting LV2 and VST is quite a complex task and requires lots of work and fixes.

> 2. If we include plugins from another project in releases, we should probably fork the project and include as a submodule in hydrogen to make pulling from upstream easier
> 3. Being able to test the plugin infrastructure is great! When I was debugging the audio processing for plugins with variable-size buffers, I really wanted to be able to create a unit test that used a plugin. Perhaps always including a fairly trivial plugin (say, a delay line) in the Hydrogen codebase would be useful for this, and for maintaining the build system and plugin finding in a testable way.

Sure. It's just a proof of concept (and a quick idea). Integrating them into our code base properly is way better than just  a `make` done in AppVeyor. Also, why not shipping them with (Linux and) Mac OS X too?

theGreatWhiteShark:
When rethinking the whole Windows plugin subject I doubt this approach will be a proper solution.

It won't mitigate the "where are my plugins?" questions of VST users and only offers a very limited and basic set of LADSPA plugins. As the latter standard is more or less dead on Windows, the feature will also become more and more obscure as time goes by.

I doubt these plugins will be used frequently. More likely effects will be applied in external applications on the audio rendered by Hydrogen. It feels like the burden of maintaining the CMT sources and their integration into the Windows build is not worth it. A proper support of VST plugins would be much more useful.

PR #1454, opened 2022-01-12
Updating default config file + fixing unit tests

theGreatWhiteShark:
some fields were added/removed and some were out of date

Most notably the `exportDialogSampleRate` was set to 44100. But the value is used in ExportSongDialog.cpp as the index of the sample rate combo box. This results in an empty sample rate box on a fresh installed H2 instance/after removing the hydrogen.conf file and a bricked audio export (unless the user manually chooses a value)


theGreatWhiteShark:
@cme I just realized we missed updating the default conf on all past releases. Not a huge problem but may cause spurious errors during the first startup the user is not able to reproduce.

PR #1456, opened 2022-01-13
use dynamic_cast to tell audio drivers apart

theGreatWhiteShark:
The previous comparison based on the class name was flawed (see comments in https://github.com/hydrogen-music/hydrogen/commit/ebc9a68ab3cfaa6ec1f84a97bab351d83a218665)

### DIFF
--- a/src/core/AudioEngine/AudioEngine.cpp
+++ b/src/core/AudioEngine/AudioEngine.cpp
@@ -494,61 +494,38 @@ AudioOutput* AudioEngine::createDriver( const QString& sDriver )
 
 	if ( sDriver == "OSS" ) {
 		pDriver = new OssDriver( m_AudioProcessCallback );
-		if ( pDriver->class_name() == NullDriver::_class_name() ) {
-			delete pDriver;
-			pDriver = nullptr;
-		}
 	} else if ( sDriver == "JACK" ) {
 		pDriver = new JackAudioDriver( m_AudioProcessCallback );
-		if ( pDriver->class_name() == NullDriver::_class_name() ) {
-			delete pDriver;
-			pDriver = nullptr;
-		} else {
 #ifdef H2CORE_HAVE_JACK
+		if ( dynamic_cast<NullDriver*>(pDriver) == nullptr ) {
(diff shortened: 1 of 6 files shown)


src/core/AudioEngine/AudioEngine.cpp
```diff
@@ -494,61 +494,38 @@ AudioOutput* AudioEngine::createDriver( const QString& sDriver )
...
 		pDriver = new JackAudioDriver( m_AudioProcessCallback );
-		if ( pDriver->class_name() == NullDriver::_class_name() ) {
-			delete pDriver;
-			pDriver = nullptr;
-		} else {
 #ifdef H2CORE_HAVE_JACK
+		if ( dynamic_cast<NullDriver*>(pDriver) == nullptr ) {
 			static_cast<JackAudioDriver*>(pDriver)->setConnectDefaults(
 						Preferences::get_instance()->m_bJackConnectDefaults
 						);
```
cme, line 503: Would seem to be more natural to write this as:
```c++
if ( auto pJackDriver = dynamic_cast<JackAudioDriver*>( pDriver ) ) {
    pJackDriver->setConnectDefaults(
        Preferences::get_instance()->m_bJackConnectDefaults
    );
}
```
since the conditional code is more conditional on it being an actual `JackAudioDriver*` than just being non-null?

src/core/AudioEngine/AudioEngine.cpp
```diff
@@ -682,7 +659,7 @@ void AudioEngine::setAudioDriver( AudioOutput* pAudioDriver ) {
 	this->unlock();
 	
 	if ( m_pAudioDriver != nullptr &&
-		 m_pAudioDriver->class_name() != DiskWriterDriver::_class_name() ) {
+		 dynamic_cast<DiskWriterDriver*>(m_pAudioDriver) == nullptr ) {
```
cme, line 662: Maybe it would be better to make the DDWD trivially `connect()` to avoid this? Or maybe having an `isDisconnectedDriver()` method that answers `true` for the DDWD (and FakeDriver?)

PR #1461, opened 2022-01-19
Performance improvements

cme:
I'd noticed some increase in CPU usage in playback and the GUI when moving to master. Some profiling on an older machine found some small easy tweaks that help a lot.

A quick check on the raspi 4 shows that these changes reduce CPU usage playing Diddley with the mixer visible from ~80% to ~25% of a core.

To do:
  - [x] add some benchmarks for consistent result tracking
  - [x] use a pixmap cache for pattern editor ruler
  - ~~[ ] refactor ADSR application~~
  - ~~[ ] refactor renderNoteResample~~
  - ~~[ ] instrument output setting with batches~~


theGreatWhiteShark:
> Commonise and loop-hoist creation of transient shared_ptrs to
Instrument and ADSR objects, as the construction and destruction
of the shared_ptrs is expensive.
>
> This change reduces the time taken for audioEngine_process playing
GM_kit_Diddley on my 2008 MacPro from 12% of a core, to only 3.2%.

Wow. I thought the time required audio processing drafts everything else.

Considering this overhead, would changing more base classes, like `Note` or `Pattern`, to use shared pointers actually be a good idea?  

> Don't update the mixer when it's not visible.

Nice one! This should probably apply for the director too. But now the activation of the Mixer should do an initial update too, right?

There is another source for bad performance I came across during updating the GUI. Some parts, like the PlayerControl, do not use the EventQueue to trigger updates but, instead, use a QTimer and update all controls every Xms.  Also, if I recall correctly, the meters in the Mixer are constantly updated. This could be also changed to be triggered by the AudioEngine and to run till transport has stopped and the Sampler is done processing.

cme:
> > This change reduces the time taken for audioEngine_process playing
> > GM_kit_Diddley on my 2008 MacPro from 12% of a core, to only 3.2%.
> 
> Wow. I thought the time required audio processing drafts everything else.

In this case there wasn't much polyphony, no resampling and no plugins, so it should be fairly minimal overhead. Other songs could have vastly different requirements.

Worth bearing in mind that some non-trivial proportion of GUI redraw CPU time will occur in the X server so won't show up on profiles etc.

> Considering this overhead, would changing more base classes, like `Note` or `Pattern`, to use shared pointers actually be a good idea?

I don't think this in itself is a reason to avoid them -- this had significant impact because it created and destroyed shared_ptrs for every frame played, I don't think there's anything that would *need* to access individual objects as frequently as that. A reasonable approach would be to continue as normal, and profile to find any that actually cause a performance problem and mitigate those. Shared ownership can be pushed out of loops or up the call tree and raw pointers used on loan. This seems to be [recommended practice](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-smart) but it wouldn't hurt to formalise it with some sort of 'loan' template.

> > Don't update the mixer when it's not visible.
> 
> Nice one! This should probably apply for the director too. But now the activation of the Mixer should do an initial update too, right?

It does. The `showEvent` method for the Mixer calls `updateMixer` (at which point `isVisible()` returns `true` and the update should happen).


> There is another source for bad performance I came across during updating the GUI. Some parts, like the PlayerControl, do not use the EventQueue to trigger updates but, instead, use a QTimer and update all controls every Xms.

Making these event-driven would be good, but it might be a smaller simpler change to make sure updates are lazy and don't update() anything if the relevant state hasn't changed -- checks are usually very cheap.

> Also, if I recall correctly, the meters in the Mixer are constantly updated. This could be also changed to be triggered by the AudioEngine and to run till transport has stopped and the Sampler is done processing.

theGreatWhiteShark:
> but it might be a smaller simpler change to make sure updates are lazy and don't update() anything if the relevant state hasn't changed -- checks are usually very cheap.

I'll do that in a separate PR. I'm just working on the autosave stuff and noticed none of the widgets in the instrument/layer editor mark the song as modified. But when integrating this into the widgets themselves, I noticed that loads of them seem to be set via Hydrogen is idle.

cme:
It *looks* like Qt makes updates on some widgets lazy already. The instrument editor doesn't show up on the profile until fairly far down, and from this one (no mixer window shown):

[sample.txt](https://github.com/hydrogen-music/hydrogen/files/7940020/sample.txt)

the largest (with only 2 samples out of ~9k) is a QLabel::setText which actually gets hit in QString::operator==() which sounds like it's checking at least the text and not actually doing any screen update. Making it lazier in the InstrumentLine would remove that and the handful of other samples though :)

With the mixer visible, the profile of the GUI is more heavily skewed towards the mixer:

[sample.txt](https://github.com/hydrogen-music/hydrogen/files/7940120/sample.txt)

I haven't looked at the code but from prominent appearance of `QBezier::addToPolygon` I would guess that there's a lot of it being redrawn when only the VU bars need to change most of the time?

cme:
As an aside, I haven't yet found any reasonably reliable, performant, non-intrusive statistical sampling profiler that just works on Linux. Which seems pretty bizarre to me. `google-pprof` is about the closest I've found to `sample` on macOS, but it seems to miss a lot, probably because using SIGPROF biases towards syscalls due to being checked on returns from kernel mode. :(

theGreatWhiteShark:
> It looks like Qt makes updates on some widgets lazy already. The instrument editor doesn't show up on the profile until fairly far down, and from this one (no mixer window shown):

Nice. But it's probably what should be expected as it is such a low hanging fruit.

> I haven't looked at the code but from prominent appearance of QBezier::addToPolygon I would guess that there's a lot of it being redrawn when only the VU bars need to change most of the time?

Hmm. Good question. I think revisiting the custom drawing routines with respect to performance is a good idea (especially since the new patch of AudioEngine is bound to make it less efficient :grimacing: . But on a quick profiling using the AudioEngine::m_processTime I only saw minor differences for very small buffer sizes and I haven't added caching/optimization yet). When porting the GUI to SVG I choose to draw a number of elements instead of loading them from file. But I assumed all  corresponding `update()` calls where required and there are no redundant ones. That might be worth checking.

Another thing that just comes to my mind: the VU bars of _all_ MixerLines are updated regardless of how many of them are visible in the Mixer (selected via the scrollbar).

theGreatWhiteShark:
> As an aside, I haven't yet found any reasonably reliable, performant, non-intrusive statistical sampling profiler that just works on Linux. Which seems pretty bizarre to me. google-pprof is about the closest I've found to sample on macOS, but it seems to miss a lot, probably because using SIGPROF biases towards syscalls due to being checked on returns from kernel mode. :(

I haven't done profiling in C++ yet. Qt [lists](https://wiki.qt.io/Profiling_and_Memory_Checking_Tools) a couple of different ones. But they seem to be quite old and memory-centered.

cme:
Well the test failures in this after merging from master have certainly filled my day with head-scratching.

  1. The test framework doesn't correctly deal with the expected nSamplePosition when a sample ends. It always expects nSamplePosition to move by the buffer size. When checking a buffer that has an end of sample, the value is actually clamped to the sample length.

  2. In the tests on master this never actually happens.

  3. This branch coincidentally changed the order of picking random numbers. This causes the velocity humanisation to end up picking a different layer sample for some notes, one of which ends in a checked buffer period, triggering unit test failure. 

It took a lot of tracing to realise this was what was going on! :D

I will 

  1. Fix the test check and
  2. Restore the random initialisations to the original order.

theGreatWhiteShark:
> Well the test failures in this after merging from master have certainly filled my day with head-scratching.

Yeah, I know. Sometimes it needs a whole day to find the source of the error and then it turns out to be within the tests itself, some wrong settings in the shipped config files or similar. I hope these tests pay out in the long run.

> The test framework doesn't correctly deal with the expected nSamplePosition when a sample ends. It always expects nSamplePosition to move by the buffer size. When checking a buffer that has an end of sample, the value is actually clamped to the sample length.

Sorry for that. You can always ping me to look into the failing tests I introduced.

> This branch coincidentally changed the order of picking random numbers. This causes the velocity humanisation to end up picking a different layer sample for some notes, one of which ends in a checked buffer period, triggering unit test failure.

In case you wondered why velocity humanisation is activated in a unit test the begin with: I used it to "uniquely" determine a note. Since this tests checks for sudden jumps and other discontinuities in the transport position, just comparing the keys of the notes was not enough.

cme:
> > Well the test failures in this after merging from master have certainly filled my day with head-scratching.
> 
> Yeah, I know. Sometimes it needs a whole day to find the source of the error and then it turns out to be within the tests itself, some wrong settings in the shipped config files or similar. I hope these tests pay out in the long run.

I'm sure they will! Sorry, I hope my tone came across as "bemused" rather than annoyed! :)


> > The test framework doesn't correctly deal with the expected nSamplePosition when a sample ends. It always expects nSamplePosition to move by the buffer size. When checking a buffer that has an end of sample, the value is actually clamped to the sample length.
> 
> Sorry for that. You can always ping me to look into the failing tests I introduced.

Cool, thanks. In this case though I was sure until the very end that it was definitely a problem I'd introduced :D

> In case you wondered why velocity humanisation is activated in a unit test the begin with: I used it to "uniquely" determine a note. Since this tests checks for sudden jumps and other discontinuities in the transport position, just comparing the keys of the notes was not enough.

Yeah, I see that, and it makes sense! And random tests do have a value all to their own, which makes me wonder -- maybe trying different random seeds in a different test mode would be useful?

theGreatWhiteShark:
> I'm sure they will! Sorry, I hope my tone came across as "bemused" rather than annoyed! :)

No, don't worry. But during the fix of the audio engine is was seriously annoyed :D

> Yeah, I see that, and it makes sense! And random tests do have a value all to their own, which makes me wonder -- maybe trying different random seeds in a different test mode would be useful?

Sure. I was quite surprised how easily things failed when writing the AudioEngine unit tests. Some small jitters, like a different sample rate, and things just blew up :facepalm: . Maybe we could add a `--long` parameter for the tests which is not set per default and tries a larger number of parameter combination and (maybe even random) random seeds.

PR #1471, opened 2022-01-30
add unit test for SSL support to 1.1 release

theGreatWhiteShark:
Since Qt does not ship its own OpenSSL library, we provide the required .dll in the Windows release for Hydrogen >1.1.1 .

A member of the Qt team stated in a forum that one has to ensure the exact version of OpenSSL is shipped that was also used to compile Qt5. As they might update the used version in future Qt releases we have to check whether the lib provided by us is the correct one.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,6 +19,10 @@ environment:
       CMAKE_FLAGS: "-DWANT_DEBUG:BOOL=OFF -DWIN64:BOOL=ON"
       appveyor_build_worker_image: Visual Studio 2019
       LIBJACK: libjack64.dll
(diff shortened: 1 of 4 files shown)


theGreatWhiteShark:
Since the SSL unit test didn't failed without the OpenSSL library being explicitly added it seems it's already included in the `PATH` of AppVeyor's Windows image. But this shouldn't be a problem. The test checks whether the right version of SSL is found within Hydrogen and since our manually provided would should have higher priority than the systems one, it still should do it's job (important for `master` and not `releases/1.1`).

P.S. Using this PR it's possible again to download drumkits from SourceForge (which requires HTTPS).

PR #1478, opened 2022-02-04
SongEditor: checks for drawing mode

theGreatWhiteShark:
If the mouse is moved in the song editor in drawing mode, the `mouseMoveEvent` does now check whether the mouse did enter a different grid cell and if it did not exceed the pattern list to the bottom before firing.


cme (approved):
Nice catch

PR #1488, opened 2022-02-14
Fix 'set automatic velocity' during Sample import + InstrumentEditor UI and UX tweaks

### LINKED ISSUE #841: ‘Set automatic velocity’ for layers works in reverse.
Hydrogen version *          : Hydrogen 0.9.7 & Hydrogen 1.0.0-Beta2
Operating system + version  : Linux Mint 19.3 MATE
Audio driver + version      : Behringer UMC202HD USB Audio Interface

_I did not find another issue mentioning this, so hopefully it’s not a duplicate issue. The problem exists in both Hydrogen 0.9.7 and 1.0.0-Beta2. Unless I've totally missed something, I'm pretty sure this is a bug._

When working with multiple layers in a single instrument, choosing ‘**Set automatic velocity**’ in the load layer import window works in reverse of the intended way.

Typical ways of adding instruments layers would be loading layers from the top down, where layer 1 = lowest velocity range. 

In the image below, note that the 16th layer contains the 16th snare wav file (which is intended to be the loudest snare hit), but is auto-mapped to the lowest velocity range.

hydro-problem

The issue with setting automatic velocity happens whether using ctrl-click to select multiple layers to batch import into the instrument layers window, or when adding each layer individually.

Ideally, batch import of sound files into layers would be the preferred way, but this is unusable when selecting ‘set automatic velocity’ because of the reverse-order velocity mapping that occurs.

My solution, which works, but is not ideal, is loading files into each layer in reverse order. So, if you have 16 layers, layer 1 in the instrument layer window would actually be the 16th wav file, instead of the 1st, and likewise, layer 16 would be the 1st wav file.

In this 2nd image, which shows my ‘fix’ for this issue, notice that the 16th layer contains the 1st snare file (which is intended to be the softest snare hit). While the files are actually loaded in reverse, using ‘Set automatic velocity’  maps the velocity ranges correctly.

hydro-solution

Unfortunately, by using this method, you cannot batch import the files, since Hydrogen will load them the conventional way, where layer 1 = sound file 1, and so on. Since you cannot reverse sort the layers, you have to add each layer individually, backwards, where sound file 1 is loaded into the 16th layer first, and so on. Then when importing the final layer, choose ‘Set Automatic Velocity’.

theGreatWhiteShark:
Fixes #841.

During debugging I noticed how hard and clumsy this part of Hydrogen was to use (haven't used it much before). So, I also added some tweaks in the UI/UX

- the colors of the InstrumentEditor do now base on the once set in the color theme and should integrate better in the overall application
- when hitting the "remove layer" button Hydrogen will autoselect the next layer or - in case the last one was already selected - the previous one. This allows for a fast removal of all loaded layers. (Beforehand one had to manually select each individual layer by mouse before pressing the button again)
- deactivate layer widgets if no Layer is present
- red color is used to highlight missing samples. Previously, there was just a banner message and a warning icon in the pattern editor instrument list to indicate missing ones.
- Another convenience layer was build into the default path selection of the AudioFileBrowser (which is used for loading the PlaybackTrack and the Layer files). If "Edit" (PlaybackTrack) or "load layer" is clicked while there is already a sample present, the dialog will select this file as default. If none is present yet, the browser will resort to the last opened directory stored in the preferences. And if this one was not set yet, the home folder will be used
- The selected layer of LayerPreview and InstrumentEditor went out of sync once a different instrument had been selected. This is fixed.
- in the custom dialog to load e.g. samples for a couple of layers of an instrument one can now use the Shift modifier to make multiple selections. (clicking at one item, press-holding shift, clicking another item). All previous ways of selection are still supported.


Naddiseo:
@theGreatWhiteShark thanks for fixing #841!!

theGreatWhiteShark:
> @theGreatWhiteShark thanks for fixing https://github.com/hydrogen-music/hydrogen/issues/841!!

You are welcome :)

PR #1503, opened 2022-02-26
Fast exponential ADSR envelopes.

cme:
As discussed in #1469 , this PR re-works the ADSR envelope application to use exponentials, and use much less CPU time to calculate and apply.


theGreatWhiteShark:
Hey,
FYI: I just got some very strange behavior due to the ADSR. It seems to be around since https://github.com/hydrogen-music/hydrogen/pull/1574 and can be reproduced like this

1. Open a new song
2. Enabled metronome
3. Start playback

The click does not have consistent volume!

 It sounds a bit like random variations but it turned out to be deterministic. After dissecting the Sampler (at a more recent commit) I found that commenting `applyADSR` in https://github.com/hydrogen-music/hydrogen/blob/master/src/core/Sampler/Sampler.cpp#L1251 does fix it. But I'm still quite puzzled why 1. it does not make the export song unit test fail and 2. I never encountered it with an instrument of the GMRockKit I used for developing quite a lot lately.

cme:
I can't seem to reproduce this. I can reproduce the random volume changes in volume of the metronome by turning up the velocity humanization (agreed, that seems wrong! :D ) but only with humanisation. What sort of range of volume variation do you get? Maybe it's just my ears aren't so good ....

theGreatWhiteShark:
> I can't seem to reproduce this.

Hmm. I'm running a JACK server with sample rate 48000, buffer size 4096 and a period of 3. I need to finish the audio engine stuff first and have a more closer look into this one afterwards.

> Maybe it's just my ears aren't so good ....

Well, the variations are quite pronounced :D . Actually, while recording a longer sample I noticed that there is a periodicity.

[click.tar.gz](https://github.com/hydrogen-music/hydrogen/files/9811155/click.tar.gz)

theGreatWhiteShark:
> Hmm. I'm running a JACK server with sample rate 48000, buffer size 4096 and a period of 3. I need to finish the audio engine stuff first and have a more closer look into this one afterwards.

Turns out I can only reproduce it for this particular buffer size.

TBH looking at the ADSR code gives me a bad vibe. It looks quite archaic with lots of comments and variable names indicating the usage of ticks instead of frames for positions (which is not the case). One thing I did already spot is that the call to `applyADSR` in the sampler when performing resampling is missing the `fStep` variable as its last argument. But this itself does not cause this bug.

I will do an audit of this part of the code during next week. Would be a shame if the `Sampler` glitches for certain plausible audio driver configurations. Or maybe reworking the audio engine did have some side effects in here.

Another thing I realized is that the basic unit for attack, decay, release but also for lead and lag are frames. An attack of `1000` will apply the attack envelop to the first `1000` frames in the sample. And it will be always this number of frames regardless whether the note is e.g. used in its original sample rate of 44100 or resampled to 96000. Changing the sample rate thus affect the way ADSR is applied to a note as well as the maximum lead and lag available in humanization. But this feels more like something we can safely postpone for v1.3.

theGreatWhiteShark:
> Another thing I realized is that the basic unit for attack, decay, release but also for lead and lag are frames. An attack of 1000 will apply the attack envelop to the first 1000 frames in the sample. And it will be always this number of frames regardless whether the note is e.g. used in its original sample rate of 44100 or resampled to 96000. Changing the sample rate thus affect the way ADSR is applied to a note as well as the maximum lead and lag available in humanization. But this feels more like something we can safely postpone for v1.3.

Ah, wait. This is what `fStep` is doing.

But the sampler should at least check whether the sample needs to be up or down sampled and apply the ADSR before or after the resampling. The lesser calls the better and the order of processing shouldn't affect sound quality.

cme:
> > Hmm. I'm running a JACK server with sample rate 48000, buffer size 4096 and a period of 3. I need to finish the audio engine stuff first and have a more closer look into this one afterwards.
> 
> Turns out I can only reproduce it for this particular buffer size.

...even weirder. I wouldn't expect buffer size to have any effect, although it's conceivable sample rate might since things are in an odd mix of time domains. I don't currently have a Linux machine I can actually *hear* so I can't even attempt to reproduce :/

I checked that the metronome notes are at least inserted with sensible default ADSR parameters, so ADSR for these should basically be a pass-through.

> TBH looking at the ADSR code gives me a bad vibe. It looks quite archaic with lots of comments and variable names indicating the usage of ticks instead of frames for positions (which is not the case).

Agreed, I had to fight the urge to make more major changes but stopped because I didn't want to change any more visible behaviour than necessary. Having things dependent on the sample rate the audio system is running at *or* the sample rate of the sample just feels wrong. It's debatable what the *right* thing for ADSR should be. Real time units is closest to 'right' for the usual uses of ADSR for synthesis, but since samples already inherently have their own envelopes, applying ADSR at all will get the product of those envelopes, and there's no sense in sustain and release unless the sample is sufficiently long...!

> One thing I did already spot is that the call to `applyADSR` in the sampler when performing resampling is missing the `fStep` variable as its last argument. But this itself does not cause this bug.

Hm, I can't remember off the top of my head what this was supposed to do, but having it set to '1' in both calls certainly seems wrong, do you think this is a bug?

theGreatWhiteShark:
> ...even weirder. I wouldn't expect buffer size to have any effect, although it's conceivable sample rate might since things are in an odd mix of time domains. I don't currently have a Linux machine I can actually hear so I can't even attempt to reproduce :/

Don't worry. I can reproduce it and try to fix it over the weekend.

> Hm, I can't remember off the top of my head what this was supposed to do, but having it set to '1' in both calls certainly seems wrong, do you think this is a bug?
> 
> Edit: yep, that's my bug for sure

It's a shame this wasn't caught by one of the unit tests. I'll add one which covers it.

> Real time units is closest to 'right' for the usual uses of ADSR for synthesis, but since samples already inherently have their own envelopes, applying ADSR at all will get the product of those envelopes, and there's no sense in sustain and release unless the sample is sufficiently long...!

Yeah. Maybe the whole content dates back to times the `Synth` class was more than just dead code. (By the way, AFAICS there is no way to use it anymore but it still consumes processing time in each cycle of the audio engine - setting a buffer to 0 and adding those zeros to the main output. We should probably get rid of it).

Anyway. There are still some drumkits hosted by us who have some of the ADSR set to non-default values.

theGreatWhiteShark:
Ok. Found it. The bug that caused these audible glitches was introduced just recently in feca16aed0d01545bab7273b19b317fdc141fae9 .

The calculation of `nNoteEnd` in the sample rendering (used to determine the release frame in `applyADSR`) is off.  When performing resampling it does not take the initial silence at the beginning of the buffer into account. E.g.  I have a buffer of `4096` frames and a metronome sample of `2700` frames placed by the audio engine in such a way it starts at frame `3520` within the buffer (second click in the audio sample attached above). Since this initial silence was missed, `nNoteEnd` is set to `2071` ( < `3520` ) and `applyADSR` goes right into the release state and the faint metronome click we hear is actually the default `1000` release frames of the original sample multiplied with the release envelope.

This problem was not caught by an unit test since it uses a buffer size of `1024` and there seem to be no samples involved which are smaller than that.

PR #1513, opened 2022-03-04
Fix ALSA audio driver

theGreatWhiteShark:
I was lucky enough to get the ALSA driver in such a bricked state that both the transport/playback was not working anymore and Hydrogen did freeze when trying to set a different audio driver in the PreferencesDialog. The latter is indeed a quite serious problem. In case the user sets a "wrong" configuration, it can not be changed via the Preferences anymore but, instead, has to be manually reverted in the hydrogen.conf file. After some debugging I found that the processing thread of the ALSA driver did freeze after two cycles. The state of the playback stream itself was still nice but the corresponding resource "temporary unavailable". Since the device is opened in blocking mode in AlsaAudioDriver::connect() the process thread did freeze at snd_pcm_writei(). The call to AlsaAudioDriver::disconnect() (required to restart/start a new audio driver) is waiting for the thread to finisih which is not going to happen. Thats why Hydrogen did freeze at both shutdown and restarting the driver in the Preferences.

One possible way to fix this is to open the device in non-blocking mode. But this would be quite an invasive change and I am not too familiar with ALSA. So, instead, I introduced a call to the snd_pcm_wait() function which returns as soon as the opened device is ready to process data or returns with an error after a provided timeout. AFAIU this is also done internally in snd_pcm_writei() in blocking mode and will thus only introduce a minimal overhead. On the other hand, it is able to discard the audio output in case the ALSA device is not working anymore. Instead of freezing, it now attempts to annoy the user with error log messages as well as flashing XRun indications and status messages in the GUI in order to indicate a change of the audio driver is required.

I used 100 milliseconds for the timeout. It is enough to make the transport run way to slow to indicate a problem. Smaller values of e.g. 10ms did cause glitches with working ALSA devices.

In addition, I made the ALSA driver use our XRUN event and used the snd_pcm_recover() functions over our custom alsa_xrun_recovery() as the former is part of the ALSA API and recovers from one more error code than our custom one.

Further changes:
- tweaking/formatting log messages used in the process of starting an audio driver
- instead of just using the default "hw:0" device for the ALSA driver the preferences does now check whether this device actually exists on the system. If not, it falls back to the first device returned by AlsaAudioDriver::getDevices() that is not "null".
- the "alsa_audio_driver" node was removed deliberately from the default config file in order for the system to pick a proper default value of an existing device.
- In case the "Auto" driver is selected, the device combobox is now properly showing the content associated with the underlying driver. In addition, all values set there, in the host API combo box or the one for the latency are now properly written to the Preferences too.
- number of XRuns are now shown in the GUI (status message)
-  display ALSA device in PreferencesDialog info
    
    The AlsaAudioDriver tries to establish a connection to the device selected by the user via the Preferences. If this fails, it tries instead to connect to the "default" device which usually works. This change of devices is, however, only shown in the log and users may easily miss it. Now, the currently connected ALSA device is shown in the info section on the right of the audio driver tab.

### DIFF
--- a/data/hydrogen.default.conf
+++ b/data/hydrogen.default.conf
@@ -60,10 +60,6 @@
 			<jack_track_outs>false</jack_track_outs>
 		</jack_driver>
 
-		<alsa_audio_driver>
-			<alsa_audio_device>hw:0</alsa_audio_device>
-		</alsa_audio_driver>
-
 		<midi_driver>
 			<driverName>ALSA</driverName>
 			<port_name>None</port_name>
--- a/src/core/AudioEngine/AudioEngine.cpp
+++ b/src/core/AudioEngine/AudioEngine.cpp
@@ -166,13 +166,13 @@ AudioEngine::~AudioEngine()
 {
 	stopAudioDrivers();
 	if ( getState() != State::Initialized ) {
-		___ERRORLOG( "Error the audio engine is not in State::Initialized" );
+		ERRORLOG( "Error the audio engine is not in State::Initialized" );
 		return;
 	}
 	m_pSampler->stopPlayingNotes();
 
 	this->lock( RIGHT_HERE );
-	___INFOLOG( "*** Hydrogen audio engine shutdown ***" );
+	INFOLOG( "*** Hydrogen audio engine shutdown ***" );
 
 	clearNoteQueue();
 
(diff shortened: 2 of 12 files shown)


cme (approved):
Nice!

PR #1515, opened 2022-03-04
Fix crash on changing song with Automation Path showing

cme:
The automation path keeps a pointer to the song's AutomationPath. This pointer is updated by the Song Editor, but only as a result of the event queue, which is asynchronous (and also potentially lossy). If some event (such as a paint event caused to repair the display after a confirmation dialog is closed...) is delivered to the view widget after the song is changed (and the old song destroyed) but before the Song Editor has processed the update event, the old pointer will be dereferenced, likely causing a crash.

This change fixes this race condition by explicitly updating the saved pointer from the song at the start of asynchronously delivered events. This doesn't feel quite right but is at least consistent with other editors where state is updated directly from the Song. One alternative approach would be to use a shared_ptr in the Song and the editor to ensure that the object at least still exists when it's referenced in such cases, but that feels further from the right fix as it presents inconsistent state.

### DIFF
--- a/src/gui/src/Widgets/AutomationPathView.cpp
+++ b/src/gui/src/Widgets/AutomationPathView.cpp
@@ -1,7 +1,7 @@
 /*
  * Hydrogen
  * Copyright(c) 2002-2008 by Alex >Comix< Cominu [comix@users.sourceforge.net]
- * Copyright(c) 2008-2021 The hydrogen development team [hydrogen-devel@lists.sourceforge.net]
+ * Copyright(c) 2008-2022 The hydrogen development team [hydrogen-devel@lists.sourceforge.net]
  *
  * http://www.hydrogen-music.org
  *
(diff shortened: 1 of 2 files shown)


theGreatWhiteShark:
These changes cause Hydrogen to crash when moving a node of the automation path.

I just had a quick look. It might be caused by

``` c++
void AutomationPathView::setAutomationPath( AutomationPath *path, bool bUpdate )
{
	_path = path;
	if( _path ) {
		_selectedPoint = _path->end();
	}
	[...]
```

interfering with 

``` c++
void AutomationPathView::mouseMoveEvent(QMouseEvent *event)
{
	updateAutomationPath();
	if (! checkBounds(event) || !_path) {
		return;
	}

	auto p = locate(event);
	float x = p.first;
	float y = p.second;

	if(m_bIsHolding) {
		_selectedPoint = _path->move(_selectedPoint, x, y);
		[...]
```

and

``` c++
AutomationPath::iterator AutomationPath::move(iterator &in, float x, float y)
{
	_points.erase(in);
	[...]
```

cme:
Oh good shoit, i hadn't spotted the sticky references. I'll investigate!

Issue #1512, opened 2022-03-03
Double played note


oddtime:
playing in pattern mode, if a note is placed on tick 0 it's triggered two times (I hear two attacks)

cme:
Interesting... can you share a reproducible song example? Is it with playback in pattern mode? That would imply that at least one of the notes entered in the queue is played at the wrong time! 

But also: the current ADSR implementation (which I'm fixing...) shares ADSR state between all notes on the same instrument. So if a note is followed immediately by another note on the same instrument, no matter how soft the velocity is, it will restart the attack for the previous note, so there's a possibility that might be what you're encountering ?

oddtime:
> Is it with playback in pattern mode?

yes

> can you share a reproducible song example?

just enter a note in tick 0 and launch the playback

theGreatWhiteShark:
@oddtime which commit are you working on? There have been quite huge changes in the audio engine lately.

oddtime:
> which commit are you working on?

I have just pulled from master

cme:
Confirmed, I hear the same thing.

cme:
A quick bisect, it sounds like this came in with 0fecc5af76ee0567ff3e346ac1d26835fa25bd2e . Sorry @theGreatWhiteShark ...

theGreatWhiteShark:
> A quick bisect, it sounds like this came in with https://github.com/hydrogen-music/hydrogen/commit/0fecc5af76ee0567ff3e346ac1d26835fa25bd2e . Sorry @theGreatWhiteShark ...

Well, since I had to rewrite half of the audio engine for fixing the tempo marker support I expected to miss at least some glitches. Let's just hope we catch all major ones before the proper release. So, thanks a lot @oddtime for reporting!

I'll have a look in 1-2 days. (A couple of hours ago I was lucky enough to bring my JACK server in a state neither it nor ALSA are able to start properly and Hydrogen is not able to do any audio processing. I don't want to spoil this setup yet :) )

theGreatWhiteShark:
Hmm. I can't reproduce it on 6d2386464dd435df822dc66dd6898ec622858835. I created a new song, inserted a single note and started playback in both pattern and song mode. Either way and regardless of picking JACK, ALSA, or PortAudio driver I just hear and see a single note.

There are two commented debug messages regarding such problems in the code: 
[...]

PR #1524, opened 2022-03-11
Jittering in unit tests and fixing #1512

theGreatWhiteShark:
As the bug reported in #1512 did not happen with the buffer size I used locally, I took the long route fixing by introducing some configuration jittering in the audio engine unit tests and adding a unit test covering the note enqueuing. Each of these unit tests are now done with up to 15 different configurations of buffer size and sample rate while 5 of them are drawn at random.

Also I realized I didn't gave the playback in pattern mode the attention it would deserve and added a unit test for it too and changed a couple of things: 

The update of the AudioEngine::m_pPlayingPatterns vector has been refactored into a separate function and is now also performed when relocating transport. This way the transport position in Song mode and the list of patterns being played back is always in sync.

This results in two changes of the UX
- it is now sufficient to relocate to a specific column in song mode in order to add the patterns there to the currently playing ones. Switching back to pattern mode with stacked mode enabled let you play back this selection right away.
- When middle-clicking a pattern in order for it to be played back one can start its playback immediately by pressing the fast forward button or changing to a tick bigger than the current pattern size in general.

In adidtion, the deletion and adding of patterns has been moved into the CoreActionController entirely. This was done to ensure the patterns are cleaned up properly and fixed a couple of bugs
- loading a pattern, start playing it back, and removing it via an Undo action was not cleaned up properly in all instances it occured in the GUI. In one case it resulted in a segfault
- adding the loaded pattern to AudioEngine::m_pNextPattern by middle-clicking it and then removing the pattern using Undo was never handled and always resulted in a segfault.

Also, adding a virtual pattern in stacked pattern mode did not result in adding all patterns contained in the virtual one. This was fixed as well.

Further changes:
- The COLUMN_CHANGED_EVENT was dropped in favor for PATTERN_CHANGED_EVENT
- Hydrogen::sequencer_setNextPattern() and Hydrogen::sequencer_setOnlyNextPattern() have been renamed and moved to AudioEngine::toggleNextPattern() and AudioEngine::flushAndAddNextPattern()
- Introduceing Song::getAllNotes() which returns all notes in the song (using the current state of the SongEditor grid) while putting the notes on the same position (apart from random contributions) they would reside on when added by AudioEngine::updateNoteQueue().
- move Sample selection into Note class

   previously the Note selection was done inline and with a lot of duplicated code inside the Sampler. This logic has been moved to the Note class in order to make it reusable.
- AudioEngine::m_pPlayingPatterns and AudioEngine::m_pNextPatterns are now only provided by the AudioEngine as const pointers to other classes. All their handling is done in AudioEngine
- Updating the sample rate in the driver or restarting it is now handled by the audio engine as this invalidates some variables.


oddtime:
H2 is crashing when I delete a pattern

oddtime:
also, some notes are not played here:
[file.zip](https://github.com/hydrogen-music/hydrogen/files/8237938/file.zip)

theGreatWhiteShark:
> H2 is crashing when I delete a pattern

My bad. Fixed

> also, some notes are not played here:
[file.zip](https://github.com/hydrogen-music/hydrogen/files/8237938/file.zip)

This is quite strange. While listening to the song in both song and pattern mode I do hear all notes. When putting the song  in the unit test "for thorough inspection" everything passes as well.

Which audio driver, sample rate, and buffer size are you using? Maybe there is a parameter region I missed

oddtime:
Mmm... alsa 44100 256

theGreatWhiteShark:
> Mmm... alsa 44100 256

Unfortunately, I can not reproduce it.

Could you uncomment the debug messages in 
- https://github.com/hydrogen-music/hydrogen/blob/master/src/core/AudioEngine/AudioEngine.cpp#L2267
- https://github.com/hydrogen-music/hydrogen/blob/master/src/core/AudioEngine/AudioEngine.cpp#L1268
- https://github.com/hydrogen-music/hydrogen/blob/master/src/core/AudioEngine/AudioEngine.cpp#L1987

and attach the log? Sorry for the inconvenience.

theGreatWhiteShark:
> I cannot reproduce it anymore, sorry

No problem. But due to the work in another branch I found yet another little bug which might be the cause of the loss of notes (although I was not able to "exploit" it and to create artifacts myself).

PR #1529, opened 2022-03-12
Set output audio buffer size on CoreAudio (macOS)

cme:
On CoreAudio, we previously accepted the default buffer size setting. Which was reasonable (and certainly better than PortAudio's default on Windows), but could easily be better, and there's no good reason to *not* set it.

So let's set it.

Playing around with this I also realised that the estimated latency display doesn't actually account for buffer size, so that's added too.


src/core/IO/CoreAudioDriver.cpp
```diff
@@ -428,6 +428,26 @@ int CoreAudioDriver::init( unsigned bufferSize )
 		ERRORLOG( "Could not Initialize AudioUnit" );
 	}
 
+	// Set buffer size
+	ERRORLOG( QString( "Setting buffer size to %1" ).arg( nBufferSize ) );
```
theGreatWhiteShark, line 432: Shouldn't this be an `INFOLOG`?

cme: It absolutely should.

PR #1530, opened 2022-03-12
Code cleanup: Add missing member initializations, remove dead code

mauser:
Fix different small issues which have been found by clang tidy.


src/core/Timeline.cpp
```diff
@@ -90,7 +90,7 @@ float Timeline::getTempoAtColumn( int nColumn ) const {
 		return m_fDefaultBpm;
 	}
 
-	float fBpm;
+	float fBpm = 120.0;
```
theGreatWhiteShark, line 93: Can we make this default to `m_fDefaultBpm`? It might make debugging easier.

mauser: Sure, that makes sense!

PR #1544, opened 2022-03-23
Tweaking Pattern and Song editor

theGreatWhiteShark:
I did refine mostly UI but also UX of the SongEditor and PatternEditor. Mostly little graphical things to ensure those parts of the GUI pick up and use the colors of the color theme as well and fine-tuned presentation. The changes in the UX are mostly minor and are intended to align the UX of the different editors. Some features seem to have grown naturally at some places but not in others where it feels odd that they are missing. Only the stuff concerning the **tags**  are major UX changes. I really hated that one can only access them via mid-click on the position ruler which makes them somewhat a hidden feature. Everything is of course up for discussion.

tweaked_editor

## UX changes SongEditor

- Tags have been moved into Timeline (next to the Tempo Markers) in order to make room to accommodate the cursor in the ruler to highlight the current position
- Tags can be inserted by left-clicking the bottom area of the Timeline (above the ruler). The area is highlighted when hovered. This both hints to the user that something is happening when clicking as well as keeps the part of the Timeline used to add a Tempo Marker apart from the one dedicated for tags. Inserting tags using middle mouse click is still supported and can be now done for the whole height of the ruler
- Clicking the ruler is now always possible and automatically switches transport into Song Mode.
- Full-size playhead (supersedes #1107)
- The icons in the pattern list indicating whether a pattern is playing in stacked pattern mode are now colored and can have four different states: on, off, off next (pattern is played till the end and then turned off), and on next (pattern is played as soon as transport is looped again) (supersedes #482)

## UX changes PatternEditor

- Relocating transport by clicking the ruler is now supported (like in the SongEditor) and automatically switches transport into Pattern Mode.
- Full-size playhead
- The ruler was decoupled from the currently selected pattern. It always has the size of the largest playing pattern and always shows the transport position using a playhead. Whether or not the current pattern is played back is indicated by a full-height  cursor.
- All note properties except of the note key can now be altered in both the drum pattern editor and the piano roll editor by right-clicking and dragging a note.
- All notes of the currently playing patterns will be hinted in stacked pattern mode. Even those exceeding the length of the current pattern.

For the full list of changes please see the commit messages. (I know, not a nice move. But they are just way too many.)

P.S. If you are already on master you probably need to import the update theme.
P.P.S These are most probably the last big UI/UX changes I'll do before 1.2. But there is still a very long list of todos and odd stuff to look into. Let's hope they don't lead down too deep into tweak-rabbit hole.


src/gui/src/Skin.h
```diff
@@ -68,12 +68,41 @@ class Skin
...
 		color are reduced by this factor.*/
 	static QColor makeWidgetColorInactive( QColor color );
 		/** If a widget is marked inactive the value of its text color
 		are reduced by this factor.*/
 	static QColor makeTextColorInactive( QColor color );
+
+	static constexpr int nPlayheadWidth = 11;
+	static constexpr int nPlayheadHeight = 8;
+	static constexpr int getPlayheadShaftOffset() {
+		return std::floor( Skin::nPlayheadWidth / 2 ); }
```
cme, line 93: `std::floor` isn't guaranteed to be `constexpr` [until C++23](https://en.cppreference.com/w/cpp/numeric/math/floor). Seems it's not `constexpr` on macOS.

cme: Oh, rats, looks like the Appveyor builds completed successfully despite it failing to build. That's almost certainly my fault, sorry.

cme:
This is looking great! Wee issue with hi-dpi displays on the pattern editor:

image

cme:
Actually, also looks like a there's a couple of 1-pixel gaps (half-pixel on regular display so invisible) between rulers that aren't repainted?

theGreatWhiteShark:
> Wee issue with hi-dpi displays on the pattern editor:

Should be fixed now.

> Actually, also looks like a there's a couple of 1-pixel gaps (half-pixel on regular display so invisible) between rulers that aren't repainted?

The devil's in the detail. I could also produce this gap when setting `QT_SCALE_FACTOR` on my device. While at it I fine-tuned a number of pixel stuff. But some things just won't align. The grid of the song editor and the one of it's ruler is apart a tiny bit. But the corresponding x values as well as the parent widgets geometry do match. It's just when painting a 2 pixel line at position 0 in the ruler, I see a line of 1px width and when doing exactly the same in the SongEditor, it's 2px wide. No idea why. But it's just a tiny detail and probably not worth worth too much time

PR #1551, opened 2022-03-31
fix #1542: notes exceeding pattern length

### LINKED ISSUE #1542: Pattern length doesn't take effect in stacked pattern mode
Altering the length of a pattern does only affect the underlying notes if it is either the only pattern currently playing or the longest one.

Steps to reproduce:
1. Prepare two patterns of the same length and fill all notes for a different instrument each
2. Play back both patterns at once (by placing them in the same column in song mode or by using stacked pattern mode)
3. Reduce the length of one of the patterns (in 1.1.1 transport has to be paused to do so)

You will notice that regardless of how much the length of one pattern is reduced, you can still hear all notes.

The cause of this problem is that `Pattern::get_notes()` just returns a reference to the object containing all notes but does not check whether their `Note::__position` is smaller than the pattern length.

theGreatWhiteShark:
Next to `Pattern::__notes`, which is a container for all notes within the pattern, and its accessor `Pattern::get_notes()` the new function `Pattern::getAccessibleNotes()` was introduced. Its purpose is to creating a fresh map similar to `__notes` but only containing accessible notes and provide it as shared pointer. The notes itself will not be copied but only their pointers. A note is accessible if its tick position within the pattern is smaller than the pattern length.

`getAccessibleNotes()` is not used at all places as a replacement for `get_notes()` since serialization, clearing all notes/instruments, and redo/undo should still work on all notes without any loss of information. Whether a note is accessible is primarly affecting notes playback, user interaction, visual rendering, and MIDI export.

Fixes #1542

### DIFF
--- a/src/core/Basics/Note.cpp
+++ b/src/core/Basics/Note.cpp
@@ -113,7 +113,12 @@ Note::Note( Note* other, std::shared_ptr<Instrument> instrument )
 	  m_nNoteStart( other->getNoteStart() ),
 	  m_fUsedTickSize( other->getUsedTickSize() )
 {
-	if ( instrument != nullptr ) __instrument = instrument;
+	// Either use the provided instrument or fall back to the one
+	// stored in the other Note
+	if ( instrument != nullptr ) {
+		__instrument = instrument;
+	}
+
 	if ( __instrument != nullptr ) {
 		__adsr = __instrument->copy_adsr();
 		__instrument_id = __instrument->get_id();
@@ -134,8 +139,12 @@ Note::~Note()
 
 static inline float check_boundary( float v, float min, float max )
 {
-	if ( v>max ) return max;
-	if ( v<min ) return min;
+	if ( v > max ) {
+		return max;
+	}
+	if ( v < min ) {
+		return min;
+	}
 	return v;
 }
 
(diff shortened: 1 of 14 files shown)


src/core/Basics/Note.cpp
```diff
@@ -134,8 +139,12 @@ Note::~Note()
 
 static inline float check_boundary( float v, float min, float max )
 {
-	if ( v>max ) return max;
-	if ( v<min ) return min;
+	if ( v > max ) {
```
cme, line 142: Maybe just `return std::clamp( v, min, max )`?

src/core/Basics/Pattern.cpp
```diff
@@ -64,6 +64,19 @@ Pattern::~Pattern()
...
+
+	std::shared_ptr<notes_t> accessibleNotes = std::make_shared<notes_t>();
+	
+	for( notes_cst_it_t it=__notes.begin(); it!=__notes.end(); it++ ) {
+		if ( it->first < __length && it->second != nullptr ) {
+			accessibleNotes->insert( std::make_pair( it->first, it->second ) );
+		}
+	}
+
+	return accessibleNotes;
```
cme, line 77: Could use `std::move` here to avoid reconstructing the shared_ptr in the caller?

cme (commented):
General question because I don't know off the top of my head. Shouldn't we just delete any notes that are outside the length of the pattern? (With an undo-able action, of course) I'm not seeing why they should be kept?

theGreatWhiteShark:
> General question because I don't know off the top of my head. Shouldn't we just delete any notes that are outside the length of the pattern? (With an undo-able action, of course) I'm not seeing why they should be kept?

Good question. I'm so used to the non-destructive way it is done right now that I haven't even thought about it...

Hmm. As I do mostly audio recording I would somewhat expect this behavior as changing the length of audio tracks in most DAWs is done non-destructively too. I have to check how they handle things with MIDI tracks. Undo/redo is only making it non-destructive within the current session (but that's not really a strong argument).

What would be the benefit of discarding them? I think from the UX point of view there would be none as there are better ways to delete notes than changing the length of the pattern. But in terms of performance dropping the notes would definitely the better choice. 

Hmm. I have to think about it a little more but, yeah, deleting the corresponding notes might indeed be the right choice.

theGreatWhiteShark:
superseded by #1562

PR #1589, opened 2022-05-16
data/drumkits: change license GPL -> CC-BY-SA

theGreatWhiteShark:
The current license of the drumkits is GPL. However, this license is more suitable to code than to creative output. How about changing it to CC-BY-SA? This is somewhat the GPL of the cultural and artistic world.

### DIFF
--- a/data/drumkits/GMRockKit/drumkit.xml
+++ b/data/drumkits/GMRockKit/drumkit.xml
@@ -7,7 +7,7 @@
 p, li { white-space: pre-wrap; }
 &lt;/style>&lt;/head>&lt;body style=" font-family:'Lucida Grande'; font-size:10pt; font-weight:400; font-style:normal;">
 &lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">A Sampled 5pc Pearl DX Series Drumkit.&lt;/p>&lt;/body>&lt;/html></info>
- <license>GPL</license>
+<license>CC-BY-SA</license>
  <image></image>
  <imageLicense>undefined license</imageLicense>
  <componentList>
--- a/data/drumkits/TR808EmulationKit/drumkit.xml
+++ b/data/drumkits/TR808EmulationKit/drumkit.xml
@@ -7,7 +7,7 @@
 p, li { white-space: pre-wrap; }
 &lt;/style>&lt;/head>&lt;body style=" font-family:'Lucida Grande'; font-size:10pt; font-weight:400; font-style:normal;">
 &lt;p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;">This drum kit contains 16 original sounds that were synthesized from scratch (using only basic synth waves, like sine, noise etc.). It simulates the sounds of the legendary Roland TR-808 Rhythm Composer. More kits from ArtemioLabs: http://artemiolabs.com/downloads/audio/hydrogen/&lt;/p>&lt;/body>&lt;/html></info>
- <license>GPL</license>
+ <license>CC-BY-SA</license>
  <image></image>
  <imageLicense>undefined license</imageLicense>
  <componentList>


elpescado:
Hi,

I'm afraid changing licenses is a messy business. As far as I understand, basically it requires all original authors to approve relicensing. That might be a problem with drum kits. [TR808EmulationKit](https://github.com/hydrogen-music/hydrogen/tree/master/data/drumkits/TR808EmulationKit) is as old as the hills, I'm not even sure who to contact (Comix? Artemio?). [GMRockKit](https://github.com/hydrogen-music/hydrogen/tree/master/data/drumkits/GMRockKit) was composed by @mauser using samples from another kit(s), mainly The Black Pearl 1.0 (by Glen MacArthur).

And while I agree that CC is a better choice for non-code assets, I think that (disclaimer: I am not a lawyer!) having a single license for whole source tree makes more sense.

theGreatWhiteShark:
Hi @elpescado,

Thanks for the reply! Well, I'm not a lawyer either but creating a new drumkit from parts of an existing one might be different from using/changing the old one and it might not be required to ask people recursively all the way back to the person recording the samples. Actually, I'm not even sure different lawyers would agree on a single take to this issue. I've spoken to a lawyer of the FSFE a couple of years ago on the topic of licensing public data and he said one should never use GPL for anything other than code while CC is making things much more clear for those use cases. That's also the main motivation for changing it. The web page of CC offers loads of accessible material to inform artists what they can and what they can't do with the kits.

> [TR808EmulationKit](https://github.com/hydrogen-music/hydrogen/tree/master/data/drumkits/TR808EmulationKit) is as old as the hills, I'm not even sure who to contact (Comix? Artemio?)

Your right. Just checking the git history which people touched the drumkit isn't sufficient as the code base is way older.

 > And while I agree that CC is a better choice for non-code assets, I think that (disclaimer: I am not a lawyer!) having a single license for whole source tree makes more sense.

All (or at least most) of the SVGs used as icons are licensed as CC-0 (or public domain). So, already not exclusively GPL in here. 

In fact, CC-BY-SA and GPL are so similar version CC-BY-SA 4.0 is even [one way compatible](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3#Considerations_for_adapters_applying_the_GPLv3) with GPL3. This means when licensing multimedia files as CC-BY-SA 4.0 people who download and reuse the code are free to relicense the CC-BY-SA 4.0 parts as GPL3 without even consulting us.

But don't get me wrong. This is nothing I'm very eager to go but just a matter that caught my attention. If we decide to keep things as they are this is fine for me too. ;)

elpescado:
> Thanks for the reply! Well, I'm not a lawyer either but creating a new drumkit from parts of an existing one might be different from using/changing the old one and it might not be required to ask people recursively all the way back to the person recording the samples.

I am not a lawyer either, but I think that might actually be the case. By reusing samples we are effectively redistributing them, so we are bound to original license terms. And while this might not apply to some trivial changes, audio samples are quite essential - after all drum kit basically is a collection of samples.

Some projects get around this by requiring all contributors sign a contributor agreement that allows relicensing their contributions. But I think Hydrogen is too small for it to make sense.

By the way, this left me wondering about the "share-alike" aspect of both GPL and CC-BY-SA with regards to sound samples. What happens if someone records a song using samples from drum kit redistributed under either of those licenses (eg. by composing a song in Hydrogen using one of stock drum kits, exporting audio and using that drum track in a song) - is this final song considered a "derivative work"? If so, is it bound by GPL/CC-BY-SA terms?

theGreatWhiteShark:
> By the way, this left me wondering about the "share-alike" aspect of both GPL and CC-BY-SA with regards to sound samples. What happens if someone records a song using samples from drum kit redistributed under either of those licenses (eg. by composing a song in Hydrogen using one of stock drum kits, exporting audio and using that drum track in a song) - is this final song considered a "derivative work"? If so, is it bound by GPL/CC-BY-SA terms?

I don't think so. With resampling, pitch shifting, effects, ADSR etc. the sample is getting modulate by Hydrogen and you use some resources you are permitted to use in order to create a piece of art which is your own. If modulating the samples wouldn't be enough to term the result a fresh and unlicensed product, the copyright for the 808 kit would still be hold by Roland (at least I assume the kit uses samples from the TR-808 instead of recreating those by ear).

> I am not a lawyer either, but I think that might actually be the case.

I'll request approval from all other people contributing to the kits and ask the former member via mail once all of them agree. (So, this PR might stay open for quite a while :D )

theGreatWhiteShark:
@oddtime, @cominu would you agree to change the license of the **GMRockKit** and **TR808EmulationKit** from `GPL` to `CC-BY-SA`?

elpescado:
> the copyright for the 808 kit would still be hold by Roland (at least I assume the kit uses samples from the TR-808 instead of recreating those by ear).

I've done some archaeology on our mailing lists. It seems that TR808 Emulation Kit was [created](https://sourceforge.net/p/hydrogen/mailman/message/6813917/) by [Artemiy Pavlov](https://www.sinevibes.com/company/) (2005, what a year it was!), and [it doesn't contain any actual samples](https://sourceforge.net/p/hydrogen/mailman/message/6815231/) from Roland:

> The acoustic ones (including the default GM kit and UltraAcoustic) so far have 
sounds which were "stolen" from Roland modules (XV-5080 and Fantom-S, 
respectively) and in future I want to remove them and replace with other 
sounds (I have some good new ones, maybe even better than these).
> 
> As about the TR-808 kit (shipped by default with 0.9.2) and Electric Empire, 
these are the sounds synthesised from scratch (out of basic waves, e.g. sine, 
noise, etc.) with lots of re-synthesis cycles. I enjoyed creating these a lot 
and soon I will also publish at least one new kit synthesised with Roland 
V-Synth.

By the way, that's precisely why GMKit was removed and replaced by GMRockKit.

Commercial sample libraries have usually "royalty-free" licenses, that explicitly allow (once you pay) to integrate samples in music.

> >  is this final song considered a "derivative work"? If so, is it bound by GPL/CC-BY-SA terms?
> I don't think so. With resampling, pitch shifting, effects, ADSR etc. the sample is getting modulate by Hydrogen and you use some resources you are permitted to use in order to create a piece of art which is your own.

I've tried to find some reference on the interwebz but without any success. There's one interesting [thread](http://lalists.stanford.edu/lad/2012/08/0171.html) on linux-audio-dev mailing list, which [doesn't have clear conclusion](http://lalists.stanford.edu/lad/2012/09/0000.html):

> In this sense, your question has no clear answer, which is worse than a Yes. 

I think that when it comes to sound libraries (eg. drum kit samples) and "derivative works" there are two distinct scenarios to consider.

- Use of samples in other sound libraries (eg. repackaging sound library to another format, creating a sound library combining samples from other libraries)
- Use of sound samples as "ingredients" in music (eg. using drum samples to create a drum track of a song)

I'm afraid that CC wasn't designed with such distinctions in mind:/

oddtime:
Hello, I was no involved in making these drumkits, maybe you wanted to tag somebody else...?

theGreatWhiteShark:
> I've done some archaeology on our mailing lists. It seems that TR808 Emulation Kit was [created](https://sourceforge.net/p/hydrogen/mailman/message/6813917/) by [Artemiy Pavlov](https://www.sinevibes.com/company/) (2005, what a year it was!), and [it doesn't contain any actual samples](https://sourceforge.net/p/hydrogen/mailman/message/6815231/) from Roland:

Good to hear. :)

> I think that when it comes to sound libraries (eg. drum kit samples) and "derivative works" there are two distinct scenarios to consider.
>
>    Use of samples in other sound libraries (eg. repackaging sound library to another format, creating a sound library combining samples from other libraries)
>    Use of sound samples as "ingredients" in music (eg. using drum samples to create a drum track of a song)

Okay. I did a little bit of digging myself and things seem to be a lot worse than I thought. Quite dire to be honest. When proposing the relicensing to CC-BY-SA I took the original license for granted and thought them to be well established and to have already stood the test of time. But almost all other software I checked have all their samples licensed as CC-0 or CC-BY. And some threads, like [this](https://linuxmusicians.com/viewtopic.php?f=24&t=24436&p=143972&hilit=GPL+samples#p143972) and [this](https://linuxmusicians.com/viewtopic.php?f=64&t=22462&p=126939&hilit=GPL+samples#p126939), in the linuxmusicians forum are quite clear about that either of too things result from our kits being licensed as GPL

1. The license doesn't apply. I'm not a 100% sure what happens then but most likely it falls back to "all rights reserved". In which case we would have to remove all of them from Sourceforge. For one there is the overall nature of the audio samples. From [gnu.org](https://www.gnu.org/licenses/gpl-faq.html#GPLOtherThanSoftware):

    > You can apply the GPL to any kind of work, as long as it is clear what constitutes the “source code” for the work. The GPL defines this as the preferred form of the work for making changes in it.

    Also, none of them includes a copy of the GPL which is strictly required.

2. If GPL does apply to the kits, all music which includes even a fraction of one of its samples must be distributed as GPL as well. If any of those artists are not making their work available for free via e.g. bandcamp, [paragraph 5.2](https://softwarefreedom.org/resources/2008/compliance-guide.html) of the GPL would take effect and they would forfeit their rights to redistribute their music. Everyone would be allowed to sell their music for money etc. In addition, quite a number of musicians have contracts with labels which include transferring ownership or sublicense their work. I don't even want to think about whether they could be required to pay reparations due to contract violation just because they used our software.

On the bright side: only we as developers of Hydrogen [seem to have the right](https://www.gnu.org/licenses/gpl-faq.html#WhoHasThePower) to enforce the license. We can/should solemnly vow to not do that but that's not even close to a proper solution.

To conclude I would suggest we try to relicense both kits as CC-BY or CC-0 as quickly as possible or try to find a replacement. The kits on Sourceforge are fine as the authors are free to choose whatever license they would like to use. But we should definitely add a section in the docs covering the license issue and probably also display a (mutable) popup when loading a GPL or CC-BY-SA kit and exporting a song done with one of those kits. Maybe we should even contact the FSF and ask if there is a proper way to abstain from the right to enforce license violations resulting from composed music. Because I can't imagine that this was the intention of the original authors.

theGreatWhiteShark:
> Hello, I was no involved in making these drumkits, maybe you wanted to tag somebody else...?

By changing the default values of the pan you also contributed to the drumkits and have the same stakes as all others in this thread ;)

theGreatWhiteShark:
I will close this one since

1. it's not realistic everyone who contributed to the kits will give their vote anytime soon
2. from the user's perspective CC-BY-SA causes the same troubles as GPL and it is not immediately clear what would be the best license for the default kits

But in the mean time I introduced two patches (#1601 and #1620) which are intended to be helpful with managing licenses of different kits as well as with informing the user about possible consequences due to licensing issues. Also, I will add some paragraphs regarding this topic in the manual

I think the way to go from here would be to replace the GMRockKit at some point with a adequate kit licensed under CC-BY or CC0.

PR #1594, opened 2022-05-28
hydrogen.default.conf: update conf file

theGreatWhiteShark:


theGreatWhiteShark:
Due to a lucky typo in the conf file I was also a able to fix a rare bug. If both the system and user level config file are bricked, Hydrogen attempts to recreate it before the core is properly initialized and crashes.

PR #1614, opened 2022-07-09
Fix off-by-one error on empty pattern group vector

cme:
This caused a crash on opening a song with an empty pattern vector, and previously worked due to an underflow in comparison which was removed as a side effect of b3f93f3e3991bc1fba0649b5b3199f7549a10e58


theGreatWhiteShark:
These columns are really a nuisance. I wonder whether it was a design decision or grew naturally to initialize/reset the variable  (`AudioEngine::m_nColumn`) used to access the column vector as index with `-1`.

Issue #1592, opened 2022-05-28
Errors are not reported in GUI


cme:
Hydrogen version *          : all
Operating system + version  : all
Audio driver + version      : all

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Hydrogen has extensive logging of events, including errors. However, error reporting in the GUI is more ad-hoc and many things the user cares about are silently lost leaving them unable to troubleshoot problems. Fundamentally, as a GUI application, Hydrogen should allow users to see errors in a GUI, rather than stderr or a standaside logfile hidden in a 'Debug' menu.

This also makes user issue reports harder to categorise initially (although the logging itself makes them much easier to deal with comprehensively after initial engagement). Additionally it hides problems that might otherwise go unreported and which could otherwise be 

It's not clear how to best reconcile this with the existing logging infrastructure. It would be good to leverage existing error messages to be able to catch as many conditions as possible that the user might care about, but care would need to be taken to avoid swamping the user with error report dialogs, since existing error log message strings are tailored towards the log format and often report multiple log entries per error condition.

theGreatWhiteShark:
> care would need to be taken to avoid swamping the user with error report dialogs

It doesn't always have to be dialogs. We could also display errors (and warnings) as status messages in the PlayerControl (possibly in a different color and with the option to enable and disable them in the Preferences).

There is still the downside that only the newest message is displayed. But this itself is worth fixing right away. How about introducing a history of length X (adjustable in the Preferences). When clicking the status message QLineEdit widget a popup list (similar to a QMenu) will show the content of the history.

cme:
> > care would need to be taken to avoid swamping the user with error report dialogs
> 
> It doesn't always have to be dialogs. We could also display errors (and warnings) as status messages in the PlayerControl (possibly in a different color and with the option to enable and disable them in the Preferences).
> 
> There is still the downside that only the newest message is displayed. But this itself is worth fixing right away. How about introducing a history of length X (adjustable in the Preferences). When clicking the status message QLineEdit widget a popup list (similar to a QMenu) will show the content of the history.

Great idea, a widget like that could also be used as a toplevel dialog for fatal errors. 

I'll have a crack at this. I'll add an API to the Logger to record and access recent messages (I think probably also change conditions so that Info messages are kept, even if they're not logged, as they will provide useful context to any actual error), add an event for Errors being loged, etc.

[...]

PR #1615, opened 2022-07-09
Add simple GUI crash reporter

cme:
Simple crash reporter GUI, which gives options to examine the log, report bug, etc, and reports any information known about the context that caused the crash (eg. LADSP plugins). This addresses #1592 .

image

Currently this is a draft as it could use some polishing, and also because it's shown some errors I hadn't noticed before in some fairly common circumstances, so holding off adding until I investigate some of those.


sirjohndfox:
Gday. When trying to compile the master branch on Debian stable with gcc v8.3. I get this error

hydrogen/src/gui/src/Reporter.cpp:179:23: error: comparison between distinct pointer types ‘QPushButton*’ and ‘QAbstractButton*’ lacks a cast
    if ( pLogButton == pPushed ) {
                       ^~~~~~~
hydrogen/src/gui/src/Reporter.cpp:182:27: error: comparison between distinct pointer types ‘QAbstractButton*’ and ‘QPushButton*’ lacks a cast
    } else if ( pPushed == pIssuesButton ) {
                           ^~~~~~~~~~~~~
                           
  Then @theGreatWhiteShark added a cast (i was trying to test one of his pull requests) and I got the error:
  hydrogen/src/gui/src/Reporter.cpp: In member function ‘void Reporter::on_finished(int, QProcess::ExitStatus)’: /home/john/SOFTWARE/hydrogen/src/gui/src/Reporter.cpp:179:49: error: invalid static_cast from type ‘QPushButton*’ to type ‘QAbstractButton*’ if ( static_cast<QAbstractButton*>(pLogButton) == pPushed ) { ^ /home/john/SOFTWARE/hydrogen/src/gui/src/Reporter.cpp:183:57: error: invalid static_cast from type ‘QPushButton*’ to type ‘QAbstractButton*’ else if ( static_cast<QAbstractButton*>(pIssuesButton) == pPushed ) {

cme:
Ooh this is weird. That should he a perfectly valid upcast as `QPushButton` inherits `QAbstractButton`.

Did that file report any warnings before the error?

Have you any extra compile flags (like, say, `-Werror`)

cme:
> Gday. When trying to compile the master branch on Debian stable with gcc v8.3. I get this error
> 
> hydrogen/src/gui/src/Reporter.cpp:179:23: error: comparison between distinct pointer types ‘QPushButton*’ and ‘QAbstractButton*’ lacks a cast if ( pLogButton == pPushed ) { ^~~~~~~ hydrogen/src/gui/src/Reporter.cpp:182:27: error: comparison between distinct pointer types ‘QAbstractButton*’ and ‘QPushButton*’ lacks a cast } else if ( pPushed == pIssuesButton ) { ^~~~~~~~~~~~~

Ah, it looks like I missed including QtWidgets explicitly. This can get missed in testing because the CMake setup has pre-compiled headers to speed up building, so it masks such an error if pre-compiled headers aren't used

(By the way, which version of CMake do you have?)

I'll merge the fix as soon as the build checks are complete but if you're super-keen you could try #1641

sirjohndfox:
Cheers for looking into this.  I've cmake version 3.13.4.  I'll have a go at compiling 1641 presently.

sirjohndfox:
Compiles no worries!
@theGreatWhiteShark  This fix means your adding of cast wasn't necessary - I bodged the corrected files back into #1640 locally and it compiled fine!  I'm not sure of the message etiquette here. I realise I might've cross-posted a bit - is there anything I need to close?

cme:
> Cheers for looking into this. I've cmake version 3.13.4. I'll have a go at compiling 1641 presently.

Ah that explains it, 3.13 is too early for pre-compiled header support.

Fix is on master now.

theGreatWhiteShark:
> I realise I might've cross-posted a bit - is there anything I need to close?

Nope. AFAICS everything is fine

PR #1624, opened 2022-07-29
[PATCH] Follow XDG Base Directory spec on Linux

### LINKED ISSUE #643: ~/.hydrogen does not follow XDG standards
Hydrogen makes a .hydrogen directory.
```
.hydrogen
├── data
└── hydrogen.conf

1 directory, 1 file
```
[XDG standards](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) specify that config files should go in ~/.config and data in ~/.local/share.

Therefor, the `~/.hydrogen/data` directory should be moved to `~/.local/share/hydrogen` and `hydrogen.conf` should be moved to `~/.config/hydrogen.conf` by default.

polirritmico:
Fix #643. Uses QStandardPaths to get cache, data and config XDG folders **only on Linux**. If old folder (e.g. `~/.hydrogen/`) is found, use that instead of the XDG paths. If no folder is found, create the XDG dirs.

According to https://doc.qt.io/qt-5/qstandardpaths.html this should be the default paths returned by QStandardPaths (Follow the ENV paths configured by the user):

| QStandardPaths | Linux  Path|
|-------------------|------------|
| AppConfigLocation |  ~/.config/hydrogen |
| AppLocalDataLocation | ~/.local/share/hydrogen |
| CacheLocation |  ~/.cache/hydrogen |

The `__usr_*_paths` are setted outside a function, so QStandardPaths isn't avaliable to get the XDG paths. Also no APPNAME is avaliable until setApplicationName() is called after the `Filesystem::check_usr_paths()` is used. To avoid this `update_usr_paths()` is added at the begining of `check_usr_paths()` to update the correct values from QStandardPaths.

### DIFF
--- a/src/core/Helpers/Filesystem.cpp
+++ b/src/core/Helpers/Filesystem.cpp
@@ -31,6 +31,10 @@
 #include <QtCore/QFileInfo>
 #include <QtCore/QCoreApplication>
 #include <QDateTime>
+#if defined(Q_OS_MACX) || defined(WIN32)
+#else
+#include <QStandardPaths>
+#endif
 
 #ifdef H2CORE_HAVE_OSC
 #include <core/NsmClient.h>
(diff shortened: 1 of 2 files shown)


theGreatWhiteShark:
Superseded by #2339

PR #1634, opened 2022-08-14
Pass SIGTERM etc to child from Reporter

cme:


theGreatWhiteShark:
Thanks a lot! Works like a charm.

PR #1643, opened 2022-08-22
Dev perf mem leaks

cme:
Fix some memory leaks (including a major one in Song Editor) and a couple of small performance improvements.

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditor.cpp
+++ b/src/gui/src/PatternEditor/PatternEditor.cpp
@@ -96,6 +96,13 @@ PatternEditor::PatternEditor( QWidget *pParent,
 	m_pBackgroundPixmap->setDevicePixelRatio( pixelRatio );
 }
 
+PatternEditor::~PatternEditor()
+{
+	if ( m_pBackgroundPixmap ) {
+		delete m_pBackgroundPixmap;
+	}
+}
+
 void PatternEditor::onPreferencesChanged( H2Core::Preferences::Changes changes )
 {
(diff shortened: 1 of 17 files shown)


src/gui/src/PatternEditor/PatternEditor.cpp
```diff
@@ -96,6 +96,13 @@ PatternEditor::PatternEditor( QWidget *pParent,
 	m_pBackgroundPixmap->setDevicePixelRatio( pixelRatio );
 }
 
+PatternEditor::~PatternEditor()
+{
+	if ( m_pBackgroundPixmap ) {
+		delete m_pBackgroundPixmap;
```
theGreatWhiteShark, line 102: Hmm. I'm a little bit puzzled. I thought Qt's runtime takes care of cleaning after all the Q* classes. Is the manual cleanup required for just specific classes?

cme: > Hmm. I'm a little bit puzzled. I thought Qt's runtime takes care of cleaning after all the Q* classes. Is the manual cleanup required for just specific classes?

It's not an automatic thing so much as it is that widgets are owned by their parent widgets, so when the top level widgets are destroyed the children are also destroyed.  So that actually covers the majority of the Qt objects since most are widgets. 

Top-level widgets and non-widget objects need taken care of explicitly. Or just ignored because the process of about to disappear anyway. The only reason to really care is, in this case at least, to make the output from valgrind meaningful.

PR #1646, opened 2022-08-24
Minimize updated areas of editors for playhead movement

cme:
While playing, the song editor and pattern editors will update
periodically to show the progress of the playhead across the song
or pattern. The backgrounds are cached as pixmaps so this is a lot
faster than redrawing the whole song or pattern, but blitting the pixmap can
still take significant amounts of time when large areas of screen space
are taken up by these widgets (particularly on, say, a Raspberry
Pi attached to a 2560x1440 display). It's unclear if Qt is using
any hardware acceleration for this, but I've seen large amounts of
CPU time used in Linux and macOS profiling.

This change modifies the timer-driven updates to call update() on
only the minimal area necessary to correctly keep the widget fresh,
and erase the old playhead and render the new one.

There's an annoying amount of code duplication here, as there's 
no common ancestor class to fold the song editor geometry into.
That feels ripe for a refactor.

On the RPi4, running one of my tunes with a full screen display on the
2560x1440 display, this reduces the CPU time taken for 1000 seconds worth of
play time by quite a bit:

 |          | master time | new time | diff | %          |
 |----------|-------------|----------|------|------------|
 | hydrogen | 453         | 368      | -85  | -18.8%     |
 | Xorg     | 83          | 15       | -68  | -81.9%     |
 | Total    | 536         | 383      | -153 | -28.5%     |

(So, overall that's going from 53% of a CPU core to 38% of a CPU core)

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditor.cpp
+++ b/src/gui/src/PatternEditor/PatternEditor.cpp
@@ -27,6 +27,7 @@
 #include "../HydrogenApp.h"
 #include "../EventListener.h"
 #include "../UndoActions.h"
+#include "../Skin.h"
 
 #include <core/Globals.h>
 #include <core/Basics/Song.h>
@@ -883,8 +884,27 @@ void PatternEditor::stackedModeActivationEvent( int nValue )
 }
 
 void PatternEditor::updatePosition( float fTick ) {
+	if ( m_nTick == (int)fTick ) {
+		return;
+	}
+
+	float fDiff = m_fGridWidth * (fTick - m_nTick);
+
 	m_nTick = fTick;
-	update();
+
+	int nOffset = Skin::getPlayheadShaftOffset();
+	int nX = static_cast<int>(static_cast<float>(PatternEditor::nMargin) +
+							  static_cast<float>(m_nTick) *
+							  m_fGridWidth );
+
+	QRect updateRect( nX -2, 0, 4 + Skin::nPlayheadWidth, height() );
+	update( updateRect );
+	if ( fDiff > 1.0 || fDiff < -1.0 ) {
+		// New cursor is far enough away from the old one that the single update rect won't cover both. So
+		// update at the old location as well.
+		updateRect.translate( -fDiff, 0 );
+		update( updateRect );
+	}
 }
 
 void PatternEditor::storeNoteProperties( const Note* pNote ) {
--- a/src/gui/src/PatternEditor/PatternEditorRuler.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorRuler.cpp
@@ -122,13 +122,24 @@ void PatternEditorRuler::updatePosition( bool bForce ) {
 	int nTick = pAudioEngine->getPatternTickPosition();
 
 	if ( nTick != m_nTick || bForce ) {
+		int nDiff = m_fGridWidth * (nTick - m_nTick);
 		m_nTick = nTick;
-		update();
+		int nX = static_cast<int>( static_cast<float>(PatternEditor::nMargin) + 1 +
+								   m_nTick * static_cast<float>(m_fGridWidth) -
+								   static_cast<float>(Skin::nPlayheadWidth) / 2 );
+		QRect updateRect( nX -2, 0, 4 + Skin::nPlayheadWidth, height() );
+		update( updateRect );
+		if ( nDiff > 1 || nDiff < -1 ) {
+			// New cursor is far enough away from the old one that the single update rect won't cover both. So
+			// update at the old location as well.
+			updateRect.translate( -nDiff, 0 );
+			update( updateRect );
+		}
 
 		if ( ! bIsSelectedPatternPlaying ) {
 			nTick = -1;
 		}
-		
+
 		auto pPatternEditorPanel = HydrogenApp::get_instance()->getPatternEditorPanel();
 		if ( pPatternEditorPanel != nullptr ) {
 			pPatternEditorPanel->getDrumPatternEditor()->updatePosition( nTick );
--- a/src/gui/src/SongEditor/SongEditor.cpp
+++ b/src/gui/src/SongEditor/SongEditor.cpp
@@ -974,9 +974,24 @@ void SongEditor::updateWidget() {
 	m_previousMousePosition = m_currentMousePosition;
 }
 
+
 void SongEditor::updatePosition( float fTick ) {
-	m_fTick = fTick;
-	update();
+	if ( fTick != m_fTick ) {
+		float fDiff = static_cast<float>(m_nGridWidth) * (fTick - m_fTick);
+		m_fTick = fTick;
+		int nX = static_cast<int>( static_cast<float>(SongEditor::nMargin) + 1 +
+								   m_fTick * static_cast<float>(m_nGridWidth) -
+								   static_cast<float>(Skin::nPlayheadWidth) / 2 );
+		int nOffset = Skin::getPlayheadShaftOffset();
+		QRect updateRect( nX + nOffset -2, 0, 4, height() );
+		update( updateRect );
+		if ( fDiff > 1.0 || fDiff < -1.0 ) {
+			// New cursor is far enough away from the old one that the single update rect won't cover both. So
+			// update at the old location as well.
+			updateRect.translate( -fDiff, 0 );
+			update( updateRect );
+		}
+	}
 }
 
 void SongEditor::paintEvent( QPaintEvent *ev )
@@ -3125,10 +3140,22 @@ void SongEditorPositionRuler::updatePosition()
 	m_pAudioEngine->unlock();
 
 	if ( fTick != m_fTick ) {
+		float fDiff = static_cast<float>(m_nGridWidth) * (fTick - m_fTick);
 
 		m_fTick = fTick;
-	
-		update();
+		int nX = static_cast<int>( static_cast<float>(SongEditor::nMargin) + 1 +
+								   m_fTick * static_cast<float>(m_nGridWidth) -
+								   static_cast<float>(Skin::nPlayheadWidth) / 2 );
+
+		QRect updateRect( nX -2, 0, 4 + Skin::nPlayheadWidth, height() );
+		update( updateRect );
+		if ( fDiff > 1.0 || fDiff < -1.0 ) {
+			// New cursor is far enough away from the old one that the single update rect won't cover both. So
+			// update at the old location as well.
+			updateRect.translate( -fDiff, 0 );
+			update( updateRect );
+		}
+
 		auto pSongEditorPanel = HydrogenApp::get_instance()->getSongEditorPanel();
 		if ( pSongEditorPanel != nullptr ) {
 			pSongEditorPanel->getSongEditor()->updatePosition( fTick );


theGreatWhiteShark:
Awesome!

I hope the performance didn't suffered too much due to the GUI update. After all, SVG rendering and direct drawing of components is most probably worse in terms of performance than loading PNG files. (I actively decided against checking the change in performance there in order to not discourage myself while updating the widgets :smile: )

We could introduce a checkable "Fullsize Playhead" option in the preferences to give better performance on old (or small) machines. If unchecked, the playhead will only the shown and updated in the ruler of both song and pattern editor. Indicating whether the current selected pattern is played back could also be represented in playhead color and would work there too.

cme:
> Awesome!
> 
> I hope the performance didn't suffered too much due to the GUI update. After all, SVG rendering and direct drawing of components is most probably worse in terms of performance than loading PNG files. (I actively decided against checking the change in performance there in order to not discourage myself while updating the widgets :smile: )

It's probably worse, but I haven't compared it against anything before the check-in. Whatever the extra CPU overhead is, it's worth it! And theoretically if we find any real significant issues most of it can be changed to cached pixmaps if really necessary. I don't think it would be though. Avoiding unnecessary updates is proving to be fruitful in improving things. 

There's some more spurious redrawing in some circumstances I'm looking into but the largest single chunk of update time now seems to be drawing the LCD text elements, which isnt so trivial to do something about. 


> We could introduce a checkable "Fullsize Playhead" option in the preferences to give better performance on old (or small) machines. If unchecked, the playhead will only the shown and updated in the ruler of both song and pattern editor. Indicating whether the current selected pattern is played back could also be represented in playhead color and would work there too.

I think it should be fine even on older machines. The Pi4 is still really pretty slow in comparison to almost any hardware folks are likely to be using these days, and even then I had to push it to an edge case (full screen updating constantly) to get it to perform so badly as that.

What *might* be worth considering, though, is making the timers that drive the GUI updates adaptive so that they will update less frequently in slower hardware (and maybe a bit more smoothly on newer hardware).

Hmm... actually, yes, I like that, although it would mean I'd have to change my measurement method :D

theGreatWhiteShark:
> Whatever the extra CPU overhead is, it's worth it!

That's true. I'm quite glad that there is no more pixel art involved in creating/changing a button.

> There's some more spurious redrawing in some circumstances I'm looking into but the largest single chunk of update time now seems to be drawing the LCD text elements, which isnt so trivial to do something about.

LCDs? Hmm. I would have expected something like the PPMs  in the mixer. The only LCD we constantly update - apart from peak levels in the mixer - should be the time display in the PlayerControl. Do you see a large drop in performance when opening the mixer? If not, this might indicate that LCD update is not properly checking for actual changes before repainting the widget.

There are also quite a number of timer-based updates in the GUI that could/should be rewrote to be triggered by core events. I almost started killing the timer in the instrument list of the pattern editor 2-3 times but decided to instead not delaying 1.2 further each time.

> What might be worth considering, though, is making the timers that drive the GUI updates adaptive so that they will update less frequently in slower hardware (and maybe a bit more smoothly on newer hardware).

Yeah. That sounds really nice!

cme:
> LCDs? Hmm. I would have expected something like the PPMs in the mixer. The only LCD we constantly update - apart from peak levels in the mixer - should be the time display in the PlayerControl. Do you see a large drop in performance when opening the mixer? 

Aha, yes good point -- I did not have the mixer open. That adds a lot of time too. 

Having a quick look at the profile on macOS (where it's easier to profile with `sample` but the results are more variable than linux on the Pi4), it's dominated by the peak updates, which are dominated by the SVG rendering for the frame/background and the knob. We can fairly easily cache those to a single overlay pixmap (with an alpha channel) but not sure how much that would save unless I try it. 

> If not, this might indicate that LCD update is not properly checking for actual changes before repainting the widget.

It does check correctly, I already found that a while ago. :) I think it's mostly time-consuming because it's a large area of anti-aliased filled curve rendering; maybe there's some caching in the font rendering system that it's too large for. 

I'm going to try and get a profile build working on Linux to get a better idea. Apparently we *can* get profiling results from gprof for multithreaded processes, only we have to make sure the interval timer is set up for the thread. I'm not sure about thread safety either but we can profile one thread at a time to get round that.


> There are also quite a number of timer-based updates in the GUI that could/should be rewrote to be triggered by core events. I almost started killing the timer in the instrument list of the pattern editor 2-3 times but decided to instead not delaying 1.2 further each time.

On which note, yes, I'll do beta release process before I look at anything further ;)

theGreatWhiteShark:
> I think it's mostly time-consuming because it's a large area of anti-aliased filled curve rendering; maybe there's some caching in the font rendering system that it's too large for.

In theory we could split the time display into individual widgets for ms, s, min etc. This would decrease the update area. But it would also make - at least IMO - the time display less appealing and probably goes too far in terms of optimization effort.

> On which note, yes, I'll do beta release process before I look at anything further ;)

Hehe. But there is definitely a lot of room to argue that these are important changes that deserve to be part of 1.2 and need to be tested properly via a beta.

In addition, I just came across some flaws in the current NSM implementation and I'm working on a patch that needs to be part of the upcoming beta (it introduces another translatable string/warning popup).

cme:
> Hehe. But there is definitely a lot of room to argue that these are important changes that deserve to be part of 1.2 and need to be tested properly via a beta.
>
> In addition, I just came across some flaws in the current NSM implementation and I'm working on a patch that needs to be part of the upcoming beta (it introduces another translatable string/warning popup).

I'll hold off for now then! But also I've found some odd timing issues in current master, including transport locations being set to conflicting values from different places in audioengine.cpp (I found this by noticing multiple calls to update functions in profiling), and also song timing breaks somehow in songs with different length patterns (which I found when transcribing a complex time tune for actually playing tomorrow!)

So, I might need your help with those :D

cme:
The weird behaviour on the complex time time tune doesn't seem to be deterministic, or perhaps it's related to changing pattern length somehow. I managed to reproduce it only one other time, when starting from scratch and adding empty patterns of the right length. (7,7,7,5,7,7,78)

cme:
The other issue (which causes emission of multiple EVENT_PATTERN_CHANGED events, but might otherwise be harmless?) seems to be incrementTransportPosition and updateNoteQueue bickering about which patterns should be currently playing, eventually calling down to updatePlayingPatterns with conflicting values. For example, the call tree is something like: 

```
audioEngine_process(  1024  ) -> incrementTransportPosition(  1024 )
  incrementTransportPosition(  1024  ) -> updateTransportPosition(  2637.04 )
    updateTransportPosition( 2637.04 ) -> updateSongTransportPosition  2637.04
      updateSongTransportPosition(  2637.04 ) -> updatePlayingPatterns(  7 ) 
        EVENT_PATTERN_CHANGED
  updateNoteQueue( 1024 ) -> updateSongTransportPosition  2646
    updateSongTransportPosition(  2646 ) -> updatePlayingPatterns(  0 ) 
      EVENT_PATTERN_CHANGED
  updateNoteQueue( 1024 ) -> updateSongTransportPosition  2647

audioEngine_process(  1024  ) -> incrementTransportPosition(  1024 )
  incrementTransportPosition(  1024  ) -> updateTransportPosition(  2639.27 )
    updateTransportPosition( 2639.27 ) -> updateSongTransportPosition  2639.27
      updateSongTransportPosition(  2639.27 ) -> updatePlayingPatterns(  7 ) 
        EVENT_PATTERN_CHANGED
...
```

This doesn't seem like it should happen?

cme:
There are also peripheral issues around here that lead to the multiple EVENT_PATTERN_CHANGED events causing a bunch of CPU usage, which should probably be fixed, but fixing any one of which will reduce CPU usage.

  * When reading the event queue, multiple consecutive stateless events of the same type should *probably* be delivered to listeners only once. Not so sure this would be useful though as it *shouldn't* actually happen really.
  * When updating widget background pixmaps based on queue events, the pixmaps should just be invalidated and the actual pixmap repaint should be done from paintEvent, to avoid using CPU time when windows are hidden
  * The pattern editor shouldn't have to be updated on an EVENT_PATTERN_CHANGED anyway

cme:
> What was weird behavior you encountered? An audible glitch, erratic jump of the playhead (during the playback or while paused; in the midst of the song or close to its end/while being looped)?

Erratic playhead movement in song editor and pattern editor. No audio *glitches* as such, so downstream of the sampler is fine, but notes played at weird times (that seemed to be at least somewhat in line with where the playhead was)

theGreatWhiteShark:
> The other issue (which causes emission of multiple EVENT_PATTERN_CHANGED events, but might otherwise be harmless?) seems to be incrementTransportPosition and updateNoteQueue bickering about which patterns should be currently playing, eventually calling down to updatePlayingPatterns with conflicting values. For example, the call tree is something like:
>
> This doesn't seem like it should happen?

I have to take some time for looking more detailed in the code again but in general it works as intended (but one of the `EVENT_PATTERN_CHANGED` should maybe be supressed).

It's a bit of a long story why I ended up rewriting the code in this way. 

Within Hydrogen we have two separate playback positions. One is the actual position of the transport and the other is where notes are evaluated (transport + lookahead in order to allow onset diffusion in the context of humanization). Historically these two states were dealt with by letting them diverge over the course of the `AudioEngine_process` callback and assuring they are more or less synced when removing the lock of the AudioEngine. This made things a little bit tricky when integrating external transport sync - as with the JACK server - but things worked.

But the moment tempo markers were introduced things went south. This is because transport position is handled both in frames and in ticks internally. The lookahead even has a constant frame and a constant tick component. But as the conversion between ticks and frames is tempo dependent making all variables in sync at the end of `audioEngine_process` is getting a lot harder. In the previous implementation rolling playback was hot-fixed, mouse navigation into areas containing a different tempo marker produced unwanted jumps and bizarre behavior, and sample lengths were not adapted in case tempo changed while playing it (depending on the tempo change this can cut its playback short).

My attempt to solve this mess was to ensure variables of the AudioEngine are consistent at each point of audio process cycles. So, instead of maintaining two intersecting sets of variables, the audio engine somewhat switches state between transport position and transport position + lookahead. This was implement in such a way that we can fairly easy could support more fine grained tempo markers or even external tempo sources or arbitrary tempo ramps/curves (Although it's probably less efficient). And last but not least the audio engine is now way more easy to debug (at least in IMO) as things are consistent.

We do actually see the two states in the GUI too. Time display is determined by the transport position. But the playhead represents transport + lookahead. I tried using transport position for the latter too but it was to weird. When pausing playback one could halt the playhead before a note while already hearing it. I have to check in more detail again. But I think we only need `EVENT_PATTERN_CHANGED` queued by the transport position + lookahead part while playback is rolling.

theGreatWhiteShark:
> I'll have a look

Turns out there was a bug in the test function checking the transport consistency when changing song size (what happens when adding patterns at the end of the song). Once fixed it fails. Maybe this is already related to your observation.

(in earlier versions of Hydrogen transport always jumped during playback as soon as transport was looped at least ones. As a side effect of fixing this we allow changing pattern sizes while transport is rolling now)

theGreatWhiteShark:
> The weird behaviour on the complex time time tune doesn't seem to be deterministic, or perhaps it's related to changing pattern length somehow. I managed to reproduce it only one other time, when starting from scratch and adding empty patterns of the right length. (7,7,7,5,7,7,78)

I couldn't reproduce it either using unit tests.

But I _did see_ some erratic jumps back and forth a column at one instance. This was most likely due to the `EVENT_PATTERN_CHANGED` usage you mentioned. I took a look into the code and it's indeed quite messy and the event must not be queued when updating the transport position via `AudioEngine::incrementTransportPosition()`. But I think I will make a detour here and split off the transport and playhead position mentioned above in two separate objects and move a lot of the position stuff into the `TransportInfo` class. This will make the `AudioEngine` slightly less efficient but much more easy to understand. In addition, it will be absolutely clear whether another part of Hydrogen is interacting with the transport or the playhead position afterwards.

PR #1650, opened 2022-09-11
Lazy background pixmap updating for better GUI performance

cme:
Defer background pixmap recalculation for more GUI widgets until repaint. This helps reduce the CPU time taken if transport time jitter causes multiple pattern change events to be emitted.

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditorPanel.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorPanel.cpp
@@ -942,6 +942,13 @@ void PatternEditorPanel::patternModifiedEvent() {
 	selectedPatternChangedEvent();
 }
 
+void PatternEditorPanel::patternChangedEvent() {
+	if ( Hydrogen::get_instance()->getPatternMode() ==
+		 Song::PatternMode::Stacked ) {
+		updateEditors( true );
+	}
+}
+
 void PatternEditorPanel::songModeActivationEvent() {
 	if ( Hydrogen::get_instance()->getPatternMode() ==
 		 Song::PatternMode::Stacked ) {
--- a/src/gui/src/PatternEditor/PatternEditorPanel.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorPanel.cpp
@@ -942,10 +942,6 @@ void PatternEditorPanel::patternModifiedEvent() {
 	selectedPatternChangedEvent();
 }
 
-void PatternEditorPanel::patternChangedEvent() {
-	updateEditors( true );
-}
-
 void PatternEditorPanel::songModeActivationEvent() {
 	if ( Hydrogen::get_instance()->getPatternMode() ==
 		 Song::PatternMode::Stacked ) {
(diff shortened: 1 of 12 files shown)


src/gui/src/PatternEditor/PatternEditorPanel.cpp
```diff
@@ -942,10 +942,6 @@ void PatternEditorPanel::patternModifiedEvent() {
 	selectedPatternChangedEvent();
 }
 
-void PatternEditorPanel::patternChangedEvent() {
```
theGreatWhiteShark, line 945: When removing this call the DrumPatternEditor is not properly updated in Stacked pattern mode when activating/deactivating other pattern without changing the selected pattern. Expected behavior would be added/removed grey dots for the played notes currently not visible and grey ticks and lines in case any of the other playing patterns is longer than the currently selected one.

But I just noticed that the above representation now yields some visual artifacts. I'll check as soon as I'm finished with your AudioEngine bugs.

cme: Good catch thanks! Can you describe the other artefacts at all? Leftover bits not updated on screen? Timing related? I can try and figure out potential causes if you can describe! :)

theGreatWhiteShark: Seems like most of the lower part of each row is in an undefined state. It's black in this example but it's different every time.

patternEditorRenderingIssues

cme: Aha! This happens on master too. I guess it's not quite clear what the editors *should* look like when this happens, but they should definitely do *something* well-defined and repeatable :D And also something that's useful for showing how the patterns actually fit together -- using a dark background to match the previous behaviour would make the shaded notes from other patterns hard to spot. Perhaps a darkened / faded version of the same background pattern for the foreground pattern? And something in the pattern ruler to mark the difference too.

cme: Hmm, it looks like the glitchiness of the background was introduced with 35dc06eeab, where a `fillRect` that would otherwise set up this area to the inactive background colour is commented out.  With it reinstated we get:

image

This doesn't look *too* bad actually now that I look at it, and I'd propose we reinstate that `fillRect`, unless you have any recollection about why it was commented out (other than maybe it seemed redundant?). Or perhaps make it conditional on `m_nActiveWidth < m_nEditorWidth`.

theGreatWhiteShark: > Hmm, it looks like the glitchiness of the background was introduced with https://github.com/hydrogen-music/hydrogen/commit/35dc06eeab8dfc96f9f06e778113e0a7cc72af66, where a fillRect that would otherwise set up this area to the inactive background colour is commented out.

Ah, I see. I remember that I commented out various lines in order to find the source of some weird behavior upon resizing.  I was not careful enough when committing the changes.

 > I guess it's not quite clear what the editors should look like when this happens

I added this part of the rendering a couple of months ago. The idea was to introduce ticks along with the faint notes from other stacked patterns to the better idea of the rhythm playing. To still make the distinction of the area covered by the actual pattern and the ones used for the shadow notes clear, I draw the ticks on the default background color of the pattern editor. Exactly as in your picture above.

One point I "saved" for later and still don't like is how shadow notes are rendered right at the border between the two regions. Due to the default coloring its left part is almost invisible and additionally I kinda suggests notes can be placed at these positions as well. The latter could be dealt with by drawing faint notes when hovering the ticks, like done with the playhead in the position ruler. But I feared this might be already much dynamic rendering and annoys more than it would help.

PR #1651, opened 2022-09-11
Fixing non-functional AudioEngine test and bugs uncovered on the way

theGreatWhiteShark:
`AudioEngine::testSongSizeChange` was not executed properly and contained a number of bugs. They were fixed and it is now running smoothly again. Apart from checking the integrity of transport position and computed tick interval in `AudioEngine::testSongSizeChange()` the consistency of the column transport resides in is checked as well. After all, glitches in this variable will result in jumps of the playhead and are annoying for the user. This also uncovered two (minor) bugs in the `AudioEngine` itself.

After changing the song size `AudioEngine::computeTickInterval` needs to take the `AudioEngine::m_fTickMismatch` into account when calculating start and end position in ticks of an interval governed by the current position in frames and its length in frames. This itsy-bitsy bug was nothing noticeable but was caught by the unit tests.

There is an edge case not properly handled in `AudioEngine::updateSongSize()` yet. Whenever the current column before toggling is beyond the end of the song after toggling transport is loop to a particular position within the song. E.g. there are several empty patterns in front of a final one, playhead resides in one of the empty ones, and the users deactivates the final pattern. From the user perspective, however, it is not obvious why transport is not just wrapped to the beginning of the song but to somewhere within it. This is now fixed.

### DIFF
--- a/src/core/AudioEngine/AudioEngine.cpp
+++ b/src/core/AudioEngine/AudioEngine.cpp
@@ -1906,26 +1906,30 @@ void AudioEngine::updateSongSize() {
 
 	bool bEndOfSongReached = false;
 
-	double fNewSongSizeInTicks = static_cast<double>( pSong->lengthInTicks() );
-
-	// WARNINGLOG( QString( "[Before] frame: %1, bpm: %2, tickSize: %3, column: %4, tick: %5, mod(tick): %6, pTickPos: %7, pStartPos: %8, m_fLastTickIntervalEnd: %9, m_fSongSizeInTicks: %10" )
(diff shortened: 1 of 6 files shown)


theGreatWhiteShark:
CI tests of [43b1da7](https://github.com/hydrogen-music/hydrogen/pull/1651/commits/43b1da7458ea1936c3d829351787f38b680ea038) did caught an error resulting from an very rare edge case and I was able to reproduce it just by looking at the AppVeyor log. The test function of the `AudioEngine` start to pay off. :smile:

PR #1667, opened 2022-10-29
Fix ADSR application and refactoring

theGreatWhiteShark:
Fixes the `ADSR` related bugs discussed in the comments of #1503.

It also introduces some minor refactoring of the `Sampler` and a bigger one for the `ADSR` class.


cme (commented):
Awesome!

src/core/Sampler/Sampler.cpp
```diff
@@ -617,53 +641,44 @@ bool Sampler::renderNote( Note* pNote, unsigned nBufferSize, std::shared_ptr<Son
...
 		}
 
 		// direct track outputs only use velocity
-		if ( Preferences::get_instance()->m_JackTrackOutputMode == Preferences::JackTrackOutputMode::preFader ) {
-			cost_track_L = cost_track_L * pNote->get_velocity();
-			cost_track_L = cost_track_L * fLayerGain;
-			cost_track_R = cost_track_L;
+		if ( Preferences::get_instance()->m_JackTrackOutputMode ==
+			 Preferences::JackTrackOutputMode::preFader ) {
+			fCostTrack_L *= pNote->get_velocity();
```
theGreatWhiteShark, line 679: Refactoring the gain contributions made more obvious that note velocity is always applied in per track outputs in PreFader mode  but only applied in PostFader mode in case the "apply velocity" checkbox in the Instrument Editor was checked. This does not seem right.

theGreatWhiteShark: In addition, PreFader gain does not take the note pan into account.

PR #1681, opened 2022-12-04
Final 1.2.0-beta fixes

theGreatWhiteShark:
Some final fixes to strike out from my local TODO list. Well, it's a never ending story but from my point of view we are ready to do the 1.2.0-beta with these ones merged.

Most notably: 
- Update of license notices
- missing PatternEditor updates when switching states (especially while being in stacked mode)
- fix playhead repositioning at the end of the song
- fix virtual patterns handling (thanks for the hint @oddtime ! ).

    Turns out these did never worked properly as soon as one of the patterns contained in the virtual one was longer than the virtual one itself. This was not accounted for when calculating the longest pattern in a column and the length of the song. While at it I activated the new stacked mode view in the pattern editor when working with virtual patterns.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 
 # Hydrogen
 # Copyright(c) 2002-2008 by Alex >Comix< Cominu [comix@users.sourceforge.net]
-# Copyright(c) 2008-2021 The hydrogen development team [hydrogen-devel@lists.sourceforge.net]
+# Copyright(c) 2008-2022 The hydrogen development team [hydrogen-devel@lists.sourceforge.net]
 #
 # http://www.hydrogen-music.org
 #
--- a/macos/build_dmg.sh
+++ b/macos/build_dmg.sh
@@ -2,7 +2,7 @@
 
 # Hydrogen
 # Copyright(c) 2002-2008 by Alex >Comix< Cominu [comix@users.sourceforge.net]
-# Copyright(c) 2008-2021 The hydrogen development team [hydrogen-devel@lists.sourceforge.net]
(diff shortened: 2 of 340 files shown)


theGreatWhiteShark:
> It's about the separate widgets for the numerator/denominator and the problem when one wants to extend the length of an already programmed pattern (e.g. from 4/4 to 17/16) without loosing the notes as they are canceled for the fact that you need to change the denominator first and then the numerator (or similar issues).

Yeah, I'm not that confident with those change anymore either. Being able to recreate the notes lost using undo is fine in case you realize your loss right away. But when noticing it after half an hour you have to either revert all your recent work or bit the bullet and loose parts of your pattern. In addition, you have to realize it within the same session and in case a bug messes with the undo history these changes will be gone too.

I just checked with MusE and there you can resize all note containers without loosing any notes. We should probably go for this too again (while avoiding the original bug that caused us to change it).

Nevertheless, I think this fix can be merged between v1.2-beta and v1.2. Not quite sure whether I find time to fix this during the weekend and the 1.2 release is overdue for almost a _year_ now. Still green light for making the release on my side.

theGreatWhiteShark:
> In my opinion there is one thing to be changed before the beta release.
It was discussed by cme if I am right, but I don't remember where.
>
> It's about the separate widgets for the numerator/denominator and the problem when one wants to extend the length of an already programmed pattern (e.g. from 4/4 to 17/16) without loosing the notes as they are canceled for the fact that you need to change the denominator first and then the numerator (or similar issues).

I addressed this one in #1717

PR #1696, opened 2022-12-22
Logger: flush stdout

theGreatWhiteShark:
Since recently (maybe because of 7f26902da0e12f88cf3b15fef3c25f01020a8e48) our logger lags quite a bit and gets stuck when printing to stdout. Flushing stdout fixes this issue. Not quite sure why. It was not flushed beforehand and did work.


cme:
Odd. The default for `stdout` should be line-buffering if it `isatty`. So I guess that *does* look like it needs a flush if you're observing it through a pipe or a file, which isn't really unreasonable, so a flush seems reasonable!

PR #1699, opened 2022-12-28
Pattern and instrument drag & drop preview

cme:
Preview drag and drop operations in the `SongEditorPatternList` and `PatternEditorInstrumentList`, rearranging the patterns (or instruments) in the list to the order which will be imposed if the drag is ended at that position.

(For imports, an empy space is shown, since no details are available until they're loaded)

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditorInstrumentList.cpp
+++ b/src/gui/src/PatternEditor/PatternEditorInstrumentList.cpp
@@ -876,8 +876,83 @@ void PatternEditorInstrumentList::dragEnterEvent(QDragEnterEvent *event)
 	event->acceptProposedAction();
 }
 
+void PatternEditorInstrumentList::dragMoveEvent(QDragMoveEvent *event)
+{
+	if ( ! event->mimeData()->hasFormat("text/plain") ) {
+		event->ignore();
+	}
+
+	QString sText = event->mimeData()->text();
+	if ( sText.startsWith( "move instrument:" ) ) {
+		Hydrogen *pHydrogen = Hydrogen::get_instance();
+		int nSourceInstrument = pHydrogen->getSelectedInstrumentNumber();
+		int nTargetInstrument = event->pos().y() / m_nGridHeight;
+
+		// Move InstrumentLines temporarily into new place
+		setDropTarget( DropTargetKind::Move, nTargetInstrument, nSourceInstrument );
+
+	} else if ( sText.startsWith( "importInstrument:" ) ) {
+		// Move existing InstrumentLines to show the gap where the new instrument will be inserted.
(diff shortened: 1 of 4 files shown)


cme:
Tempted to refactor DropTargetKind and associated parameters to a new class...

theGreatWhiteShark:
> (For imports, an empy space is shown, since no details are available until they're loaded)

Nowadays all drumkits accessible in the Sound Library widget are loaded (without samples) and stored in the [SoundLibraryDatabase](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/SoundLibrary/SoundLibraryDatabase.h). The absolute path to the kits folder serves as the key to access them. [SoundLibraryPanel::m_drumkitRegister](https://github.com/hydrogen-music/hydrogen/blob/master/src/gui/src/SoundLibrary/SoundLibraryPanel.h#L109) maps the labels shown in the Sound Library widgets to the keys of the SoundLibraryDatabase. E.g. https://github.com/hydrogen-music/hydrogen/blob/master/src/gui/src/SoundLibrary/SoundLibraryPanel.cpp#L478

theGreatWhiteShark:
I like it. But I have some remarks. 
- it feels a little strange that the associated row along with its notes/patterns is not moving as well
- the dragged item in the instrument list keeps the "selected" color but its counterpart in the pattern list does not
- maybe inserting and moving could have two different cursor icons. I haven't looked into to the drag and drop support of Qt5 yet but there are `Qt::DragMoveCursor` and `Qt::DragCopyCursor` in the [docs](https://doc.qt.io/qt-5/qt.html#CursorShape-enum). But for some reasons those icons aren't displayed.

theGreatWhiteShark:
> it feels a little strange that the associated row along with its notes/patterns is not moving as well

That's probably a more complex thing to implement.

I just thought about it a little more. I like your solution and I definitely don't want to step in and tell what it should look like. Just a thought :).  In other applications drag move and drag additions are rendered by highlighting the border between two columns. This would probably be way less work then moving notes around. Also, we should think in advance about how to display a drag replacement of an instrument with another one (it's on the TODO list). Maybe something like highlighting its borders similar to what is happening when right-clicking a instrument/pattern element.

cme:
I've sort of sat on this one for a while because, while I liked the initial idea, when it came to actually *using* it for some aggressive instrument reshuffling, I didn't like the way it *felt*. It seemed too disorienting, and reflecting about it I think this was because it shows only the "after" result of the action, and the fact that instruments and patterns move only in fixed amounts doesn't actually convey much "movement" when the cursor moves to a different block.

Maybe this could be fixed by **highlighting** the inserted/moved pattern or instrument as well, but on reflection I think showing a highlight at the drop site is actually *better*.

theGreatWhiteShark:
Superseded by #2243 and #2242

PR #1702, opened 2023-01-17
Annual timestamp update

theGreatWhiteShark:
- fix timestamps in MIDI test files causing test failures
- updating copyright signature


theGreatWhiteShark:
Merged without passing the build pipeline because current the Windows one is broken and its fix requires this PR to pass all tests

Issue #1694, opened 2022-12-18
wrong translation tips?


trebmuh:
https://github.com/hydrogen-music/hydrogen/blob/769ee9d3844dbd72448b18c36523741a6ee30b8c/src/gui/src/CommonStrings.cpp#L52

shouldn't that be "eight" and not "three" ?

https://github.com/hydrogen-music/hydrogen/blob/769ee9d3844dbd72448b18c36523741a6ee30b8c/src/gui/src/CommonStrings.cpp#L55

shouldn't that be "two" and not "eight" ?

theGreatWhiteShark:
Yes, looks like a classic copy-paste error. Were those translation tips useful?

trebmuh:
> Were those translation tips useful?

Definitely, I would encourage devs to add this type of comments more often.

trebmuh:
@theGreatWhiteShark did you add the "invalid" tag on purpose?

theGreatWhiteShark:
> Definitely, I would encourage devs to add this type of comments more often.

Awesome. I introduced a dedicated class bundling translatable strings. Hopefully, this will ease translating in the future.

> did you add the "invalid" tag on purpose?

Yeah. As those string are invalid :)

trebmuh:
:+1:

trebmuh:
:+1:

PR #1710, opened 2023-02-01
Fix translation tips and update German translation

theGreatWhiteShark:
In addition, I updated all .ts files so translators do not work on old and incomplete versions (again).

- our `updateTranslations.py` script does not like "//~" comments and prints a warning message for _every_ translation file and comment occurrence. I changed those into "// ~" in order to make the scripts output readable
- the update of each and every .ts file in `updateTranslations.py` was hard-coded and the most recent one, zh_CN, was missing. The script does now run the update on all files in the i18n folder holding the suffix ".ts"

fixes https://github.com/hydrogen-music/hydrogen/issues/1694

### DIFF
--- a/data/i18n/hydrogen_fr.ts
+++ b/data/i18n/hydrogen_fr.ts
@@ -347,7 +347,7 @@ Text displayed on the button indicating that the Beat Counter will only set temp
     </message>
     <message>
         <source>Edit</source>
-        <extracomment>Text displayed on the button for editin an element. Its size is designed for a four characters.</extracomment>
+        <extracomment>Text displayed on the button for editing an element. Its size is designed for a four characters.</extracomment>
         <translation>Édit</translation>
     </message>
     <message>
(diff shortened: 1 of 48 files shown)


data/i18n/hydrogen_fr.ts
```diff
@@ -768,7 +768,7 @@ B</translation>
     </message>
     <message>
         <source>The Timeline is only available in Song Mode</source>
-        <translation>La base de temps est uniquement disponible en mode morceau></translation>
+        <translation>La base de temps est uniquement disponible en mode morceau&gt;</translation>
```
trebmuh, line 771: the "morceau>" should be "morceau" (my bad probably)
@theGreatWhiteShark : could you please remove the &gt; here? Alternatively, I'll try to remember this and remove it after it'll be merged

trebmuh (changes requested):

data/i18n/hydrogen_fr.ts
theGreatWhiteShark: Sure

trebmuh:
thanks :+1:

Issue #1711, opened 2023-02-02
Insufficient feedback when exporting/saving to read-only directory


theGreatWhiteShark:
Turns out when attempting to export a song as audio file into a folder the current use has only read access for GUI just fails silently. The only thing that comes up in the error log is

> (E) ::void* H2Core::diskWriterDriver_thread(void*) Error during sf_write_float

That's not much and does not even hint what is going on.

Same thing when 
- exporting MIDI files
- exporting LiliPond files

When exporting patterns or drumkits as well as saving songs there is at least a dialog telling the user that something went wrong. But not what.

I think we can do better, e.g. check explicitly for write-permissions and display a dedicated error dialog in case it fails.

PR #1712, opened 2023-02-02
Checking write permissions when exporting or saving files

theGreatWhiteShark:
Whenever a song, drumkit, MIDI, audio file etc. is now saved or exported, the custom QFileDialog class will check whether the user has write permissions for the selected folder. This is done when pressing the "Export", "Save" etc. button and a warning dialog will popup asking the user to select a different folder without closing the overall file dialog

fixes https://github.com/hydrogen-music/hydrogen/issues/1711


theGreatWhiteShark:
Hey @trebmuh ,

I know I promised a string-freeze. But this patch required yet another translatable string :grimacing:

PR #1714, opened 2023-02-05
rtclock: fix implicit sleep declaration

nekopsykose:
otherwise clang16 fails to compile this code, and detects this as false


theGreatWhiteShark (approved):
Thanks!

PR #1740, opened 2023-04-09
Object: include sys/time.h

nekopsykose:
this is needed for `timeval`

### DIFF
--- a/src/core/Object.h
+++ b/src/core/Object.h
@@ -27,6 +27,9 @@
 #include <core/config.h>
 #include "core/Globals.h"
 
+#ifndef _WIN32
+#include <sys/time.h>
+#endif
 #include <unistd.h>
 #include <iostream>
 #include <atomic>


theGreatWhiteShark:
Hey @nekopsykose ,

Is this missing include causing trouble in your OS + compiler setup? What compiler in which version are you using?

In our Linux build pipeline as well as locally it works fine. But your patch is not working for the Windows build. Maybe I can reproduce the issue with the specific compiler you are using.

nekopsykose:
sure, it's needed when building against musl:  

```
/home/demon/src/aports/community/hydrogen/src/hydrogen-1.2.0/src/core/Object.h:153:16: error: 'timeval' does not name a type
  153 |         static timeval __last_clock;
```

windows indeed doesn't have this header though, so amended with the correct guard

theGreatWhiteShark:
Out of curiosity: you are using an underscore in `_WIN32`. Do you know where this comes from? `cmake` [sets](https://cmake.org/cmake/help/latest/variable/WIN32.html?highlight=win32) `WIN32`

nekopsykose:
i don't have a very precise reference, so i suppose https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c is the best info that describes the difference:  

- _WIN32 is the actual platform, so the compiler implicitly defines this (and i think so do microsoft headers?)
- WIN32 is a build system define, so i guess cmake defines it a bit differently (in addition to the above). iirc mingw also defines the same.

in every case where the latter is defined though, the former is too.

nekopsykose:
you might find microsoft's official docs useful too: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170  

which state _WIN32 always:  

> _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.

and also a _WIN64 for only 64-targets:

> _WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined.  

i am not sure the 'otherwise' is ever true for the _WIN32 one.

nekopsykose:
and the actual header reference is https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/time.h.html

theGreatWhiteShark (approved):

theGreatWhiteShark:
Thanks a lot!

PR #1743, opened 2023-04-12
AppVeyor: tweak artifacts upload strategy

theGreatWhiteShark:
I'm still not a 100% sure why but somehow I lost a couple of commits on the `releases/1.2` . In the end I managed to have the 1.2.0 tag at the HEAD of `releases/1.2-artifacts` but the `ARTIFACTS_BRANCH` in `.appveyor.yml` was reset to `master-artifacts`. That was a little bit frustrating as I have no idea how to put a commit altering the branch _in front_ of the tagged commit. In the end I put artifacts build with the some sources as the tag but from a different commit inside the release.

To ensure this will not happen again I did revise the artifact upload in the `.appveyor.yml`. (and I will try to reproduce what went wrong in another repo)

Instead of only building with artifacts in case the branch name matches a certain string specified in the `.appveyor.yml` file, artifact upload is now triggered whenever branch name has a "-artifacts" suffix or in case HEAD is at a tag. No editing of branch names in configs anymore and adding a tag on `master` does already trigger artifact generation. For the latter I'm not sure if it already works properly. I just simulated adding a tag by setting environmental variables according to the  AppVeyor Doc.

@cme As we only push sparsely to release branches we could also whitelist all branches with "releases/" prefix and offer rolling releases in addition to our regular ones.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -78,11 +78,6 @@ environment:
 build:
   verbosity: detailed
 
-matrix:
-  exclude:
-    - job_name: Windows32
-      branches: "master"
-
 # Upload artifacts in case either the branch is named *-artifacts or
 # HEAD is located on a tag (affects both Linux and Mac OS jobs).
 init:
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
By now I also reconstructed why I did need a force push to `releases/1.2` during the 1.2.0 releases. I did merge `master` into `releases/1.2` and `releases/1.2-artifacts` and created an annotated tag on `master`. But as I did not merged the release branch back into `master`, the two's HEADs were at different commits and there was no way to fast-forward merge the tag to the release branch in order to keep it at HEAD. I also was not aware that in contrast to all other merges one has to explicitly use `git merge -ff` with annotated tags as their default behavior is always `--no-ff`.

Anyway, with this patch there won't be a need for any force push to "official" branches anymore. Even in case the tag is "covered" in a merge commit one can easily check it out and push it to a new auxiliary branch in order to build corresponding artifacts.

PR #1775, opened 2023-05-17
MidiCommon: merge SYSTEM_EXCLUSIVE into SYSEX

### LINKED ISSUE #1773: Sysex MIDI messages in PortMidi
Operating system + version  :  Windows
Audio driver + version      :  PortMidi

Determining the actual type of an incoming MIDI message in [CoreMidiDriver::midiProc](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/IO/CoreMidiDriver.cpp#L43) and [PortMidiDriver::PortMidiDriver_thread](https://github.com/hydrogen-music/hydrogen/blob/master/src/core/IO/PortMidiDriver.cpp#L50) are both done by the first byte of the incoming message. The **only** difference is that CoreMidiDriver assigns type `MidiMessage::SYSEX` for a certain range of values and PortMidiDriver assigns type `MidiMessage::SYSTEM_EXCLUSIVE`.

But the result is totally different since `MidiMessage::SYSEX` is used for incoming MMC_x events in the `MidiMap` while `MidiMessage::SYSTEM_EXCLUSIVE` are explicitly discarded.

SYSEX and SYSTEM_EXCLUSIVE are treated as synonyms in all references I came across during a short search.

It feels a lot like this was done on accident but, unfortunately, it is there since the beginning of our git history.

theGreatWhiteShark:
After reading into both MIDI specs and doc as well as PortMidi API doc I came to the conclusion that the separate handling (and mere existence) of `SYSTEM_EXCLUSIVE` and `SYSEX` MIDI type must have been an accident.

This fix should enable MMC MIDI control using PortMidi (I have no hardware to test this).

Fixes #1773

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,8 @@ XXXX-XX-XX the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 	* Release 1.2.1
 	* Update French translation
 	* Bugfixes
+		- Fix MIDI Machine Control (MMC) event type handling on Windows
+		  (#1773)
 		- Fix loading of legacy drumkits. All layers but the first one
 		  were dropped during drumkit upgrade (#1759)
 		- Fix MIDI input handling with "Discard MIDI messages after action
--- a/src/core/IO/MidiCommon.h
+++ b/src/core/IO/MidiCommon.h
@@ -47,7 +47,6 @@ public:
 		PROGRAM_CHANGE,
 		CHANNEL_PRESSURE,
 		PITCH_WHEEL,
-		SYSTEM_EXCLUSIVE,
 		START,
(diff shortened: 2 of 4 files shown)


theGreatWhiteShark:
Note: this does not enable MMC message handling in PortMidi yet. Collecting the data corresponding to a SysEx message is not implemented yet. (Probably that's why `SYSTEM_EXCLUSIVE` was introduced to distinct them from `SYSEX` ones in macOS and Linux, which are handled properly. The reason implementation was skipped is most likely because it is not straight-forward as the data has to be collected from multiple `PmEvent`s with arbitrary amounts of Real-Time MIDI messages being mixed into it.)

PR #1806, opened 2023-06-30
Es trans

ignotus666:
Updated Spanish translation. Not sure where you are in the release cycle but I took the chance to use some spare time ;)


theGreatWhiteShark (approved):

theGreatWhiteShark:
Thanks a lot!

> Not sure where you are in the release cycle

We are more or less at the beginning of the development phase for v1.3. But I will take care porting your translations back to v1.2 in order to integrate them in a (potential) next patch release.

I also have to admit that there are quite a number of new translations which are not reflected in the .ts files yet. Shame on me. I will commit them ASAP. In addition, I will come up with a way to automate the translation update in order to always have the latest .ts files on `master`.

PR #1813, opened 2023-07-14
Update French translation

trebmuh:


theGreatWhiteShark (approved):

theGreatWhiteShark:
Thanks!

There is still one issue I just got aware of: the upcoming versions of Hydrogen will feature a new tab within the Preferences using which all shortcuts can be configured. Including those for quick access of menu options. This, however, uncouples Qt's default shortcuts created by putting `&amp;` before a character of the option's translatable name. ("&New" allows the option to be triggered with <kbd>Alt</kbd> + <kbd>N</kbd>)

I will take care of removing the amperands and fixing the affected translations.

But in case you which for the French version of Hydrogen to use different shortcuts per default, we have to explicitly define them. It can done in here https://github.com/hydrogen-music/hydrogen/blob/master/src/core/Preferences/Shortcuts.cpp#L148 and I already do so for the virtual keyboard. The shortcuts in the linked functions will be set when starting Hydrogen without a `hydrogen.conf` file being present or when resetting all shortcuts. Hydrogen will check the system's locale to determine which shortcuts to use.

theGreatWhiteShark:
> This, however, uncouples Qt's default shortcuts created by putting &amp; before a character of the option's translatable name. ("&New" allows the option to be triggered with Alt + N)

Sorry. My bad. The shortcuts in the Preferences and those defined by ampersands are entirely different things. The former triggers an action immediately and the latter provides quick access to menu options. There is no problem in keeping them next to eachother.

PR #1816, opened 2023-07-16
Fix app-id to match and remove the old appdata file.

hfiguiere:
- Also install it
- Remove external releases. This does work with most tooling.
- Use oars-1.1

~~Note that there are other issues, notable the use of `external` releases that isn't really supported that aren't addressed.~~


theGreatWhiteShark:
AppStream [1] standard states that 

> it is strongly encouraged to only use lowercase letters for the ID

and

> Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.

For the latter it is just unfortunate that the proper ID wasn't used in the first place, as we have ownership of `hydrogen-music.org`, not `hydrogenmusic.org`.  But changing it would probably cause too much trouble.

Fixing the capitalized `.Hydrogen.*` to `.hydrogen.*`, on the other hand, is probably way less troublesome. After all it is still in the same namespace, the change suggests it is still the same application, and in case `.Hydrogen.` and `.hydrogen.` files are installed on the same system, the `releases` section clearly points out which is the most recent one.

> Remove external releases. This does work with most tooling.

Ah, okay. There was not a single note about this feature being not widely supported and being introduced recently in the docs but viewing them using the waybackmachine showed that they indeed were. Well, classic freedesktop.


[1] https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

linux/po/appstream.pot
```diff
@@ -10,402 +10,57 @@ msgstr ""
...
 msgid "High quality time-stretching and pitch shifting"
 msgstr ""
 
 #. (itstool) path: ul/li
-#: org.hydrogenmusic.hydrogen.metainfo.xml.in:34
+#: org.hydrogenmusic.Hydrogen.metainfo.xml.in:34
 msgid "Automate it using MIDI or OSC API"
 msgstr ""
 
-#. (itstool) path: description/p
```
theGreatWhiteShark, line 67: The merging of the metainfo and releases file broken some parts of the internationalization toolchain. The `linux/po/appstream-metainfo.its` file has to be adopt as well. Otherwise large parts of the existing translations will be lost.

theGreatWhiteShark (changes requested):

hfiguiere:
> AppStream [1] standard states that
> 
> > it is strongly encouraged to only use lowercase letters for the ID
> 

That's not a requirement. Existing prevail. 
Also the last component is supposed to be the name of the app, it can be written the same way.


> and
> 
> > Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.
> 
> For the latter it is just unfortunate that the proper ID wasn't used in the first place, as we have ownership of `hydrogen-music.org`, not `hydrogenmusic.org`. But changing it would probably cause too much trouble.

Mistakes where made sadly. But given that this wasn't changed in your original patch...

> Fixing the capitalized `.Hydrogen.*` to `.hydrogen.*`, on the other hand, is probably way less troublesome.

Actually it's more troublesome.

> After all it is still in the same namespace, the change suggests it is still the same application, and in case `.Hydrogen.` and `.hydrogen.` files are installed on the same system, 

Which mean the flatpak must be resubmitted and because it's a case change, it's even more of a complicated process thanks to github.

linux/po/appstream.pot
hfiguiere: I don't see where this need changing....

hfiguiere: nevermind. I also had to fix the makefile to have a proper dependency.

theGreatWhiteShark (approved):

theGreatWhiteShark:
Thanks a lot for your input! Are you fine with the changes I applied?

Issue #1820, opened 2023-07-27
Failing sample export test in macOS pipeline


theGreatWhiteShark:
Since quite a while we have sporadic failures in the macOS pipeline due to the sample export test hanging and the build being terminated because of a timeout.

After introducing a number of debug messages (#1766, #1790, #1814 among others) the problem was narrowed down to: the separate thread of the `DiskWriterDriver` is not able to acquire the lock for the audio engine (for more than 50 minutes!) and the unit test waits for the `EVENT_PROGRESS` of value `100` indefinitely.

Now, there are two separate things that need to be addressed here:
1. The thread of the `DiskWriterDriver` needs to be able to indicate failure, like not being able to lock the audio engine, being unable to open/write to file etc.
2. The (what looks like a) deadlock in the `AudioEngine` has to be tracked down. No idea yet where this could be coming from.

theGreatWhiteShark:
Turns out the deadlock was caused because the `DiskWriterDriver` was stopped while still exporting a song (see #1826).

This lead to the question: why didn't Hydrogen wait long enough for the export to finish?

PR #1832, opened 2023-08-12
debug: add logs to debug #1820

theGreatWhiteShark:
it seems the `DiskWriterDriver` is getting killed in the macOS pipeline _before_ the song in `TransportTests::testSampleConsistency` was fully exported. I added some debug logs that might help determining why this is the case

### DIFF
--- a/src/core/Hydrogen.cpp
+++ b/src/core/Hydrogen.cpp
@@ -662,6 +662,7 @@ void Hydrogen::restartDrivers()
 
 bool Hydrogen::startExportSession( int nSampleRate, int nSampleDepth )
 {
+	DEBUGLOG( "" );
 	AudioEngine* pAudioEngine = m_pAudioEngine;
 	
 	if ( pAudioEngine->getState() == AudioEngine::State::Playing ) {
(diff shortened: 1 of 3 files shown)


cme:
Would it be helpful to have access to a mac vm so you can run stuff like this at will rather than having to use Appveyor? I can sort you with one if it'd help.

theGreatWhiteShark:
> Would it be helpful to have access to a mac vm so you can run stuff like this at will rather than having to use Appveyor? I can sort you with one if it'd help.

Well, just for #1820 the current workflow is fine. At least as long as it feels like a bug in the macOS pipeline/unit tests rather than an actual Hydrogen bug. The priority is pretty low and I just tweak some log messages now and then in case the build fails.

But in general this might be a good idea as I just have Linux device and an old Windows workstation at hand. I would ping you when I'm in the need of it (or it turns out there are indeed problems with the AudioEngine on macOS).

TBH I'm kinda saturated with new technologies right now after looking into Flatpak and AppImage for the past weeks. I need to do some developing I intended to do for almost a year first :laughing:  . Also, I do not want to stall the 1.2.2 patch release (which will be the test balloon for a Hydrogen.AppImage release generate in our build pipeline).

PR #1868, opened 2023-10-06
Sampler: fix note length calculation

### LINKED ISSUE #1852: Hydrogen crashes with "Can not reroute empty drumkit paths"
Hydrogen version *          : 1.2.2
Operating system + version  : Ubuntu 23.04
Audio driver + version      : Jack

Error message:

```
(E) Filesystem::rerouteDrumkitPath Can not reroute empty drumkit paths
(E) ::void handleFatalSignal(int) Fatal signal 11
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/AppRun.wrapped() [0x68636a]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libc.so.6(+0x3c4b0) [0x7fd60743c4b0]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core4ADSR9applyADSREPfS1_iif+0x6b1) [0x7fd60a9eaf7d]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core7Sampler20renderNoteNoResampleESt10shared_ptrINS_6SampleEEPNS_4NoteES1_INS_17SelectedLayerInfoEES1_INS_19InstrumentComponentEES1_INS_16DrumkitComponentEEiiffff+0x674) [0x7fd60ab59848]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core7Sampler10renderNoteEPNS_4NoteEj+0x17cb) [0x7fd60ab576f9]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core7Sampler7processEj+0x704) [0x7fd60ab5473c]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core11AudioEngine12processAudioEj+0x104) [0x7fd60a9b178e]
(E) ::void handleFatalSignal(int) /tmp/.mount_HydrogrEfY4a/usr/bin/../lib/libhydrogen-core-1.2.2.so(_ZN6H2Core11AudioEngine19audioEngine_processEjPv+0x7d9) [0x7fd60a9b1487]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libjack.so.0(+0x14a0d) [0x7fd60af6ca0d]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libjack.so.0(+0x146d8) [0x7fd60af6c6d8]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libjack.so.0(+0x31650) [0x7fd60af89650]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libc.so.6(+0x8f18a) [0x7fd60748f18a]
(E) ::void handleFatalSignal(int) /lib/x86_64-linux-gnu/libc.so.6(+0x11dbd0) [0x7fd60751dbd0]
```

My song plays fine on Hydrogen 1.1.1. The problem has been for all versions after that (though I'm not sure whether the error message has been the same).

I attach my song [attack.txt](https://github.com/hydrogen-music/hydrogen/files/12774288/attack.txt) (changed to `.txt` so that GitHub would accept it). I've done manual edits to the drumkit XML, and it's possible that I've screwed something up.

Unfortunately, I can't include the samples. They add up to 147 MB and contain copyrighted samples.

theGreatWhiteShark:
there was a bug in the calculation of user defined note length in the `Sampler`. The delay introduced due to humanization, like lead lag, is given in frames but was mixed with variables given in ticks. This resulted in absurdly large delays and negative note lengths.

Fixes #1852

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,7 @@ XXXX-XX-XX the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 				  copied into the new one.
 		- Audio Engine: In Song Mode with Loop Mode deactivated Hydrogen
 			missed notes very close to the end of the song.
+		- Fix crash on playing back notes with custom length (#1852).
 		- Pattern Editor (#1859):
 				- Only delete NoteOff notes clicked by the user.
 				- Proper undo of moving notes out of DrumPatternEditor.
--- a/src/core/Sampler/Sampler.cpp
+++ b/src/core/Sampler/Sampler.cpp
@@ -969,8 +969,8 @@ bool Sampler::renderNoteNoResample(
 		
 		double fTickMismatch;
 		nNoteLength = TransportPosition::computeFrameFromTick(
-			pNote->get_position() + nDelay + pNote->get_length(),
-			&fTickMismatch ) - pNote->getNoteStart();
+			pNote->get_position() + pNote->get_length(), &fTickMismatch ) +
+			nDelay - pNote->getNoteStart();
 	}
 	
 	// The number of frames of the sample left to process.
@@ -1171,21 +1171,14 @@ bool Sampler::renderNoteResample(
 		// PatternEditor. This will be used instead of the full sample
 		// length.
 		double fTickMismatch;
-
-		// Delay is already introduced into the note start used below
-		// in Note::computeNoteStart. We need to account for it in
-		// here to in order to get the length of the note right.
-		const int nDelay = std::clamp(
-			pNote->get_humanize_delay(), -1 * AudioEngine::nMaxTimeHumanize,
-			AudioEngine::nMaxTimeHumanize );
 		
 		nNoteLength = 
-			TransportPosition::computeFrameFromTick( pNote->get_position() + nDelay +
-													 pNote->get_length(), &fTickMismatch,
-													 pSample->get_sample_rate() ) -
-			TransportPosition::computeFrameFromTick( pNote->get_position() + nDelay,
-													 &fTickMismatch,
-													 pSample->get_sample_rate() );
+			TransportPosition::computeFrameFromTick(
+				pNote->get_position() + pNote->get_length(),
+				&fTickMismatch, pSample->get_sample_rate() ) -
+			TransportPosition::computeFrameFromTick(
+				pNote->get_position(), &fTickMismatch,
+				pSample->get_sample_rate() );
 	}
 
 	float fNotePitch = pNote->get_total_pitch() + fLayerPitch;


theGreatWhiteShark:
While the crash of #1852 does not occur on my machine anymore, there is still something off.

1. I still get occasional negative `nReleasesFrames` passed to `ADSR::applyADSR` because `pSelectedLayerInfo->SamplePosition` is already _ahead_ of the note length before the ADSR is applied. This doesn't sound right
2. There is still a conceptional flaw. When changing tempo or passing a tempo marker the length of a note with user-defined length does change. This is because the latter is defined (in the PatternEditor) in ticks and rendering is done based on frames. Since each tempo corresponds to a different tick size (tick - frame conversion), it does also correspond to a different length in frames. 

But right now only the length of the note itself is scaled, not the **sample position** (the point till which the sample was already rendered). This means when we have a note which is 55% rendered and tempo is increased dramatically effectively cutting the note length in half, the next time the `Sampler` tries to render this note it will false assume it is fully rendered. After all the untouched sample position is beyond the freshly updated end of the note.

I will have a look at these in another PR. Depending on how much and which code needs to be touched I might also put the fix in 1.3. only.

cme:
Ooft yes, as soon as the note's played it should no longer care about the tempo I guess. Since it's still possible in the meantime, I've added #1870 to make the ADSR tolerant to the possibility by issuing a warning and moving directly to Release phase. Better than trampling past the buffer!

theGreatWhiteShark:
> Ooft yes, as soon as the note's played it should no longer care about the tempo I guess.

This could lead to audible glitches and odd behavior when changing tempo or toggling the timeline. E.g. playing two notes with the same user-defined length in a row and manually increasing tempo by a margin might result in the second note to finish before the first one.

But this is just required for tempo changes done using the BPM widget, OSC/MIDI actions, or adding of Tempo Marker. All changes applied through the (fixed) timeline are known in advanced and are covered within `computeFrameFromTick`.

PR #1870, opened 2023-10-07
Tolerate negative length notes (impossibly early release)

cme:
The AudioEngine could (and still can in rare circumstances) calculate the length of a note as negative. Previously this would have caused the ADSR implementation to write past the beginning of the buffer as pre-Release states would assume that Release must be in the future.

Instead, detect an early Release point and move directly to the Release state.


theGreatWhiteShark:
> All checks have failed

I did also see this locally. `ADSRTest::testBufferChunks()` seems to pass negative parameters at the end of its loop too.

cme:
> > All checks have failed
> 
> I did also see this locally. `ADSRTest::testBufferChunks()` seems to pass negative parameters at the end of its loop too.

The test can also enter in 'idle' state, which I didn't account for!

PR #1872, opened 2023-10-07
Sampler: account for note length during rendering

theGreatWhiteShark:
it seems the custom note length defined by the user never made it into the `Sampler`. At least properly.

Previously, the remaining frames of a samples were always calculated from the original length of the sample. Not the user defined one. The original length was also the one added to the sample position of the selected layer being rendered. At the end of the user defined length this caused the sample position of the selected layer to be _beyond_ the note length and the note end provided to `ADSR::applyADSR()` to be negative. The latter than ended rendering.

### DIFF
--- a/src/core/Sampler/Sampler.cpp
+++ b/src/core/Sampler/Sampler.cpp
@@ -954,8 +954,12 @@ bool Sampler::renderNoteNoResample(
 	auto pInstrument = pNote->get_instrument();
 	bool bRetValue = true; // the note is ended
 
-	int nNoteLength = -1;
-	if ( pNote->get_length() != -1 ) {
+	int nNoteLength;
+	if ( pNote->get_length() == -1 ) {
+		// No custom length set by the user. Play the whole sample.
+		nNoteLength = pSample->get_frames();
+	}
+	else {
 		// The user set a custom duration of the note in the
 		// PatternEditor. This will be used instead of the full sample
 		// length.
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
I have to admit that looking at `Sampler::renderNoteResample()` and `Sampler::renderNoteNoResample()` does not make me too confident about that the `Sampler` is doing fine. They are mostly the same but differ at a couple of places where I can't shake the feeling that just one of them is right.

cme:
Yeah, I keep meaning to refactor this to avoid the duplication. I think the NoResample variant only exists as an optimisation of the common case, because the resample loop is *hideously* inefficient. There are far better ways to tackle that though.

I'm apprehensive about the correctness because the logic has diverged and the more complex case, the Resample case, is the one that should be retained and optimised, but because it's more commonly used, NoResample has had more testing exposure :/

cme (approved):
Looks good and the intent is a lot clearer!

theGreatWhiteShark:
It's a bit of a shame these very essential parts of the code did grew so naturally. The Timeline was another example. Neither `AudioEngine` nor `Sampler` were properly adapted and before rewriting the AudioEngine we basically had glitches every single time a tempo marker was passed or a manual tempo change occurred.

> NoResample has had more testing exposure

I'm not sure about that. E.g. the samples of the GMRockKit have a sample rate of 44100 while my JACK server is running on 48000. This requires 100% resampling.

PR #1877, opened 2023-10-13
Sampler: refactoring

theGreatWhiteShark:
- `Sampler::renderNoteResample` and `Sampler::renderNoteNoResample` have been fused and just the former one survived.
- refactor names and check constness

---

The rendering unit tests fail due to a problem which might be unrelated to the `Sampler` itself. It's `src/tests/CoreActionControllerTest.cpp` and in particular the tear down method which messes things up. Commenting it and the rendering tests pass just fine. Seems creating and assigning an empty song has some strange side effects. I'll check (next week)

Addresses #1873


cme (approved):

theGreatWhiteShark:
> The rendering unit tests fail due to a problem which might be unrelated to the Sampler itself. It's src/tests/CoreActionControllerTest.cpp and in particular the tear down method which messes things up. Commenting it and the rendering tests pass just fine. Seems creating and assigning an empty song has some strange side effects. I'll check (next week)

We are seriously talking about a **single** frame which is consistently off 😅 . Well...

theGreatWhiteShark:
> >  The rendering unit tests fail due to a problem which might be unrelated to the Sampler itself. It's src/tests/CoreActionControllerTest.cpp and in particular the tear down method which messes things up. Commenting it and the rendering tests pass just fine. Seems creating and assigning an empty song has some strange side effects. I'll check (next week)

> We are seriously talking about a single frame which is consistently off 😅 . Well...

Turns out it was a bug in the refactoring after all. In case of no resampling the `Sampler` read one frame _beyond_ the `Sample` buffer.

Issue #1849, opened 2023-09-29
Proper notion of 'current drumkit'


theGreatWhiteShark:
Right now, a `.h2song` file and the corresponding `Song` class contains instrument list, components, and some selected meta data of the last loaded drumkit. After loading the user can add/remove instruments and components. All of these constitute what we call "current drumkit" right now (which can be altered via the main menu). It is about 95% what makes out the drumkit. But the remaining 5% cause trouble and inconsistencies.

Two things are bothersome in particular:

1. Editing properties of the current drumkit will save it to disk and thus override the original. E.g. load kit A, add several instruments, alter the license, and kit A will be overridden and will contain additional instruments. This is counter intuitive since the properties of kit A can be edited without side effects via right clicking the kit in the Sound Library
2. Loading a drumkit into a song will discard the instrument IDs of the kit. E.g. 1. load "GMRockKit" into a song, 2. load "Audiophob", 3. Save the drumkit. The original "Audiophob" kit used instrument IDs `0, 21, 19, 16, ...` and was overridden with `0, 1, 2, 3, ...`

Especially 2. is a nuisance in implementing #1691.

I would fix this by putting an actual `Drumkit` into the `Song` class. It's a massive refactoring :grimacing: . But it would significantly clean up the `Song` class.

On the downside the resulting `.h2song` will not be forward-compatible. This means, `.h2song` created with <=1.2.2 can be loaded in >=1.3 but not the other way around. That's quite unfortunate but will, at least in my point of view, pay off in the long run. To ease the pain, there will be the option of saving a song in "legacy" mode compatible with <=1.2.2.

theGreatWhiteShark:
In addition, all the stuff specified in the `SampleEditor` is only stored to the `.h2song` right now. When adjusting settings there, saving the resulting drumkit, and loading them again the adjustments are lost.

That can't be right.

PR #1878, opened 2023-10-13
Add option to align notes to the current grid.

cme:
This can be used to automatically quantise notes already recorded or entered, and also can be used to change patterns to swung by selecting a triplet grid.

A straight 4/4 pattern...

image

Select 1/8T and hit 'align'

image

...and now it's swung!

image


src/gui/src/PatternEditor/PatternEditor.cpp
```diff
@@ -409,6 +410,60 @@ void PatternEditor::selectInstrumentNotes( int nInstrument )
...
+	auto pInstrumentList = pHydrogen->getSong()->getInstrumentList();
+	QUndoStack *pUndo = HydrogenApp::get_instance()->m_pUndoStack;
+	validateSelection();
+
+	std::list< Note * > notesToAlign;
+	for ( auto pNote : m_selection ) {
+		notesToAlign.push_back( pNote );
+	}
+	// If selection is empty, use all notes in the pattern
+	if ( notesToAlign.empty() ) {
```
theGreatWhiteShark, line 435: I was a little bit surprised by this behavior. Mostly, because the action is listed next a lot of selection-based actions, like Cut, Copy, Select All, Clear Selection.

Possible side effects: When working on a pattern, shrinking it to e.g. half its size, and pressing <kbd>Ctrl</kbd> + <kbd>A</kbd> only notes currently accessible within the reduced pattern length will be selected. When using `get_notes()` all notes regardless whether they are accessible or not will be selected.

theGreatWhiteShark (commented):
Nice. I like it!

I would also suggest we provide this action as a keyboard shortcut as well.

theGreatWhiteShark:
There is one thing I just noted (which is caused by the PR itself). Deduplication of notes is not triggered in all cases.

Steps to reproduce
1. Set resolution to `off`
2. Fill all notes of an instrument
3. Set resolution to `1/4`
4. Align to grid
5. Click into the Song Editor

Now, whenever one uses the Pattern Editor it does not deduplicate the notes automatically anymore.

cme:
> There is one thing I just noted (which is caused by the PR itself). Deduplication of notes is not triggered in all cases.
> 
> Steps to reproduce
> 
>     1. Set resolution to `off`
> 
>     2. Fill all notes of an instrument
> 
>     3. Set resolution to `1/4`
> 
>     4. Align to grid
> 
>     5. Click into the Song Editor
> 
> 
> Now, whenever one uses the Pattern Editor it does not deduplicate the notes automatically anymore.

I guess this only happens when using all-notes, rather than selected notes. SInce it's unexpected perhaps it's better to just make it a selection-only operation? Perhaps also disabling the menu option when there's no selection?

theGreatWhiteShark:
> I guess this only happens when using all-notes, rather than selected notes. SInce it's unexpected perhaps it's better to just make it a selection-only operation? Perhaps also disabling the menu option when there's no selection?

I would be very much in favor of making "randomize velocity" and "fill notes" selection-only operations and to move them into the context menu opening when right clicking the pattern editor. I think it is both quite helpful and more clear. With your recent Pattern Editor UX update these old per-instrument operations feel kinda out of place.

But in order to do so we have to slightly rethink mouse-based selection. When done using keyboard one can select arbitrary rectangle and could in general also provide operations to act on the empty space or whole area. But using the mouse only note selection is possible for now.

cme:
> I would be very much in favor of making "randomize velocity" and "fill notes" selection-only operations and to move them into the context menu opening when right clicking the pattern editor. I think it is both quite helpful and more clear. With your recent Pattern Editor UX update these old per-instrument operations feel kinda out of place.
> 
> But in order to do so we have to slightly rethink mouse-based selection. When done using keyboard one can select arbitrary rectangle and could in general also provide operations to act on the empty space or whole area. But using the mouse only note selection is possible for now.

Agreed, "randomize" should probably be a selection type operation. Perhaps some sort of "draw" type operation for "fill notes" might work?

theGreatWhiteShark:
> Agreed, "randomize" should probably be a selection type operation. Perhaps some sort of "draw" type operation for "fill notes" might work?

+1 

But we do not need to have all of this in a single PR.

cme:
> But we do not need to have all of this in a single PR. 

Aye. I've moved the randomisation into the pattern editor class and runs on selections now using the more generic note properties change UndoAction and deleted the redundant randomisation one. UndoActions still needs refactored,  but not in this PR.

theGreatWhiteShark:
Nice. I'll check it out tomorrow.

> UndoActions still needs refactored, but not in this PR.

I'm doing a _massive_ refactoring right now for #1849 (https://github.com/theGreatWhiteShark/hydrogen/tree/phil-current-drumkit). Unfortunately, I did not find a nice way to split it into smaller PRs without risking to give users a hard time who build from `master` in between.

It would be probably best to wait with the UndoAction refactoring until the other one is done.

PR #1884, opened 2023-10-19
Sampler: trigger note release on custom length

theGreatWhiteShark:
custom note lengths are used to trigger the release of a note and the `Sampler` will play them back till the additional release frames are rendered. Previously, the custom length was used to stop audio rendering of a note creating a pronounced artifact.

From the looks of the code it is very likely that this worked for a long time but was broken just recently. Well, happens. But should not happen again as custom note lengths are now covered by the unit tests and even a tiny tweak on the release knob for the test song should make them fail.

Addresses #1873


theGreatWhiteShark:
Turns out these patches also fix the rendering issues with active filtering. 🥳

PR #1915, opened 2023-12-24
Fix interpolation at beginning / end of samples

cme:
Fix undetected typo error in finding the source frames at the beginning / end of samples which would only ever get the first two frames from the sample. Barely noticable audio impact since, but was still wrong.


cme:
@theGreatWhiteShark I have a problem recreating the reference FLACs for the tests this affects. When taking the calculated output wav, `flac` gives a warning about the WAV mismatching:
```
test4800032-aZeVBo.wav: WARNING: legacy WAVE file has format type 1 but bits-per-sample=32
test4800032-aZeVBo.wav: wrote 840759 bytes, ratio=0.673
```

and then libsoundfile fails to load the resultant flac with
```
- File contains data in an unimplemented format.
```

Is there secret sauce needed?

theGreatWhiteShark:
> Is there secret sauce needed?

Hmm... IIRC I just used the temporary file created during the test and converted it using `ffmpeg`.

Sounds like you found a bug while fixing another one :D

cme:
> > Is there secret sauce needed?
> 
> Hmm... IIRC I just used the temporary file created during the test and converted it using `ffmpeg`.

Aha! FFMPEG does the job just fine. Thanks.

Looks like the FLACs for the reference files are 24-bit fixed resolution, and that's what FFMPEG generates from the 32-bit float WAV files, which seems entirely reasonable. No idea what the `flac` program is emitting!

PR #1917, opened 2023-12-31
feat: add pt_BR translation 

giovana-morais:
i left some translations with the `type="unfinished"` tag because i'm not 100% sure of the translated sentence, so if anyone has any input, it would be appreciated!


theGreatWhiteShark:
Hey @giovana-morais ,

Thanks for your work!

> i left some translations with the type="unfinished" tag because i'm not 100% sure of the translated sentence, so if anyone has any input, it would be appreciated!

Sure. Any specific one? (Since there are 632 "unfinished" tags left I can't elaborate on all of them)

giovana-morais:
hi @theGreatWhiteShark! the PR is still in progress, which is why i still have so many "unfinished" tags. (: 
i'm aiming to finish everything by the weekend! then we can take a look at what remains tagged.

giovana-morais:
ok, this is good to review! i have no idea how to translate some of the `unfinished` ones, but now they are 81 instead of 632, so i'm assuming it's fine? 

anyway, let me know if there is anything else i can do. and thanks again for the fantastic project (:

theGreatWhiteShark:
> i have no idea how to translate some of the unfinished ones, but now they are 81 instead of 632, so i'm assuming it's fine?

Just to be sure: have you checked the [manual](http://hydrogen-music.org/documentation/manual/manual_en.html#chpt.instrument_editor) regarding the unfinished Instrument Editor labels? How to translate things like "Auto-Stop Note" or "HH Press. Grp" might be less difficult when you understand what the feature is doing. (and at least for the latter that is, unfortunately, not evident when looking at its name or toying with it in the GUI).

I think for the remaining ones we do things in batches instead of all at ones:
- triplets are a special form of [tuplets](https://en.wikipedia.org/wiki/Tuplet). The resulting string is shown in the "Res"(olution) combo box in the pattern editor
- "L. Gain" is the gain of an instrument layer
- "C. Gain" is the gain of an instrument component
- "JACK Timebase master" is a technology to synchronize tempo between independent applications running on the same device. The "master" (unfortunate choice of word) is the one allowed to set/change speed while all other applications just listen to changes.
- The "license string" is the actual string we write to disk as part of the .h2song, .h2drumkit, and .h2pattern files. The "License" is parsed from the former and displayed as a set of predefined option in a combo box. (You can view those tooltips in Drumkit > Properties)
- " VK note-on pitch X" indicates to which MIDI NOTE-ON message of pitch X a certain keystroke of the virtual keyboard (VK) will be mapped. You can check the translations in Options >  Preferences > Shortcuts

giovana-morais:
hi @theGreatWhiteShark! thanks for your patience. I think now I'm done with the translations. I have no idea how to translate the remaining items.

theGreatWhiteShark:
> I think now I'm done with the translations.

Nice. Thank's a lot!

(Are you interested in doing further tweaks to the translation in the future? Development for 1.3 is not done yet but there will most probably be a bunch of new strings. Usually we announce a so-called "string freeze" period about a month before the first beta release on our [mailing list](http://lists.sourceforge.net/mailman/listinfo/hydrogen-devel). At that point we do our best to not change any translatable string till the final release. If you want, you could subscribe and check out future changes)

giovana-morais:
@theGreatWhiteShark nice, i'll take a look. i'm not a professional translator by any means, but i'd be happy to contribute. (:

theGreatWhiteShark:
> i'm not a professional translator by any means, but i'd be happy to contribute. (:

And we are glad you did :)

By the way, the translation you contributed was based on the strings of the 1.3 release (at some not so near point in the future). For all upcoming 1.2.X patch releases [a different translation file](https://github.com/hydrogen-music/hydrogen/blob/releases/1.2/data/i18n/hydrogen_pt_BR.ts) (the one in the `releases/1.2` branch) will be used. I back ported your changes as careful as possible. Nevertheless, there are still some untranslated strings, which won't be present in 1.3 anymore. A tiny number.

PR #1918, opened 2023-12-31
Use imported drumkit immediately on open from Finder

cme:
Previously, importing drumkit by opening it in Finder would import the kit, but not update the widget or provide any other feedback to let the user know it was actually successful, or allow them to start using it.

### DIFF
--- a/src/gui/src/MainForm.cpp
+++ b/src/gui/src/MainForm.cpp
@@ -1775,8 +1775,22 @@ bool MainForm::eventFilter( QObject *o, QEvent *e )
 			}
 
 		} else if ( sFileName.endsWith( H2Core::Filesystem::drumkit_ext ) ) {
-			H2Core::Drumkit::install( sFileName );
-
+			QString sImportedPath;
+			if ( H2Core::Drumkit::install( sFileName, "", &sImportedPath ) ) {
+				auto pSoundLibraryDatabase = pHydrogen->getSoundLibraryDatabase();
+				if ( pSoundLibraryDatabase != nullptr ) {
+					pSoundLibraryDatabase->update();
+					auto pDrumkit = pSoundLibraryDatabase->getDrumkit( sImportedPath );
+					if ( pDrumkit == nullptr ) {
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
> Previously, importing drumkit by opening it in Finder would import the kit, but not update the widget or provide any other feedback to let the user know it was actually successful, or allow them to start using it.

It does not? Any error appearing in the logs?

On `master` I get both a dialog "drumkit was successfully imported" (or something similar) and the Sound Library widget is updated with the imported kit appearing as well.

Since #1905 (on `master` as well) this can be done by right-clicking within the Sound Library widget or using the "Drumkit" item in the main menu bar. But only for the latter the imported drumkit will be loaded right away (in order to emphasis that actions within the Sound Library do not affect the current song and those accessible from the menu do not affect the kits in the SL).

> On reflection, a far more reasonable user expectation is that opening a drumkit in Finder should select that drumkit, even if it's already imported.

Right. What happens if the kit is already imported? I think right now we just overwrite the existing one which is not good. On the other hand, maybe the kit in the SL is a modified version and the user wants to reset it by reimporting it. Maybe we should show a dialog "Kit is already present" with the options load, replace, and cancel or similar.

cme:
The `FileOpen` event is macOS specific and is sent to the QApplication when the user opens something in Finder. 

I think you're right, a confirmation dialog would be good. I should probably make the behaviour consistent between opening in Finder and using the Import dialog.

theGreatWhiteShark:
> The FileOpen event is macOS specific and is sent to the QApplication when the user opens something in Finder.

Ah. I see. I thought you used the import option from within Hydrogen and `Finder` was used instead of a `QFileDialog`.

> I should probably make the behaviour consistent between opening in Finder and using the Import dialog.

Yeah. Maybe we can put all of it into `CoreActionController` (and cover potentially missing GUI-related updates with a dedicated `EventType`). We also have an OSC event for importing kits (`/Hydrogen/EXTRACT_DRUMKIT/` called with just one argument) as well as a command line option in `h2cli`. They should all behave the same.

PR #1922, opened 2023-12-31
Sampler performance improvements and refactoring

cme:
Refactoring some of the sampler for maintainability and performance.

This restores the slight performance loss from removing the old `renderNoteNoResample()` and improves further by simplifying some things and removing cruft to get rid of things standing in the way of compiler optimisation, so now linear interpolation is as fast as no-resampling *used* to be, and cubic interpolation is as fast as linear was previously. Overall improvement in audio engine performance is about 25% on "slow" hardware (using a raspberry pi 4 AArch64 as reference).

The downward trend in processing time is illustrated really roughly as:

image

(10 runs through the benchmark for each commit, hence the steps)


cme:
There's potential for more as well since the resampling *should* be able to be vectorised with a few tweaks.

cme:
> There's potential for more as well since the resampling _should_ be able to be vectorised with a few tweaks.

After experimenting with this a bit, there's only marginal gain in making it vectorisable, since there's data movement needed to gather/repack the loaded samples into output-vector format for processing and that removes almost all the benefit (on the RPi anyway). Since it comes at the expense of more code, seems best to leave it.

PR #1945, opened 2024-03-02
Fix missing include QFileDialog

cme:
This include was necessary but was missed because precompiled header usage masked the problem.

### DIFF
--- a/src/gui/src/Widgets/FileDialog.h
+++ b/src/gui/src/Widgets/FileDialog.h
@@ -23,6 +23,7 @@
 #define FILEDIALOG_H
 
 #include <QDialog>
+#include <QFileDialog>
 #include <core/Object.h>
 
 /** Custom file dialog checking whether the user has write access to


theGreatWhiteShark (approved):
Looks good.

With the additional include I can compile the code without precompiled headers.

The Windows pipeline broke a couple of weeks ago. I did hope it would eventually heal itself since it seems to be related to the pacman config/cache/installed package in the AppVeyor instance. But it seems we have to fix it ourselves.

cme:
There seems to be something broken in the appveyor cache that's stoping the macOS build as well. I've hopefully cleared the cache (why isn't there a real API for that in appveyor?) so that should be unstuck at least. :/

cme:
Clearing the cache sorted the OS X build and #1946 should sort the Windows build (it's a bit of a hack, and I can't work out what changed in MSYS to cause it to break, but better a working CI than not!)

PR #1958, opened 2024-03-25
Preferences: drop restoreLastSong|Playlist

theGreatWhiteShark:
`Preferences` options `restoreLastSong` and `restoreLastPlaylist` were dropped. Instead, both will always be restored automatically. I think this is a reasonable UX nowadays. Especially, since the default for `restoreLastPlaylist` is currently `false`. This took me by surprise at least twice by now.

In case anyone want's to start from a new song/playlist, she could (with #1955) just open a new song/playlist and close Hydrogen. Since (https://github.com/hydrogen-music/hydrogen/pull/1955/commits/b0cda77a5b40889e3ae6a1c2ee4159f158de527c) an empty string will be stored as last file. In my point of view this is reasonable since it is done after an explicit user interaction to discard the current file. For songs we even do not loose anything, since the previous last song is still listed in "recently opened".

But it is a breaking change after all. So, I split it off #1955. @cme what do you think? Is this an improvement?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,6 +42,10 @@ XXXX-XX-XX the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 	* Fixed
 		- Components can now carry arbitrary names and name duplication is handled
 			properly.
+	* Removed
+		- Preferences options `restoreLastSong` and `restoreLastPlaylist` were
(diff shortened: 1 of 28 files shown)


cme:
I think user expectation these days is tending towards a pseudo-persistent model, especially since Hydrogen is a single-document app, so I think this is probably a good idea :)

PR #1962, opened 2024-03-30
Remove autosave files when user declines to restore from them

cme:
Remove autosave files when user declines to restore from them
    
When Hydrogen starts up, if a file has been autosaved the user will
be offered the opportunity to restore from that file. If they decline
it, Hydrogen should delete the autosaved file; it may be soon overwritten anyway by new autosave data.
    
This avoids the annoying cycle I often find myself in:
   1. edit file
   2. quit Hydrogen discarding edits (but autosave remains)
   3. open Hydrogen
   4. hit "Don't Save" because I didn't want those edits
   5. listen to original file without modifying (hence no autosave, but old autosave remains)
   6. quit Hydrogen (silently)
   7. goto 3
    
Also, when opening a file and restoring from its autosave, Hydrogen
should count this as the song being Modified because the current
content is from the autosave and is not the same as the content of
the named file.

This change also removes spurious sets of `IsModified` in parts of the GUI, which would `setIsModified( true )` in the GUI initialisation despite not making any actual changes, as well as `setModified( false )` calls that were needed to work around these spurious sets after song loading completed.

### DIFF
--- a/src/core/CoreActionController.cpp
+++ b/src/core/CoreActionController.cpp
@@ -77,14 +77,16 @@ bool CoreActionController::setStripVolume( int nStrip, float fVolumeValue, bool 
 	
 	auto pInstr = getStrip( nStrip );
 	if ( pInstr != nullptr ) {
-	
+
+		if ( pInstr->get_volume() != fVolumeValue ) {
+			pHydrogen->setIsModified( true );
+		}
+
 		pInstr->set_volume( fVolumeValue );
 	
 		if ( bSelectStrip ) {
 			pHydrogen->setSelectedInstrumentNumber( nStrip );
 		}
-	
-		pHydrogen->setIsModified( true );
 
 		return sendStripVolumeFeedback( nStrip );
 	}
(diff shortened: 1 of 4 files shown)


theGreatWhiteShark:
Haha, setting autosave to `0` is also the first thing I do. It's quite annoying during development.

> If they decline it, Hydrogen should delete the autosaved file

I also noted that. In #1955 I used a "No" instead of "Discard" button. But breaking the cycles sound like a very good idea.

How about we offer all three options: yes, no, discard. This way the user can view both original and modified version before deciding whether to discard or apply the changes.

cme:
> How about we offer all three options: yes, no, discard. This way the user can view both original and modified version before deciding whether to discard or apply the changes.

Excellent idea, I like it.

theGreatWhiteShark (commented):
Looks good. But there still seem to be some spurious calls to `setIsModified` during startup. When

1. Create a new song
2. Save it
3. Quit Hydrogen and open it again

the empty song is marked "modified"

cme:
Oh good catch, I'll look into it. Thanks.

PR #1978, opened 2024-06-05
Do not mask CMAKE_CXX_FLAGS environment variable

tartina:
Thr user's own CMAKE_CXX_FLAGS got lost using $ENV assignement

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,9 +171,9 @@ else()
 endif()
 
 if(WANT_DEBUG)
-	set(CMAKE_CXX_FLAGS "$ENV{CMAKE_CXX_FLAGS} -O0")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0")
 else()
-	set(CMAKE_CXX_FLAGS "$ENV{CMAKE_CXX_FLAGS} -O3 -ffast-math")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -ffast-math")
 endif()
 
 if(WANT_APPIMAGE)


theGreatWhiteShark (commented):
Hey @tartina ,

I'm puzzled. I introduced these `ENV` modifiers for that very reason: to not mask the user's variable.

The way I understand `cmake` `${VAR}` does access a variable defined _within_ the `CMakeLists.txt` or any other file directly imported. `$ENV{VAR}`, however, does access a variable defined on the _system_.

When I add these debug statements into `CMakeLists.txt`

```cmake
color_message("${_escape}[1;34m *** CMAKE_CXX_FLAGS: ${CMAKE_CXX_FLAGS} ${_escape}[0m")
color_message("${_escape}[1;34m *** ENV(CMAKE_CXX_FLAGS): $ENV{CMAKE_CXX_FLAGS} ${_escape}[0m")
```

I get

```bash
$ CMAKE_CXX_FLAGS="-test-flag" cmake ..
[...]
 *** CMAKE_CXX_FLAGS:  
 *** ENV(CMAKE_CXX_FLAGS): -test-flag 
[...]
```

How would you alter/provide CXX flags?

tartina:
It was exported by rom macro from Fedora build system %cmake. I got it only
removing the ENV part. Also its not possible to choose a release with debug
build type.

Il Ven 7 Giu 2024, 10:46 theGreatWhiteShark ***@***.***> ha
scritto:

tartina:
Sorry I was partially wrong.

The Fedora macro defines CXXFLAGS, not CMAKE_CXX_FLAGS which is not defined in the environment.
But cmake docs say that CMAKE_CXX_FLAGS is initialized from some internal configs and CXXFLAGS.
So using ENV you are deleting this initialization, if CMAKE_CXX_FLAGS env variable is empty. It should not be defined as env variable according to the docs

theGreatWhiteShark (approved):
> The Fedora macro defines CXXFLAGS, not CMAKE_CXX_FLAGS which is not defined in the environment.
But cmake docs say that CMAKE_CXX_FLAGS is initialized from some internal configs and CXXFLAGS.
So using ENV you are deleting this initialization, if CMAKE_CXX_FLAGS env variable is empty. It should not be defined as env variable according to the docs

Ah. I see. Seems like I misread the `cmake` doc.

Thanks for bringing this up! Luckily it was not part of any release code yet.

Issue #1070, opened 2021-01-05
Mapping pattern between different drumkits


theGreatWhiteShark:
If a pattern contains e.g. a snare rimshot but the drumkit does not it would most probably better to replace it by a snare hit than omitting it.

If we adopt the `#instrument primary`, `#instrument secondary`, ... notation, all notes belonging to unavailable instruments could be mapped to the lowest ones (primary > secondary > ...).

For other instruments we could hard-code a mapping table, e.g. `tom tiny` > `tom not so tiny`; `tom not so tiny` > `tom high` (please don't mind the names),  and apply it over and over until all instruments are present in the target drumkit.

oddtime:
I am not sure about the "hard code" behaviour for any case.

example:
pattern has 1 splash, 1 china and no other cymbals,
the drumkit to map to has 2 crashes and no other cymbals
Shouldn't H2 map splash and china to different crashes? (trying to keep the instruments separate)

theGreatWhiteShark:
Ideally, there will also be a widget for the user to modify the results of the mapping. This hopefully won't take much longer than the mapping itself. So, I think it's not so important to cover every case. But switching from a something like a 90% GM-compatible kit to a 80% one should provide plausible mappings and be just answered with an OK.

nabero:
First contribution here, and not aware of the internals. Sorry if this comment is garbage. 

What about adding the concept of instrument group as a user defined list.
Hydrogen would come with default instruments (tom), drumkits would come with their own values of size (floor tom: 12, big tom: 42, tiny tom: 55), which can then be mapped to the target drumkit. Maybe labelled values (tiny=25) would help standardize some while keeping flexibility.

theGreatWhiteShark:
> What about adding the concept of instrument group as a user defined list.
Hydrogen would come with default instruments (tom), drumkits would come with their own values of size (floor tom: 12, big tom: 42, tiny tom: 55), which can then be mapped to the target drumkit. Maybe labelled values (tiny=25) would help standardize some while keeping flexibility.

So, If I get your concept right each drumkit comes with an unique integer number for each instrument and the current Hydrogen session holds a user defined list that maps maps these integer numbers to the instrument of a pattern, right?

nabero:
Yes. But an instrument would also have an additional instrument category so
that selection of alternatives can only be done in that category.

I was thinking of bundling hydrogen with a default list, editable by the
user, and from which you can assign the value from for each instrument in
[...]

Issue #1068, opened 2021-01-05
Add general instrument property to Note and Drumkit 


theGreatWhiteShark:
To make a `Pattern` independent from a specific `Drumkit` some intermediate property is required to map between them.

In this issue I would like to discuss the naming and implementation of the property itself. Since we need to specify them in a xsd file, the list must, unfortunately, the exhaustive.

For starters I would use naming like `snare primary`, `snare secondary`, ... and go to a depth that still needs to be defined. This concept I would apply to all main instruments and it's up to the person creating the drumkit to decide what is primary, secondary etc. But what's a main instrument? Will there be eight different main toms (with the aforementioned depth since e.g. in *Big Mono* toms are separated into "hit by left/right hand")? Same for the hihat. All the others are more or less straight forward.

oddtime:
What about the splash and china cymbals?

trebmuh:
Thinking out loud.

Wouldn't using GM (General MIDI) drumkit be a first step?
See #38 for a bit of background information.

As a user, I'd then thins that when I've got a song and I want to change the drumkit, there is 2 possibilities:
1. both the current drumkit and the one I want to change to are GM compliant, hence 99% all good
2. one or both isn't/aren't GM compliant, hence I would propose H2 to show me a windows  asking how hydrogen should "translate" the different instrument from the current drumkit, to the one I want to change to. On this windows, I could/should be able to play every sounds from both the drumkits to be able to choose what's needed.

This way, it would avoid the burden of being exhaustive, or more likely, trying to be.
Thoughts?

oddtime:
> On this windows, I could/should be able to play every sounds from both the drumkits to be able to choose what's needed

User should be able to save the mapping configuration, to import another pattern later

oddtime:
@theGreatWhiteShark do you mean "Add general instrument property to Instrument and Instruments in Drumkit"?

trebmuh:
_Still thinking out loud here._

[...]

PR #2000, opened 2024-07-26
Drumkit-independent patterns (#1984)

theGreatWhiteShark:
Implementation of [Drumkit-independent patterns proposal V2](https://github.com/hydrogen-music/hydrogen/discussions/1984).

Patterns are now independent of Drumkits and the latter can switched without the need to adjust the patterns. Mapping between the two will be done using "instrument types".

I also added `.h2map` files for all kits we host at SourceForge (and a couple more).

Both the implementation of the dialog to remap notes of one or many patterns to different instruments of the current drumkit and the rendering of notes not associate with any instrument of the current kit will be postponed. It requires some refactoring of `DrumPatternEditor` which I want to split off into a separate PR.

All in all things might be still a little rough around the edges cycling through kits already works fine.

In addition the following MIDI and OSC actions were added:
- `LOAD_PREV_DRUMKIT` (cycling through drumkits)
- `LOAD_NEXT_DRUMKIT` (cycling through drumkits)

As well as the CLI option:
- `kitToDrumkitMap`: to extract a .h2map file from a drumkit

The proposal was added in `/docs/proposals` and I will start using ADRs from now on.

Implements #1984, #1068, #1070


theGreatWhiteShark:
A massive PR again. Sorry about that. But it was either this or leaving `master` in a somewhat inconsistent state.

PR #2012, opened 2024-08-14
Rework copy/paste of notes of mult. instruments

theGreatWhiteShark:
Fixes a regression causing multiple instrument copy/paste to fail.

Seems like the parts for multi instrument cut/copy/paste were some ancient relics. They used `std::list<Pattern*>` instead of `PatternList*` and featured some PatternList deserialization which already diverged from `PatternList::load_from`. I reworked those parts replaced the wrappers in `Song` with our default de-/serialization code for `PatternList`.

Missing note mapping kept the pasted notes from appearing in the pattern editor. I fixed that one was well.


theGreatWhiteShark:
Superseded by #2014

PR #2031, opened 2024-09-17
Show reporter on normal child exit if exit code non-zero

cme:
On Windows, `QProcess::exitStatus()` returns `Normal` for many circumstances that would be `CrashExit` on other operating systems. So check the exit *code* as well, which will be non-zero on most Windows abnormal exits.


theGreatWhiteShark (approved):
Nice. Tested on my local Windows machine. Work's just fine. Thanks!

PR #2036, opened 2024-10-01
Director: clear 'next' tag when tags are the same

aldimond:
Fixes an issue apparently introduced in commit d27917977b when some
variables were renamed. In the Director window, when the 'current' and
'next' tags are the same, it's the 'next' tag that should be cleared,
not the 'current' one.


theGreatWhiteShark (approved):
Thanks!

PR #2042, opened 2024-10-10
org.hydrogenmusic.Hydrogen.desktop, add X-NSM-Exec: when used in NSM,…

grammoboy2:
… hydrogen can't be opened with a file, as it should wait for the open message from the NSM daemon

X-NSM-Capable, is used to detect if the application has NSM support. It should ideally become false if the application is build without NSM support.
X-NSM-Exec, is the exec name of the application when used in NSM. It can't contain a path (not /usr/bin/ for example) or command line arguments like %f.

### DIFF
--- a/linux/org.hydrogenmusic.Hydrogen.desktop
+++ b/linux/org.hydrogenmusic.Hydrogen.desktop
@@ -28,3 +28,4 @@ StartupNotify=true
 
 Icon=org.hydrogenmusic.Hydrogen
 X-NSM-Capable=true
+X-NSM-Exec=hydrogen


theGreatWhiteShark:
Hey @grammoboy2 

File validates. Format seems fine.

Do you have any link or document about the `X-NSM-Exec` key? I couldn't find any?

grammoboy2:
Hi @theGreatWhiteShark 

It's discussed on the nsmd forks github issue tracker, issue 40. 

X-NSM-Capable is to tell if the application has NSM support or not.

X-NSM-Exec is to tell which exec name should be used when running in NSM, as this can't contain a path, nor command line arguments. X-NSM-Exec can be left out from the *.desktop file if the Exec name is the same as X-NSM-Exec, but I think it's not bad to add it as at some point one can decide to pass %f to Exec, add a path or something.

For Hydrogen X-NSM-Exec it can't be left out, cause Exec=hydrogen %F, is using a command line option.

theGreatWhiteShark (approved):
> It's discussed on the nsmd forks github issue tracker, issue 40.

Ah, I see. Thanks for the PR!

theGreatWhiteShark:
> X-NSM-Capable, is used to detect if the application has NSM support. It should ideally become false if the application is build without NSM support.

AFAICS this has to be done by packagers using a patch. It has to be known _prior_ to calling hydrogen. So, runtime is not a good place to adjust it based on compilation settings. Changing it in an automated way during a compilation does not feel like a good idea either. It adds complexity and is more likely to mess up the setup of packagers than helping them.

grammoboy2:
Thx.

> X-NSM-Capable, is used to detect if the application has NSM support. It should ideally become false if the application is build without NSM support.

I meant at compilation time, Ardour seems to use *.desktop.in with variables, I didn't study it yet.

 If the packagers can easily use those compile options (they set how they like to compile their package right?), I don't see a problem for them.

theGreatWhiteShark:
> I did expect more questions about this indeed as well as about the API.

Haha. Well, we all have just limited resources. Plus, community is small. I still know you from the non mailing list and was entirely sure you could back it up with a reference. Just wanted it to be posted here for completeness (and out of interest).

> I meant at compilation time, Ardour seems to use *.desktop.in with variables, I didn't study it yet. Most developers just merged the pull request like you did though. I did expect more questions about this indeed as well as about the API.
>
> If the packagers can easily use those compile options (they set how they like to compile their package right?), I don't see a problem for them.

An *.in file would be an obvious entry point, you're right.

Hmm.. But still. I consider the whole OSC API an integral part of Hydrogen and once included NSM support is included too. Plus, it's activate per default and the packager would be required to explicitly decide against it (hopefully no one does). The AppImage we provide for all Linux users has full NSM support.

I would just keep it simple until it turns into an real-life problem.

PR #2049, opened 2024-10-17
Update libnsdfile.rb to recent

cme:


theGreatWhiteShark:
Hmmm... If the previous version of `libsndfile` was `1.0.31`, both FLAC and Ogg/Opus support should have been working. But `cmake` within the AppVeyor pipeline reported otherwise.

```
* libsndfile                   : + used ( /usr/local/lib/libsndfile.dylib )
*                                FLAC: not supported, OGG/Vorbis: not supported (>=1.0.17 required)
 				 OGG/Opus: not supported (>=1.0.28 required)
 				 MP3: not supported (>=1.0.31 required)
```
(from your latest commit https://ci.appveyor.com/project/mauser/hydrogen/builds/50817210/job/namux946l3r43sqw)

Is MP3 support in the resulting binary working or does the [version check in CMakeLists.txt](https://github.com/hydrogen-music/hydrogen/blob/master/CMakeLists.txt#L361) messes up things because it can not determine the version of the shared library?

theGreatWhiteShark:
Well, I already thought about possible ways to circumvent the 60 minute timeout to allow for `brew` to compile all dependencies. Like having two separated OS jobs using the same cache which both compile half of the dependencies and do only run in case of a cache miss.

But it turns out the timeout can be configured in AppVeyor's preferences. I increased it to 120 minute and run the OS job again. Let's see if this already does the job.

theGreatWhiteShark:
> But it turns out the timeout can be configured in AppVeyor's preferences. I increased it to 120 minute and run the OS job again. Let's see if this already does the job.

Well. Seems like "greater values does not apply to [our] plan".

cme:
> > But it turns out the timeout can be configured in AppVeyor's preferences. I increased it to 120 minute and run the OS job again. Let's see if this already does the job.
> 
> Well. Seems like "greater values does not apply to [our] plan".

Good grief, it seems to be rebuilding qt. That 'aint right. Maybe Howebrew needs to be updated before building the things that need explicitly rebuilt for the ABI change. I'll check.

cme:
Aaaha, Homebrew have removed support for Monterey. Looks like we'll have to update to a newer macOS image. I'll look into that.

PR #2066, opened 2024-11-19
Fix `libsndfile` support

theGreatWhiteShark:
I was a little bit too conservative when checking for format support in `CMakeLists.txt` based on the version found.

A proper version can only be found in case the package was discovered via `pkg-config`. If not, like on macOS when retrieved via `brew`, FLAC, OPUS, OGG, and MP3 support would be disabled. Regardless of the capabilities of the underlying `libsndfile` shared library.

Instead, the fallback is now to allow all possible format support in order to not introduce any regression.

(still a draft since I can only check whether this is actually working on an (old) macOS in a couple of days).

@cme FYI


theGreatWhiteShark:
I checked it on an old macbook and it works fine.

It actually worked better than expected. Even MP3 export succeeded although the version of `libsndfile` in our macOS pipeline cache - `v1.0.31` - should not support MP3 export yet (only provided from `v1.1.0` onward). But I'm not too familiar with macOS or the device I use in particular.

PR #2069, opened 2024-11-24
Fix lifetime issue with utf8 string in Drumkit::exportTo

cme:
The `QString::toUtf8()` function returns a `QByteArray` object. Previously, a reference to the contents of this was taken with `.constData()` but the object itself was transient and destroyed at the end of the statement, leaving a `const char*` pointing at freed memory.

### DIFF
--- a/src/core/Basics/Drumkit.cpp
+++ b/src/core/Basics/Drumkit.cpp
@@ -1047,7 +1047,8 @@ bool Drumkit::exportTo( const QString& sTargetDir, bool* pUtf8Encoded,
 	const auto targetPath = sTargetNamePadded.toStdWString();
 	nRet = archive_write_open_filename_w( a, targetPath.c_str() );
 #else
-	const auto targetPath = sTargetName.toUtf8().constData();
+	const auto targetPathUtf8 = sTargetName.toUtf8();
+	const auto targetPath = targetPathUtf8.constData();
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark (approved):
Damn. We use `toUtf8().constData()`, `toLocal8Bit().data()` and things alike in quite a number of places. Strange it did not cause issues yet. Well, maybe it did and we just couldn't reproduce the tickets.

I'll take care of fixing lifetime handling of the remaining occurrences.

cme:
> I'll take care of fixing lifetime handling of the remaining occurrences.

Using the memory sanitizer (-fsanitize=memory) should find these I think!

cme:
> Using the memory sanitizer (-fsanitize=memory) should find these I think!

Oops, I mean `-fsanitize=address`. I've just tried builds with that and it does indeed find this problem during the unit tests, and with this fix it is clean with no further asan violations :)

theGreatWhiteShark:
> I've just tried builds with that and it does indeed find this problem during the unit tests, and with this fix it is clean with no further asan violations :)

Well, not on my end :D I get both a couple of asan violations and classical test failures when adding this compile option. I'll take a look.

But I'm very glad we have all these unit tests helping us to track down potential issues.

cme:
Potentially GCC's asan covering some different cases from LLVM's. I certainly have asan failures if I try and actually run the GUI.

PR #2071, opened 2024-11-27
Fix issues found by address sanitization (1.2)

theGreatWhiteShark:
#2070 picked for `releases/1.2`.


src/core/AudioEngine/AudioEngineTests.cpp
```diff
@@ -3057,8 +3057,9 @@ void AudioEngineTests::throwException( const QString& sMsg ) {
 
 	pAE->setState( AudioEngine::State::Ready );
 	pAE->unlock();
-	
-	throw std::runtime_error( sMsg.toLocal8Bit().data() );
```
cme, line 3061: Did asan complain about this? I'd have thought that runtime_error should take a copy of the error string, which is still in-scope here

Issue #2051, opened 2024-10-17
`NotePropertiesRuler` tweaks


theGreatWhiteShark:
Some things in `NotePropertiesRuler` are not ideal yet

- [x] dragging of a non-selected note behaves differently than dragging a (single) selected one. (For the former the cursor does not change to a dragging one and it has to be located in the same column as the note.)

  It feels weird anyway to not discard the current selection when dragging a note which is not contained in it. So, I would suggest clicking a note for dragging should discard the previous selection and selects the clicked one instead.
- [x] dragging in NoteKey editor does not integrate well. There we have two rows: octave and key. But when selected - they are always selected in pairs - only the keys can be dragged. Instead, key and octave should be decoupled, key should be the default, but one should also be able to drag-select different octaves.
- [x] scale of the left side of keys in NoteKey editor does not integrate well
- [x] scale of the left side of octaves in NoteKey editor is missing

theGreatWhiteShark:
Implemented in #2089

PR #2085, opened 2025-01-17
PatternEditor: drumkit-independent pattern and modernization

theGreatWhiteShark:
@cme, @trebmuh could you check whether UI and UX still feels like Hydrogen and whether drumkit switching and pattern loading is intuitive? (P.S. If you are still on Hydrogen <= 1.2.3 you need to backup your `$USER/.hydrogen/data/drumkits` folder. Version 2.0 will make incompatible changes and requires at least >=1.2.4 to handle them)

---

TL;DR: The pattern editor does now show _all_ notes in a pattern and decouples it from the current drumkit. This is best illustrated by loading a (demo) song and selecting Drumkit > New in the main menu. In addition, UI and UX of all parts of the pattern editor have been align and user input is now based on existing notes.

---

Sorry for **such** a big PR. But the changes were quite severe and I thought it would be best to make many iterations of checking and tweaking UI and UX until everything felt right and there were no repressions anymore.

### Drumkit-independent patterns

The big new feature in Hydrogen 2.0 will be a decoupling of our patterns from our drumkits. This way we can seemlessly switch between drumkits as well as load patterns programmed for another kit. For most users the latter will be probably more interesting and I intend add an online import feature for both songs and patterns in 2.0 too. A user donated 260 patterns based on a public domain book that will serve as the initial seed.

Handling this within the audio engine and the remainder of the core part is already implemented on `master`. But prior to this patch only notes corresponding to instruments of the current drumkit could be shown in the pattern editor.

With this patch the sidebar of the pattern editor has been made wider to hold both labels and buttons for the instruments of the current kit (grey stuff on the left) as well as the instrument types (or ids as fallback). All instruments in the kits shipped by us will have types and the user can add/change them by double clicking the corresponding label or via the drum properties. When creating a note for an instrument, this note will inherit its type (and id). If we then switch to another kit, the note will be mapped to the instrument of the new kit holding the same type or - if none such exists - will be appended to the pattern editor in a line not associated with the current drumkit.

Even in case the drumkit the pattern was created with does not at all match the target kit, the user can still use it by moving/duplicating unmapped notes while relying on their instrument type as a hint of which sound they were intended to trigger.

I have to admit I do not like this second column of the sidebar much. It makes pattern creation more complex and takes some space. But at least from my point of view its benefits do outweighed its downsides.

### UI / UX alignment in the pattern editor

Previous UI:
old

Patchted UI:
new_dpe
new_pre

To manage all these rows not corresponding to the current drumkit in the different parts of the pattern editor, I decided to put the corresponding data structure into `PatternEditorPanel` and to make a large refactoring to use the former as the single source of truth. While the pattern editors somewhat felt like different entities previously, I did a lot of changes that were intended to convey the feeling that they are just different, synchronized views of the same data set.

Property drawing in `NotePropertiesRuler` is now bound to right-click drag exclusively because it got confused with lasso selection quite a lot. I'll do the some in the SongEditor as well to provide consistent mouse action throughout all editors.

The pattern name label has been moved to the very top of the pattern editor into a tab bar. In case we have several active patterns, all of them are represented as tabs and the pattern's properties can be accessed by double clicking it.

The piano roll editor does now look very similar to the drum pattern one and its left-hand side labels have been converted into a proper sidebar and do now tell which actual note they represent. This sidebar features a right-click popup menu similar to the one of the drum pattern editor. (Quite a wall of sidebars. I know. I thought about a feature to collapse the type labels to the size of the mute or solo button but decided against it. It wouldn't make things easier).

Dragging and drawing within the Key/Octave view of the ruler was fixed (addresses #2051).

### Pattern editor modernization

During testing I more and more grew annoyed of the previous way to interact with notes. Or at least that was what I would intuitively expect to do in the pattern editor. Instead, we did interact with the grid. If a note resided on the grid, we were lucky and could interact with it. If it was off grid, we could still select it and remove it. But not editing its properties. When selecting it, we could move it, delete it etc. but it felt strange to lasso select a single note in order to move it.

With this patch interaction is now based on notes. When we hover a note using mouse or keyboard we can interact with it regardless of its position with respect to the grid. If the note is part of the current selection, the corresponding action is applied to all selected notes. If not, the current selection is discarded and only the hovered note (or notes in case of a chord) will be affected. So, we can now e.g. right-click and align to grid or move a single note without prior selection.

Note insertion is still grid-based and can be done by clicking while not hovering a note (for better interaction the mouse cursor is sensible to notes within a resolution-dependent radius).

### Quantization

While aligning input actions of the different parts I also toyed with off-grid placement of cursor, moved notes, or dragged note lengths when pressing the <kdb>Alt</kbd> modifier in the same way as we use it for more fine-grained changes in the properties ruler. And I liked it a lot.

How about we binds this feature to the quantization button in the pattern editor panel? It already has this very same effect on recorded notes. Plus, quantization is activated by default. What do you think?

And if you find any glitches, bugs, or inconsistencies, please let me know.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -85,16 +85,19 @@ XXXX-XX-XX the hydrogen team <hydrogen-devel@lists.sourceforge.net>
 				- Alt key modifier can be used to add/move/change off the current grid.
 				- Instruments of the current kit and instrument types of all notes in a
 				  pattern are now handled separately in the sidebar.
-				- Single notes can now be dragged without prior selection.
+				- Single notes can now be dragged or altered using right-click popup
+				  menu without prior selection.
 				- Notes not associated with an instrument of the current kit are now
 				  visible and can be handled like regular notes.
-				- Pattern name is now shown on top and several active patterns are
-				  presented in independent tabs.
-				- Property drawing can now only bound to right mouse button.
+				- Pattern name is now shown together with other active patterns tab bar.
+				- Property drawing (NotePropertiesRuler) is now only bound to right
+				  mouse button.
 				- Cursor, selection lasso, and selected and hovered notes are now synced
 				  between among all editors.
 				- Property changes based on wheel and key event on the same set of notes
 				  as well as property drawing can now be undone/redone in a single action.
+				- PianoRoll does now have a proper sidebar including per-line action in
+				  right click popup menu.
 	* Fixed
 		- Components can now carry arbitrary names and name duplication is handled
 			properly.
--- a/data/i18n/hydrogen_ca.ts
+++ b/data/i18n/hydrogen_ca.ts
@@ -927,7 +927,7 @@ Are you sure?</source>
     </message>
     <message>
         <source>Type</source>
-        <extracomment>Shown in table headers when referring to an instrument&apos;s type (as part * of a Drumkit Map .h2map).</extracomment>
+        <extracomment>Shown in table headers when referring to an instrument&apos;s type (as part * of a Drumkit Map .h2map) as well as part of status messages referring to * this property.</extracomment>
         <translation type="unfinished"></translation>
     </message>
     <message>
@@ -1307,6 +1307,19 @@ Please set your system&apos;s locale to UTF-8!</source>
     <message>
         <source>Load drumkit</source>
         <extracomment>Status message displayed when loading an drumkit into the current song. * Separated by a whitespace the name of the drumkit will be appended to * the translated message.</extracomment>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Edit type for notes in row</source>
+        <extracomment>Shown in the status bar when editting the type information of a row not * mapped to the current drumkit and as the title of the associated dialog. * Separated by a whitespace the corresponding column number will be * appended.</extracomment>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>An instrument type must not be empty!</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Instrument types must be unique!</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
@@ -1717,10 +1730,6 @@ Overwrite the existing file?</source>
     </message>
     <message>
         <source>Save a copy of the current drumkit to NSM session folder</source>
-        <translation type="unfinished"></translation>
-    </message>
-    <message>
-        <source>Instrument types must be unique!</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
@@ -3057,6 +3066,10 @@ It should work like a charm provided that you use the GMRockKit, and that you do
     </message>
     <message>
         <source>Set note lead/lag</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Set note type</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
--- a/data/i18n/hydrogen_cs.ts
+++ b/data/i18n/hydrogen_cs.ts
@@ -926,7 +926,7 @@ Are you sure?</source>
     </message>
     <message>
         <source>Type</source>
-        <extracomment>Shown in table headers when referring to an instrument&apos;s type (as part * of a Drumkit Map .h2map).</extracomment>
+        <extracomment>Shown in table headers when referring to an instrument&apos;s type (as part * of a Drumkit Map .h2map) as well as part of status messages referring to * this property.</extracomment>
         <translation type="unfinished"></translation>
     </message>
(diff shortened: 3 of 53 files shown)


cme:
Wow this is a lot of work! :)

I have built this and running from a pre-existing hydrogen config and an old h2song, I got a somewhat weird display with '+' icons next to each instrument. Not sure what that was about, I removed my config to start from scratch, and tried opening demos as you recommended, but consistently crash out with:

image


I'm surprised to see the role shown next to the instruments in the pattern editor, I think I assumed this would be more of a drumkit editor feature? Will see how that actually feels when I get it running though! :)

cme:
Aha, all it needed was:

```
diff --git a/src/core/SoundLibrary/SoundLibraryDatabase.cpp b/src/core/SoundLibrary/SoundLibraryDatabase.cpp
index 88c5bc595..ebb103135 100644
--- a/src/core/SoundLibrary/SoundLibraryDatabase.cpp
+++ b/src/core/SoundLibrary/SoundLibraryDatabase.cpp
@@ -301,7 +301,7 @@ void SoundLibraryDatabase::registerUniqueLabel( const QString& sDrumkitPath,
 }
 
 QString SoundLibraryDatabase::getUniqueLabel( const QString& sDrumkitPath ) const {
-       return m_drumkitUniqueLabels.at( sDrumkitPath );
+       return m_drumkitUniqueLabels.at( Filesystem::absolute_path( sDrumkitPath ) );
 }
 
 void SoundLibraryDatabase::registerDrumkitFolder( const QString& sDrumkitFolder ) {
```

theGreatWhiteShark:
> I have built this and running from a pre-existing hydrogen config and an old h2song, I got a somewhat weird display with '+' icons next to each instrument. 

And a number displayed on the left of the '+'? Seems you are using non of the kits we host on SourceForge in that song. The label right of the solo button is intended to display the instrument type of both instruments and notes. But those properties are very new and not present in versions < 2.0. I created so called `.h2map` shipped with Hydrogen's system data which will patch existing kits based on those hosted on SourceForge (you have to installed the compiled version in order to use them). For custom kits or new instruments the instrument type has to be added manually.

The idea of '+' is to indicate that the instrument type is missing and should be added. The number is a fallback to the instrument ID of the instrument / note. This way we can always display all notes of a pattern. Even legacy ones with no types but only ids and a reference to a kit that does not exist anymore.

> I'm surprised to see the role shown next to the instruments in the pattern editor, I think I assumed this would be more of a drumkit editor feature?

Yeah. In the beginning I thought so too. At first I just exposed them in the drumkit property dialog and intended to create a new widget - the Patch Bay - to map notes of a pattern to drumkit and the other way around on pattern and drumkit load. But having to go back into this widget to remap the pattern in case one was not satisfied with the result was very cumbersome. Instead, showing all notes of the pattern and using keyboard and mouse editing to fix issues in the original mapping felt way more easy and naturally.

The main idea of putting the type next to the instrument and using different colors was to indicate that the types could be thought of as properties (of the notes) of a pattern and the drumkit uses them to 'dock' its instrument. This should emphasize that drumkits are interchangeable. But maybe I overdid it and there is a better, more subtle way to indicate and handle it. Maybe even a stronger visual disconnection between the type and the instrument would make things more clear.

theGreatWhiteShark:
> Aha, all it needed was:
> 
> ```
> diff --git a/src/core/SoundLibrary/SoundLibraryDatabase.cpp b/src/core/SoundLibrary/SoundLibraryDatabase.cpp
> index 88c5bc595..ebb103135 100644
> --- a/src/core/SoundLibrary/SoundLibraryDatabase.cpp
> +++ b/src/core/SoundLibrary/SoundLibraryDatabase.cpp
> @@ -301,7 +301,7 @@ void SoundLibraryDatabase::registerUniqueLabel( const QString& sDrumkitPath,
>  }
>  
>  QString SoundLibraryDatabase::getUniqueLabel( const QString& sDrumkitPath ) const {
> -       return m_drumkitUniqueLabels.at( sDrumkitPath );
> +       return m_drumkitUniqueLabels.at( Filesystem::absolute_path( sDrumkitPath ) );
>  }
>  
>  void SoundLibraryDatabase::registerDrumkitFolder( const QString& sDrumkitFolder ) {
> ```

Uh. I'll have a look. Seems drumkits loading has to get more resilient.

theGreatWhiteShark:
> I'm surprised to see the role shown next to the instruments in the pattern editor

Another thing we could do is to hide the type labels in case there is no note in the current pattern not associated with the current drumkit (which should be the scenario faced most often). An option stored in the `hydrogen.conf` / `Preferences` and accessible in the popup of the Pattern Editor's sidebar could be called "Always show instrument types" and disable this behavior (off by default).

PR #2200, opened 2025-10-04
Fix the section / date format of the manpage

snd1:
Since the master branch does contain an older manpage than in the 1.2.6 release i changed in the 1.2.6 artifacts.
Please apply it in the right branch. Thanks :)


theGreatWhiteShark:
Hmm. The man page does need some love. And there isn't any for the `h2cli` and `h2player` tools...

@snd1 thanks for your contribution! Are you aware of any format checker for man pages / roff files we could use in our build pipelines (one that returns with a non-zero return code in case something is off)?

snd1:
@theGreatWhiteShark Sorry, unfortunately not, only lintian that does the job in debian.

PR #2261, opened 2026-01-07
Rework NOTE_OFF handling (#63)

### LINKED ISSUE #63: midi : midi out only : using instrument as sequencer for external app
currently if you want to use a H2 instrument as a pure sequencer (ie the notes in the pattern are only sent out as MIDI notes. H2 does not generate any sound) you MUST still load a long empty wav file into that instrument.
if you do not do this the MIDI notes will have a duration of 0. as a consequence the external app will not play anything.

the best option would be that H2 always sends out MIDI notes regardless of the duration of the loaded WAV file
an alternative would be to add a 'MIDI out only' tickbox on the General tab of the Instrument editor that automatically loads a 'very long' empty WAV into that instrument

this second option however is only a quick workaround

comment by Chris Mennie :
The purpose of having an associated sound file (which admittedly is a huge hack) is to tell Hydrogen how long to hold the note before a note off message is sent. So it's not just a case of specifying a "very long" duration of sound file because different drum sounds require different play lengths.

As far as the code goes, I believe it'd be straight forward to define a default note length to use when there's no associated sound sample. However, this doesn't really solve the real issue (though it does go much farther than what's currently in place). I don't know how it should be presented, but the user needs to be able to specify the note length (via the UI? via midi input?).

theGreatWhiteShark:
In order to address a couple of frequent requests/issues regarding MIDI message handling, a couple of patches were applied:

- An option in `Preferences` was added to omit NOTE_OFF messages 1. for notes without custom length, 2. at all.
- Instruments without any samples can now send NOTE_ON and NOTE_OFF messages as well. The user can thus finally create "MIDI-only" drumkits. 
- MIDI message sending was aligned with mute/solo states of the overall instrument, component, and layer and does now behaves identical to audio rendering.

Fixes #63

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ All notable changes to this project will be documented in this file.
     messages if the corresponding settings are checked.
   - The channel for MIDI feedback (and clock messages) can now be configured as
     well (setting it to "Off" makes Hydrogen drop those messages) (#779).
+  - Whether or not Hydrogen is sending `NOTE_OFF` messages can now be configured
+    in the MIDI control dialog (#63).
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
Unit tests for Linux and Windows succeeded. But Windows pipeline in AppVeyor failed since it has trouble with its Qt setup (unrelated to this PR). The macOS AppVeyor image, on the other hand, is currently dysfunctional.

PR #2286, opened 2026-02-07
Improve UX of notes with custom lengths

theGreatWhiteShark:
While working on some other issues related to notes of custom length, I found them quite tedious to work with. For a pattern containing only notes of custom length, each single one had to be created via mouse click and then in a second operation the length could be set by click-dragging.

With this patch notes of custom length / property can be created right away. No prior creation of the "standard" note required anymore.

The pattern editor shows the transient ("soon to be created") note during dragging in the same style it does while moving notes. This helps a lot when fine tuning note length e.g. while bypassing quantization with the <kbd>Alt</kbd> modifier.

Another problem was that they are not easy to select. Although they took up quite some space in the pattern editor due to their tails, they still could only be interacted with using the note "head". Changing the length of the tail could not be done relative to the previous one but required the user to also start anew (and remembering the old value).

With this patch note tails under point will be taken into account as well, while notes heads have priority.

### DIFF
--- a/src/gui/src/PatternEditor/PatternEditor.cpp
+++ b/src/gui/src/PatternEditor/PatternEditor.cpp
@@ -1549,9 +1549,25 @@ std::vector<std::shared_ptr<Note> > PatternEditor::getElementsAtPoint(
 				}
 			}
 		}
+	}
 
-			if ( nDistance <= nLastDistance &&
-				 ppNote->getPosition() == nLastPosition ) {
+	if ( notesUnderPoint.size() == 0 ) {
+		// No notes near point. Are there might be notes of custom length, which
+		// have a tail under point? Those tails could span the whole row and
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
Hey @devloop0123,

since you work with notes of custom lengths a lot, could you check whether this patch makes handling them more convenient?

devloop0123:
>  could you check whether this patch makes handling them more convenient?

It does indeed!

devloop0123:
Custom length notes drawn in "drumkit editor" mode end up at pitch B6 instead of C3.

theGreatWhiteShark:
> Tails of notes at the initial position in a pattern are not interactive.

Thanks for testing again!

Should be fixed in 00e38d726f690a5096f6abbc5a4fb7f29a67d290

PR #2288, opened 2026-02-08
Fix Note-Off handling on stop (#2285)

### LINKED ISSUE #2285: note-off on stop
With setting " always send note-off", H2 sends a note-off on "stop" for all the notes defined in the instrument midi output mapping tab. 

But, since stopping playback during a note with custom length doesn't stop H2 sending a proper note-off at the correct time anyway, that "note-off on stop" is either superfluous (if the currently playing notes are offset from the default), or it prematurely stops the currently playing notes.

theGreatWhiteShark:
Previously, Hydrogen send `NOTE_OFF` messages to all instruments (at their default note value) whenever transport was stopped by the user or the end of the song was reached. That's neither expected nor helpful.

Instead, it now does so only when stopping audio rendering using `panic` - <kbd>F12</kbd>. Also, it takes a more fine-grained approach and only pairs `NOTE_OFF` messages with all `NOTE_ON` messages from the last 15 seconds, which were not followed by the corresponding `NOTE_OFF` yet.

It does so using the backlog of `MidiBaseDriver` and suffers from its constraints: it can't handle more than 200 messages and will be discarded when pressing the bin icon in the `MidiControlDialog`. But I think this is totally fine in here.

In addition to those individual `NOTE_OFF` messages, Hydrogen does also send a `CHANNEL_MODE` "All Notes Off" message to the feedback channel. According to the MIDI 1.0 Spec those `CHANNEL_MODE` messages should be send to the "Basic Channel" of the receiver, which defaults to `1` as our feedback channel does. 

Fixes #2285.

### DIFF
--- a/src/core/IO/MidiBaseDriver.cpp
+++ b/src/core/IO/MidiBaseDriver.cpp
@@ -180,6 +180,11 @@ MidiBaseDriver::getHandledOutputs()
 
 void MidiBaseDriver::sendAllNotesOff()
 {
+	const auto pPref = Preferences::get_instance();
+	if ( pPref->getMidiSendNoteOff() == Preferences::MidiSendNoteOff::Never ) {
+		return;
+	}
+
 	const auto threshold =
 		Clock::now() -
 		std::chrono::seconds( MidiBaseDriver::nAllNotesOffThresholdInSeconds );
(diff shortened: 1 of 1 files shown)


devloop0123:
Purely theoretical, but if one daisy-chains gear, sending midi over different channels to different instruments, an instrument listening to channel 2 messages probably won't respond to an all-notes-off message on channel 1?

devloop0123:
> that the target channel for the all-notes-of message is not the channel of a particular instrument/voice. Rather it is a channel you tell the target MIDI device to listen to for such kind of messages.

Thanks for the clarification. (I don't have any outboard gear, just Pianoteq to test with...)

PR #2300, opened 2026-02-26
Link Backtrace_LIBRARIES to hydrogen-core

rrrapha:
On OpenBSD (and FreeBSD/NetBSD as well) `-lexecinfo` is needed for libhydrogen-core.
(`find_package(Backtrace)` defines `Backtrace_LIBRARIES` and `Backtrace_INCLUDE_DIRS`)

This is the linker error I get on OpenBSD:

```
ld: error: undefined reference: backtrace
>>> referenced by src/core/libhydrogen-core-1.2.6.so.2.0 (disallowed by --no-allow-shlib-undefined)

ld: error: undefined reference: backtrace_symbols
>>> referenced by src/core/libhydrogen-core-1.2.6.so.2.0 (disallowed by --no-allow-shlib-undefined)
c++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -55,6 +55,16 @@ target_link_libraries(hydrogen-core-${VERSION}
 	Qt${QT_VERSION_MAJOR}::Gui # For QColor
 )
 
+if(Backtrace_FOUND)
+    target_include_directories(hydrogen-core-${VERSION}
+        PRIVATE
+        ${Backtrace_INCLUDE_DIRS}
+    )
+    target_link_libraries(hydrogen-core-${VERSION}
+        ${Backtrace_LIBRARIES}
+    )
+endif()
+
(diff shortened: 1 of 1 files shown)


theGreatWhiteShark:
Hey @rrrapha ,

Thank's for your contribution!

It seems to build fine on both Linux and macOS. But it does break our Windows pipeline https://ci.appveyor.com/project/mauser/hydrogen/builds/53609767/job/90cnhl9fj3vvnvaa#L1513 .

We already have two checks for *BSD in our [CMakeLists.txt](https://github.com/hydrogen-music/hydrogen/blob/main/CMakeLists.txt#L100)

```
if("${CMAKE_SYSTEM_NAME}" MATCHES "NetBSD")
```

How about reusing this one to restrict linking to this library to *BSD?

rrrapha:
The following is already in `src/gui/CMakeLists.txt` and seems not to cause problems:
```
if(Backtrace_FOUND)
        target_link_libraries(hydrogen
                ${Backtrace_LIBRARIES}
        )
endif()
```
I will try to do it similarly in `src/core/CMakeLists.txt`.

PR #2330, opened 2026-05-06
Sound Library rewrite

theGreatWhiteShark:
## Problem

While Hydrogen has three (major) file types meant for sharing: `.h2drumkit`, `.h2song`, and `.h2pattern`, our `SoundLibraryPanel` did only yield proper support for drumkits. Although patterns and songs were also be displayed and could be loaded, the supported feature set was much larger for drumkits. And it did not scale well growing number of drumkits.

All these downsides haven't been mended yet because prior to 2.0 it only made sense to shared drumkits. Songs and patterns were associate with a certain kit and had to be rewritten / altered when porting to the user's favorite kit. Drumkits, on the other hand, are much harder to create at a quality level worth sharing. So, there were only few kits and almost no one thought about sharing songs or patterns.

But with version 2.0 this is going to change. Patterns - and by extension songs - are now independent of drumkits and are worth sharing. This will probably result in the sound library growing rather large. In fact, we already ship quite a number of patterns ourselves (https://github.com/hydrogen-music/hydrogen/pull/2322).

### Before

image

### After

image

## Changes

internally both the `SoundLibraryPanel` and `SoundLibraryTree` as well as `SoundLibraryInfo` and large parts of file related handling in `Filesystem` and others have been rewritten with the goal to be artifact-agnostic. This way both songs and patterns do now support (mostly) the same set of actions available for drumkits.

Major user-facing changes have been:

- Drumkit, pattern, and song trees have been split into individual tabs
- subfolders of arbitrary depth are now allowed in the system and user data folder
- A search bar on top allows to narrow the tree leafs displayed by both names and tags.
- A button next to the search bar triggers a rescan of the entire library (e.g. in case a resource was changed or added externally)
- A detail display at the bottom shows metadata of the selected artifact. Using a right-click context menu the user can adjust which metadata to show.
- For both patterns and instrument multi selection and grouped import via drag & drop or right-click context menu are now supported (implements #1087)


trebmuh:
Looks awesome @theGreatWhiteShark ! :+1:

PR #2337, opened 2026-05-18
Minor tweaks and annual timestamp update

theGreatWhiteShark:


theGreatWhiteShark:
Merged regardless of CI failure. Qt6 in AppVeyors Ubuntu images is bricked right now. Not related to the changes in here.

PR #2339, opened 2026-05-19
Comply with XDG paths on Linux (#643)

### LINKED ISSUE #643: ~/.hydrogen does not follow XDG standards

theGreatWhiteShark:
With this patch new Hydrogen installations on Linux will respect XDG paths:

- The config file is written to  `$XDG_CONFIG_HOME/hydrogen/hydrogen.conf`
- The cache folder goes to `$XDG_CACHE_HOME/hydrogen/`
- All the data subfolders (drumkits, plugins, playlists, etc.) along with the hydrogen.log file go to `$XDG_DATA_HOME/hydrogen/ `

If old folder (`~/.hydrogen/`) is found, `hydrogen` uses this one instead. To port an existing installation to XDG paths, move the `~/.hydrogen` folder to a different location (and copy all data and the configuration into the new folders).

Note that this only affects Linux since it only was requested by Linux users.

Fixes #643
Supersedes #1624

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -110,7 +110,7 @@ for:
 
       echo $DISPLAY_VERSION
 
-      cache_tag=usr_cache_jack # this can be modified to rebuild deps
+      cache_tag=usr_cache_online # this can be modified to rebuild deps
 
       cdir=$HOME/cache_dir
       cache_tar=$cdir/$cache_tag.tar
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,10 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- On Linux user-level files are now stored in XDG paths
+  (`$XDG_CONFIG_HOME/hydrogen/hydrogen.conf`, `$XDG_CACHE_HOME/hydrogen/`,
+  `$XDG_DATA_HOME/hydrogen/ `) in case no `~/.hydrogen` folder is present.
+  (#643).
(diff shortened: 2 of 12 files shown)


trebmuh:
> If old folder (~/.hydrogen/) is found, hydrogen uses this one instead. To port an existing installation to XDG paths, move the ~/.hydrogen folder to a different location (and copy all data and the configuration into the new folders).

What about Hydrogen moving this folder automatically? It would be better for user not knowing how to do this, or who do not want to spend time doing it.

theGreatWhiteShark:
> What about Hydrogen moving this folder automatically? It would be better for user not knowing how to do this, or who do not want to spend time doing it.

Yes. This migration path is only for those few users eager to remove as many folders from their home as possible. All others are most probably happy to continue using `~/.hydrogen` for the time being.

If possible, I would shy away from moving folders automatically. Too much can go wrong, like restoring an old backup including `~/.hydrogen` could cause a loss of data in the XDG counterparts already existing on disk. Emacs is having a similar approach.

But the folder hierarchy should make it into the man files. I forgot about this one.

